PR #3, opened 2022-01-31
Skeleton for new framework.

hamadise:
The current code does :

- parse configuration
- discover pods under test
- build a shim layer to provide all the necessary data
- run pre-Stop test on all containers under test
- develop infrastructure to run commands on containers
- develop cnf-fsdiff to test if the container installed new packages


edcdavid (approved):
Looks like a great start to me :-)

greyerof (approved):

PR #21, opened 2022-02-10
migrate lifecycle suite

hamadise:


cnf-certification-test/lifecycle/ownerreference/ownerreference.go
```diff
@@ -0,0 +1,62 @@
...
+
+func NewOwnerReference(put *v1.Pod) *OwnerReference {
+	o := OwnerReference{
+		put:    put,
+		result: tnf.ERROR,
+	}
+	return &o
+}
+
+// func (f *FsDiff) RunTest(
```
edcdavid, line 45: fix comment

cnf-certification-test/lifecycle/suite.go
```diff
@@ -36,19 +37,88 @@ var _ = ginkgo.Describe(common.LifecycleTestKey, func() {
...
+//nolint:dupl
+func testContainersReadinessProbe(env *provider.TestEnvironment) {
+	testID := identifiers.XformToGinkgoItIdentifier(identifiers.TestReadinessProbeIdentifier)
+	ginkgo.It(testID, ginkgo.Label(testID), func() {
+		badcontainers := []string{}
+		for _, cut := range env.Containers {
+			logrus.Debugln("check container ", cut.Namespace, " ", cut.Podname, " ", cut.Data.Name, " readiness probe ")
+			if cut.Data.ReadinessProbe == nil {
+				badcontainers = append(badcontainers, cut.Data.Name)
+				logrus.Errorln("container ", cut.Data.Name, " does not have ReadinessProbe defined")
```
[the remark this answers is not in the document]
hamadise: fixed.

cnf-certification-test/lifecycle/ownerreference/ownerreference.go
hamadise: fixed.

cnf-certification-test/lifecycle/suite.go
```diff
@@ -36,19 +37,103 @@ var _ = ginkgo.Describe(common.LifecycleTestKey, func() {
...
+
+func testContainersPreStop(env *provider.TestEnvironment) {
 	testID := identifiers.XformToGinkgoItIdentifier(identifiers.TestShudtownIdentifier)
 	ginkgo.It(testID, ginkgo.Label(testID), func() {
-		gomega.Expect(true).To(gomega.Equal(true))
 		badcontainers := []string{}
 		for _, cut := range env.Containers {
-			fmt.Println("container ", cut.Data.Name)
-			logrus.Debugln("check container ", cut.Data.Name) // maybe use different platform ?
+			logrus.Debugln("check container ", cut.Namespace, " ", cut.Podname, " ", cut.Data.Name, " pre stop lifecycle ")
```
edcdavid, line 52: Shouldn't we also store this log in the claim file instead of logrus? I think that it would provide more detailed information

hamadise: What I did like about logrus is the ability to see logs when running a particular test in the console.
I will look into overriding logrus calls so that it will print to claim files or to the claim file. Jira ticket will be created to track this, if not successfull, then I'll switch all the logrus here into claim file

edcdavid: I believe ClaimFilePrintf already prints both on console (same as the ginkgo output) and in the claim file

hamadise: I did not know that, in that case, I'll switch to tnf claim file printing

hamadise:
```
**bad containers [test-6bfb745768-cvdt8 test-6bfb745768-lgjj9]**
------------------------------
• [FAILED] [0.332 seconds]
lifecycle
/Users/shamadi/redhat/go/cnf-certification-test/cnf-certification-test/lifecycle/suite.go:34
  [It] lifecycle-pod-owner-type [lifecycle-pod-owner-type]
  /Users/shamadi/redhat/go/cnf-certification-test/cnf-certification-test/lifecycle/suite.go:123

  Begin Captured GinkgoWriter Output >>
    STEP: Testing owners of CNF pod, should be replicas Set 02/13/22 16:24:49.684
    **bad containers [test-6bfb745768-cvdt8 test-6bfb745768-lgjj9]**
  << End Captured GinkgoWriter Output
  ```
  
  code between ** is being logged by ginkgo, It does actually log to stdout, it just does not have the notion of logging level in it.  We don't know if that's an error/trace/debug message. I'm not sure which way to go here.
  Either we can create our own abstraction and add level notion to "tnf.claimPrint" then inside it filter based on level, and print using logrus. 
  
  For now, I did send all the failed pods to Ginkgo, until we decide what to do :) We can even explore using zap or any other logging library if needed.

File: cnf-certification-test/lifecycle/suite.go [hunk 4]
```diff
@@ -36,19 +37,103 @@ var _ = ginkgo.Describe(common.LifecycleTestKey, func() {
...
 			if cut.Data.Lifecycle.PreStop == nil {
 				badcontainers = append(badcontainers, cut.Data.Name)
 				logrus.Errorln("container ", cut.Data.Name, " does not have preStop defined")
```
greyerof, line 55: I'd use ClaimFilePrintf here instead of logrus. Also, add the pod name and namespace so it has more information so the customer can find this target easier.

cnf-certification-test/lifecycle/suite.go
```diff
@@ -36,19 +37,103 @@ var _ = ginkgo.Describe(common.LifecycleTestKey, func() {
...
 	testID := identifiers.XformToGinkgoItIdentifier(identifiers.TestShudtownIdentifier)
 	ginkgo.It(testID, ginkgo.Label(testID), func() {
-		gomega.Expect(true).To(gomega.Equal(true))
 		badcontainers := []string{}
 		for _, cut := range env.Containers {
-			fmt.Println("container ", cut.Data.Name)
-			logrus.Debugln("check container ", cut.Data.Name) // maybe use different platform ?
+			logrus.Debugln("check container ", cut.Namespace, " ", cut.Podname, " ", cut.Data.Name, " pre stop lifecycle ")
 			if cut.Data.Lifecycle.PreStop == nil {
 				badcontainers = append(badcontainers, cut.Data.Name)
```
greyerof, line 54: Add the pod name and namespace as part of the string so they can be located easier.

cnf-certification-test/lifecycle/suite.go
```diff
@@ -36,19 +37,103 @@ var _ = ginkgo.Describe(common.LifecycleTestKey, func() {
...
-})
+func testContainersImagePolicy(env *provider.TestEnvironment) {
+	testID := identifiers.XformToGinkgoItIdentifier(identifiers.TestImagePullPolicyIdentifier)
+	ginkgo.It(testID, ginkgo.Label(testID), func() {
+		badcontainers := []string{}
+		for _, cut := range env.Containers {
+			logrus.Debugln("check container ", cut.Namespace, " ", cut.Podname, " ", cut.Data.Name, " pull policy, should be ", v1.PullIfNotPresent)
+			if cut.Data.ImagePullPolicy != v1.PullIfNotPresent {
+				badcontainers = append(badcontainers, cut.Data.Name)
+				logrus.Errorln("container ", cut.Data.Name, " is using ", cut.Data.ImagePullPolicy, " as image policy")
```
greyerof, line 73: Same here. When reviewing the claim file, the customer will only see a list of containers name (that are usually the same across pods) but no reason why they failed the TC, so I'll switch this to ClaimFilePrintf. I prefer this way because customers normally check the DCI's job web results first, which are built by parsing the claim file. The console log should be the last resource for debugging. Also, add pod name and namespace in the trace, please.

greyerof (commented):
See my comments about the ClaimFilePrintf usage when the target fails the check.
Also, the need for the "ownerreference.go" file seems unnecessary to me, since the real check (the loop over the OwnerReferences ) is quite small and can be inlined in the real test case. Maybe we should discuss whether we should implement all the checks **for all TCs** of the new tnf arch in separate files to mimic the old handlers to allow other organizations to create different test suites based on them.

cnf-certification-test/lifecycle/suite.go
greyerof: Sure, just something like:
`badContainers = append(badContainers, fmt.Sprintf("%s.%s.%s", cut.Namespace, cut.Podname, cut.Data.Name))`
When I said "located easier", I as was referring to a failing TC and we/the customer trying to find the "bad" container spec to fix it.

cnf-certification-test/lifecycle/suite.go
hamadise: we are logging all the failed pods in the claimfile so 
```
if(len (badcontainers) > 0{
   tnf...
}
```
should we write that information twice ?

greyerof (approved):

PR #22, opened 2022-02-11
Adding runLocalCommand utils

edcdavid:
Adding  utility command to execute local commands

### DIFF
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -1,0 +1,38 @@
+// Copyright (C) 2020-2021 Red Hat, Inc.
+//
+// This program is free software; you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation; either version 2 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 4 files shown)


pkg/utils/utils.go
```diff
@@ -0,0 +1,38 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License along
+// with this program; if not, write to the Free Software Foundation, Inc.,
+// 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+package utils
+
```
hamadise, line 18: a package name `utils` is something we should avoid if possible.
package commandrunner maybe a  better naming. what do you think ?

edcdavid: sounds good to me, will rename it

edcdavid: fixed

hamadise (approved):

sebrandon1 (approved):

PR #24, opened 2022-02-16
Add Access control namespace testcases

edcdavid:
- Adding access control namespace testcase. Using 2 go clients:
  - apiextention for listing CRDs
  - Dynamic client for listing CRs

- Improved CR listing from original testcase to return every namespace an invalid CR is found in, instead of last one
- CRs with no namespace now will show invalid


cnf-certification-test/accesscontrol/namespace/namespace.go
```diff
@@ -0,0 +1,90 @@
+// Copyright (C) 2020-2021 Red Hat, Inc.
```
sebrandon1, line 1: 2022?

pkg/stringhelper/stringhelper.go
```diff
@@ -0,0 +1,37 @@
...
+// 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+package stringhelper
+
+import (
+	"strings"
+)
+
+// StringInSlice checks a slice for a given string.
+func StringInSlice(s []string, str string, contains bool) bool {
```
sebrandon1, line 24: Copy the unit tests over for this func as well?

internal/ocpclient/ocpclient.go
```diff
@@ -7,29 +7,30 @@ import (
...
+	RestConfig    *rest.Config
 
 	ready bool
 }
 
 var ocpClient = OcpClient{}
 
 // NewOcpClient instantiate an ocp client
-func NewOcpClient(filenames ...string) OcpClient {
+func NewOcpClient(filenames ...string) OcpClient { //nolint:funlen // this is a spectial function with lots of assignements
```
sebrandon1, line 30:
```suggestion
func NewOcpClient(filenames ...string) OcpClient { //nolint:funlen // this is a special function with lots of assignments
```

sebrandon1 (commented):
Just some comments, looks good otherwise!

cnf-certification-test/accesscontrol/namespace/namespace.go
edcdavid: fixed. I am not sure how this works. I read this page (https://matija.suklje.name/how-and-why-to-properly-write-copyright-statements-in-your-code). Seems like it is better to have one date instead of a range an not touch it unless something significant changes. Maybe we should discuss our rule and add it to the developing guide?

pkg/stringhelper/stringhelper.go
edcdavid: fixed

internal/ocpclient/ocpclient.go
edcdavid: fixed

cnf-certification-test/accesscontrol/namespace/namespace.go
sebrandon1: From what I know about copyrights in code, it's supposed to be like when the code was first committed.  And whenever you "touch" it, make sure the year is updated to the latest year.  We'll have to figure it out.

edcdavid: fixed. Note sure what is the rule we follow, maybe we could document it in the dev guide? Found this : https://matija.suklje.name/how-and-why-to-properly-write-copyright-statements-in-your-code

pkg/stringhelper/stringhelper.go
edcdavid: fixed

cnf-certification-test/accesscontrol/namespace/namespace.go
```diff
@@ -0,0 +1,90 @@
...
+			namespaceStr = ""
+		} else {
+			namespaceStr = fmt.Sprintf("%s", namespace)
+		}
+		if name == nil {
+			nameStr = ""
+		} else {
+			nameStr = fmt.Sprintf("%s", name)
+		}
+		crdNamespaces[namespaceStr] = append(crdNamespaces[namespaceStr], nameStr)
```
hamadise, line 87: do you want to append here even if namespace == "" ? or  should continue ?

cnf-certification-test/accesscontrol/namespace/namespace.go
```diff
@@ -0,0 +1,90 @@
...
+	if invalidCrs == nil {
+		invalidCrs = make(map[string]map[string][]string)
+	}
+	for _, crd := range crds {
+		crNamespaces, err := getCrsPerNamespaces(crd)
+		if err != nil {
+			return invalidCrs, fmt.Errorf("failed to get CRs for CRD %s - Error: %v", crd.Name, err)
+		}
+		for namespace, crNames := range crNamespaces {
+			if !stringhelper.StringInSlice(configNamespaces, namespace, false) {
```
hamadise, line 44: @sebrandon1 I think both of you did port this function over. The second to merge should fix the conflict.
To reduce it, Brandon can use stringhelper name in his code

pkg/stringhelper/stringhelper.go
hamadise: I would prefer removing the contains bool, it's confusing.

``
func StringInSlice(s []string, str string) bool{
}
``
the caller should decide if he's interested in true or false output.

hamadise: Also in your code, you may want to add helper function 
``
func hasIntersection(s []string, str []string) bool{ }
``
to simplify your code above

edcdavid: fixed

cnf-certification-test/accesscontrol/namespace/namespace.go
```diff
@@ -28,32 +28,36 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 )
 
+// TestCrsNamespaces finds the list of the input CRDs (crds parameter) instances (CRs) and verify that they are only in namespaces provided as input.
+// The list of CRs not belonging to the namespaces passed as input is returned as invalid
 func TestCrsNamespaces(crds []*apiextv1beta1.CustomResourceDefinition, configNamespaces []string) (invalidCrs map[string]map[string][]string, err error) {
+	// Initialize the top level map
 	if invalidCrs == nil {
```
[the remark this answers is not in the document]
edcdavid: fixed

cnf-certification-test/accesscontrol/namespace/namespace.go
edcdavid: This is the question I asked before about CR that are cluster wide. The decision was to fail them and to rely on the config file to not include them or skip them. Based on that, we should not continue

cnf-certification-test/accesscontrol/namespace/namespace.go
edcdavid: Did refactor

pkg/stringhelper/stringhelper.go
```diff
@@ -0,0 +1,51 @@
...
+	return fullMatch
+}
+
+// StringInSlicePartialMatch checks a slice for a given substring.
+func SubStringInSlice(s []string, str string) bool {
+	fullMatch, partialMatch := hasIntersection(s, []string{str})
+	return fullMatch || partialMatch
+}
+
+func hasIntersection(s, str []string) (fullMatch, partialMatch bool) {
```
hamadise, line 35: I don't understand the logic of this function. 
what do you mean by full match ? that s and str are equals ? or one element is s equals one element in str ?

edcdavid: I tried to follow the logic of the initial function. So with full match, we try to find a string that matches equally. with partial match we check if the string is a substring of one of the strings in the slice

edcdavid: Reverted the function to the original one as it is simpler for what we need there.

pkg/stringhelper/stringhelper.go
sebrandon1: > I would prefer removing the contains bool, it's confusing.
> 
> `func StringInSlice(s []string, str string) bool{ }` the caller should decide if he's interested in true or false output.

I just added the contains like two weeks ago.  🥲

pkg/stringhelper/stringhelper_test.go
```diff
@@ -0,0 +1,78 @@
...
+// You should have received a copy of the GNU General Public License along
+// with this program; if not, write to the Free Software Foundation, Inc.,
+// 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+package stringhelper
+
+import (
+	"testing"
+
+	"gotest.tools/assert"
```
sebrandon1, line 22: Should this be the `testify` library?

sebrandon1 (commented):
Two comments.  Otherwise looks good.

sebrandon1 (approved):

pkg/stringhelper/stringhelper_test.go
edcdavid: fixed

hamadise (approved):

PR #23, opened 2022-02-15
Operator TS - First TC migrated: operator-install-status-succeeded

greyerof:
This TC was part of the old combo TCs that were used to run against each
operator/csv under test. An independent (and public) test ID has been
created for this TC so in order to emulate the old combo TCs selection,
the use of the skip flag (-s) is required when launching the TNF.

+ Added OLM client to the ocp client struct.

**Backward Compatibility Notes:**
TNF v3.3 TC operator-install-status ID/TC has been divided into two different TCs + IDs, one for each part of the old arch TC:
- operator-install-status-succeeded: Implemented in this PR
- operator-install-status-no-privileges: Implemented in further PR/s.

PR #30, opened 2022-02-22
Operator TS - TC migration: install-status-no-privileges

greyerof:
This TC was part of the old combo TCs that were used to run against each
operator/csv under test. An independent (and public) test ID has been
created for this TC so in order to emulate the old combo TCs selection,
the use of the skip flag (-s) is required when launching the TNF.

**Backward Compatibility Notes:**
TNF v3.3 TC operator-install-status ID/TC has been divided into two different TCs + IDs, one for each part of the old arch TC:

operator-install-status-succeeded: Already implemented in PR #23
operator-install-status-no-privileges: Implemented in this PR.

### DIFF
--- a/cnf-certification-test/operator/suite.go
+++ b/cnf-certification-test/operator/suite.go
@@ -21,6 +21,7 @@ import (
 
 	"github.com/onsi/ginkgo/v2"
 	"github.com/operator-framework/api/pkg/operators/v1alpha1"
+	"github.com/sirupsen/logrus"
 	"github.com/test-network-function/cnf-certification-test/cnf-certification-test/common"
 	"github.com/test-network-function/cnf-certification-test/cnf-certification-test/identifiers"
 	"github.com/test-network-function/cnf-certification-test/pkg/provider"
@@ -41,6 +42,11 @@ var _ = ginkgo.Describe(common.OperatorTestKey, func() {
 	ginkgo.It(testID, ginkgo.Label(testID), func() {
 		testOperatorInstallationPhaseSucceeded(&env)
 	})
+
+	testID = identifiers.XformToGinkgoItIdentifier(identifiers.TestOperatorNoPrivileges)
+	ginkgo.It(testID, ginkgo.Label(testID), func() {
+		testOperatorInstallationWithoutPrivileges(&env)
+	})
 })
 
 func testOperatorInstallationPhaseSucceeded(env *provider.TestEnvironment) {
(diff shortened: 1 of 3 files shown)


cnf-certification-test/operator/suite.go
```diff
@@ -61,3 +67,32 @@ func testOperatorInstallationPhaseSucceeded(env *provider.TestEnvironment) {
...
+			logrus.Debugf("No clusterPermissions found in csv %s (ns %s)", csv.Name, csv.Namespace)
+			continue
+		}
+
+		for i := range clusterPermissions {
+			permission := clusterPermissions[i]
+			for ruleIndex := range permission.Rules {
+				if n := len(permission.Rules[ruleIndex].ResourceNames); n > 0 {
+					tnf.ClaimFilePrintf("CSV %s (ns %s) has cluster permissions on %d resource names.", csv.Name, csv.Namespace, n)
+					badCsvs = append(badCsvs, fmt.Sprintf("%s.%s", csv.Namespace, csv.Name))
```
hamadise, line 89: shouldn't you add break here and go to test next csv ?

greyerof: You're right. Done.

cnf-certification-test/operator/suite.go
```diff
@@ -61,3 +67,34 @@ func testOperatorInstallationPhaseSucceeded(env *provider.TestEnvironment) {
...
+	}
+
+	for _, csv := range env.Csvs {
+		clusterPermissions := csv.Spec.InstallStrategy.StrategySpec.ClusterPermissions
+		if len(clusterPermissions) == 0 {
+			logrus.Debugf("No clusterPermissions found in csv %s (ns %s)", csv.Name, csv.Namespace)
+			continue
+		}
+
+	LOOP:
```
sebrandon1, line 84: I'm torn about using labels/goto functions.  Obviously it works but I don't know if I like it...

sebrandon1: I think this whole thing needs to be broken down to reduce complexity.  Also allows for more granule unit testing to be added.

cnf-certification-test/operator/suite.go
```diff
@@ -61,3 +67,34 @@ func testOperatorInstallationPhaseSucceeded(env *provider.TestEnvironment) {
...
+		for i := range clusterPermissions {
+			permission := clusterPermissions[i]
+			for ruleIndex := range permission.Rules {
+				if n := len(permission.Rules[ruleIndex].ResourceNames); n > 0 {
```
sebrandon1, line 88:
```suggestion
			for _, rule := range permission.Rules {
				if n := len(rule.ResourceNames); n > 0 {
```

greyerof: I like that way, I've done it that way tons of times before... But now, working with external types (k8s ones), you don't really know the size of those objects, so each iteration means copying the full object to the iteration variable (rule, in this case). In case the object is too big, the linter will complain about this. Even if it works now because the type spec is small and the linter does not complain, you cannot guarantee that it will work forever, because in the next k8s version they might want to add new fields to the resources, making it bigger enough for the linter to complain. 
So, my way to address this now is: when iterating over arrays for external object types, I try to use iteration variables only if it's an array of pointers (like in the previous StrategySpec.ClusterPermissions loop). Otherwise, I use indexes variables like in this case.

cnf-certification-test/operator/suite.go
greyerof: I'll remove the loop break based on David's suggestion.

cnf-certification-test/operator/suite.go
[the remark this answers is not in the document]
greyerof: Yep, you're right, I didn't thought about it, but it's better for customers having the log of all invalid rules they need to fix.

edcdavid (approved):

sebrandon1 (approved):
Thanks!

PR #32, opened 2022-02-23
Add deployment scaling test

hamadise:
Add statefulset autodiscovery
Add HPA autodiscovery
Satrt working on HPA scale test:

### DIFF
--- a/cnf-certification-test/lifecycle/suite.go
+++ b/cnf-certification-test/lifecycle/suite.go
@@ -17,16 +17,24 @@
 package lifecycle
 
 import (
+	"time"
+
 	"github.com/onsi/ginkgo/v2"
 	"github.com/onsi/gomega"
 	"github.com/sirupsen/logrus"
 	"github.com/test-network-function/cnf-certification-test/cnf-certification-test/common"
 	"github.com/test-network-function/cnf-certification-test/cnf-certification-test/identifiers"
 	"github.com/test-network-function/cnf-certification-test/cnf-certification-test/lifecycle/ownerreference"
+	"github.com/test-network-function/cnf-certification-test/cnf-certification-test/lifecycle/scaling"
 	"github.com/test-network-function/cnf-certification-test/pkg/provider"
 	"github.com/test-network-function/cnf-certification-test/pkg/testhelper"
 	"github.com/test-network-function/cnf-certification-test/pkg/tnf"
+
 	v1 "k8s.io/api/core/v1"
+)
+
+const (
+	timeout = 60 * time.Second
 )
 
 //
@@ -35,7 +43,6 @@ import (
 var _ = ginkgo.Describe(common.LifecycleTestKey, func() {
 	var env provider.TestEnvironment
 	ginkgo.BeforeEach(func() {
-		provider.BuildTestEnvironment()
 		env = provider.GetTestEnvironment()
 	})
 	testContainersPreStop(&env)
@@ -43,6 +50,7 @@ var _ = ginkgo.Describe(common.LifecycleTestKey, func() {
 	testContainersReadinessProbe(&env)
 	testContainersLivenessProbe(&env)
 	testPodsOwnerReference(&env)
+	testScaling(&env, timeout)
 })
 
 func testContainersPreStop(env *provider.TestEnvironment) {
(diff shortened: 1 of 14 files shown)


cnf-certification-test/lifecycle/suite.go
```diff
@@ -23,8 +23,16 @@ import (
...
 	"github.com/test-network-function/cnf-certification-test/cnf-certification-test/identifiers"
 	"github.com/test-network-function/cnf-certification-test/cnf-certification-test/lifecycle/ownerreference"
+	"github.com/test-network-function/cnf-certification-test/internal/ocpclient"
 	"github.com/test-network-function/cnf-certification-test/pkg/provider"
 	"github.com/test-network-function/cnf-certification-test/pkg/tnf"
+
+	v1app "k8s.io/api/apps/v1"
+	v2 "k8s.io/api/autoscaling/v2"
+
+	//autoscalingv2 "k8s.io/client-go/applyconfigurations/autoscaling/v2"
```
sebrandon1, line 33: This comment can be removed.

internal/ocpclient/ocpclient.go
```diff
@@ -8,19 +8,27 @@ import (
...
+
+	//"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/kubernetes"
+	appv1client "k8s.io/client-go/kubernetes/typed/apps/v1"
 	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
 )
 
 type OcpClient struct {
```
sebrandon1, line 22: Should we rename this struct since we're using the k8s client now?

pkg/autodiscover/autodiscover.go
```diff
@@ -84,7 +89,10 @@ func DoAutoDiscover() (env configuration.TestParameters,
 	debugPods = findPodsByLabel(oc.Coreclient, debugLabels, debugNS)
 	crds = FindTestCrdNames(testData.CrdFilters)
 	csvs = findOperatorsByLabel(oc.OlmClient, []configuration.Label{{Name: tnfCsvTargetLabelName, Prefix: tnfLabelPrefix, Value: tnfCsvTargetLabelValue}}, testData.TargetNameSpaces)
-	return env, testData, pods, debugPods, crds, namespaces, csvs
+	dps := findDeploymentByLabel(oc.AppsClients, testData.TargetPodLabels, namespaces)
+	statefulSets := findStatefulSetByLabel(oc.AppsClients, testData.TargetPodLabels, namespaces)
+	hpas := findHpaControllers(*oc.K8sClient, namespaces)
+	return env, testData, pods, debugPods, crds, namespaces, csvs, dps, statefulSets, hpas
```
sebrandon1, line 95: Just a thought, instead of returning all of these values should we encapsulate these into a struct and then just return that?  That way if we need to add/remove more values in the future it's easier just to add the variable into the struct instead of modifying this list of returned vars?

cnf-certification-test/lifecycle/suite.go
```diff
@@ -141,3 +149,93 @@ func testPodsOwnerReference(env *provider.TestEnvironment) {
...
+		max = hpa.Spec.MaxReplicas
+
+		// scale up the deployment
+		min++
+		max++
+		hpa.Spec.MinReplicas = &min
+		hpa.Spec.MaxReplicas = max
+		//config := autoscalingv2.HorizontalPodAutoscalerApplyConfiguration{}
+
+		//ocp.K8sClient.AutoscalingV2().HorizontalPodAutoscalers(deployment.Namespace).Apply(context.TODO(), hpa.Spec, v1app.ApplyOptions{})
```
sebrandon1, line 220: These can be removed.

cnf-certification-test/lifecycle/suite.go
```diff
@@ -141,3 +149,93 @@ func testPodsOwnerReference(env *provider.TestEnvironment) {
...
+		} else {
+			min = 1
+		}
+		max = hpa.Spec.MaxReplicas
+
+		// scale up the deployment
+		min++
+		max++
+		hpa.Spec.MinReplicas = &min
+		hpa.Spec.MaxReplicas = max
```
sebrandon1, line 217: Why is `min` a pointer and `max` is not?

sebrandon1 (commented):
Just some initial comments.  I see you have a bunch of commented code so I'll re-review when you're ready.

cnf-certification-test/lifecycle/suite.go
hamadise: min is a pointer, so if it's nil,  k8s will use the default value which is 0. having a pointers allows k8s to know if the user provided a value or not.
Not sure why max is not a pointer, maybe should be always more than 1 ? 🤔

PR #35, opened 2022-02-24
helper function to wait for the daemonset to be ready

edcdavid:
This is to wait for the daemonset debug pods to be ready. useful for testcases using debug pods


hamadise (approved):

pkg/provider/provider.go
```diff
@@ -98,3 +109,42 @@ func GetTestEnvironment() TestEnvironment {
...
+		}
+		if daemonSet.Status.DesiredNumberScheduled == daemonSet.Status.CurrentNumberScheduled && //nolint:gocritic
+			daemonSet.Status.DesiredNumberScheduled == daemonSet.Status.NumberAvailable &&
+			daemonSet.Status.DesiredNumberScheduled == daemonSet.Status.NumberReady &&
+			daemonSet.Status.NumberMisscheduled == 0 {
+			isReady = true
+		}
+		logrus.Debugf("Waiting for debug pods to be ready: %v", &daemonSet.Status)
+		time.Sleep(time.Second)
+	}
```
sebrandon1, line 143: Could this section be broken out into its own function and unit tested?

sebrandon1 (changes requested):
Needs a unit test on conditions for the daemonset to be ready but looks good otherwise.

pkg/provider/provider.go
edcdavid: fixed

sebrandon1 (approved):

PR #59, opened 2022-03-09
Added Get function to get the clientsholder's singleton.

greyerof:
As a singleton, the "NewClientsHolder" function name was a bit
misleading, since it will only return a new struct the first time it's
called.


sebrandon1 (approved):
As a singleton this makes sense to me.  👍

hamadise (approved):

PR #61, opened 2022-03-10
Use moq library to mock out testing AutomountServiceToken func

sebrandon1:
Notable changes:
- Added `make generate` and `make install-moq` Makefile paths.
- Created `AutomountTokenFuncs` interface to allow for mocking.
- Added `tnf.IsUnitTest` function to distinguish if we are running 'unit test' mode.  Wrappers for various ginkgo/gomega calls then utilize this to determine if we need to expect a value to be empty, for example.
- Unit tests now cover all of the scenarios the actual test would hit when running in the real world in suite.go.


sebrandon1:
Marked WIP as I finalize some stuff.

hamadise (approved):

edcdavid (commented):
Looks good for current code but will this be needed when/if we remove Ginkgo?

cnf-certification-test/accesscontrol/suite.go
```diff
@@ -449,6 +452,12 @@ func TestAutomountServiceToken(env *provider.TestEnvironment) {
 	if n := len(failedPods); n > 0 {
 		logrus.Debugf("Pods that failed automount test: %+v", failedPods)
 		tnf.ClaimFilePrintf("Pods that failed automount test: %+v", failedPods)
-		ginkgo.Fail(fmt.Sprintf("% d pods that failed automount test", n))
+		if !tnf.IsUnitTest() {
```
edcdavid, line 455: Why not create a ginkgoFail() function that would make the ginkgo call and the if

cnf-certification-test/accesscontrol/suite.go
```diff
@@ -388,15 +388,19 @@ func TestPodClusterRoleBindings(env *provider.TestEnvironment) {
...
+	if !tnf.IsUnitTest() {
+		ginkgo.By("Should have automountServiceAccountToken set to false")
+	}
 
 	msg := []string{}
 	failedPods := []string{}
 	for _, put := range env.Pods {
-		ginkgo.By(fmt.Sprintf("check the existence of pod service account %s (ns= %s )", put.Namespace, put.Name))
-		gomega.Expect(put.Spec.ServiceAccountName).ToNot(gomega.BeEmpty())
+		if !tnf.IsUnitTest() {
```
edcdavid, line 400: Why not create a GinkgoBy() and GinkgoExpect function that would make the ginkgo calls and the if

cnf-certification-test/accesscontrol/suite.go
sebrandon1: Done.

cnf-certification-test/accesscontrol/suite.go
sebrandon1: Done.

PR #63, opened 2022-03-11
Create interfaces for nodetained using moq

sebrandon1:
Similar to: https://github.com/test-network-function/cnf-certification-test/pull/61

Using `moq` to build out mocked versions of the functions used in the `nodetainted` test and test the suite.go flow functions.  Eventually the goal being that we are able to fully test our code in unit test start to finish and cover as much as possible.

Marked WIP as I finish up these test cases.


cnf-certification-test/platform/nodetainted/nodetainted.go
```diff
@@ -29,22 +28,35 @@ import (
...
-// NewNodeTainted creates a new NodeTainted tnf.Test.
-func NewNodeTaintedTester(timeout time.Duration, client *clientsholder.ClientsHolder, ctx clientsholder.Context) *NodeTainted {
+//go:generate moq -out nodetainted_moq.go . TaintedFuncs
+type TaintedFuncs interface {
+	runCommand(cmd string, ctx clientsholder.Context) (string, error)
+	GetKernelTaintInfo(ctx clientsholder.Context) (string, error)
+	GetModulesFromNode(ctx clientsholder.Context) []string
+	ModuleInTree(moduleName string, ctx clientsholder.Context) bool
+	GetOutOfTreeModules(modules []string, ctx clientsholder.Context) []string
+	SetTestingResult(result bool)
```
edcdavid, line 43: In the end, moq is summarizing all (most) ins/outs of a function and collects it in an interface so that it can be extracted out of context and tested independently. I would put all the required functions inside the interface, including the Ginkgo ones. This way, no need for the SetTestingResult function. I don't like that we have conditional unit test code in the runtime since now the unit test is part of the main code as well (so it cannot test itself). Also make getting the env part of the interface

cnf-certification-test/platform/suite_test.go
```diff
@@ -15,3 +15,211 @@
...
+				return tc.modulesFromNodeFuncRet
+			},
+			GetOutOfTreeModulesFunc: func(modules []string, ctx clientsholder.Context) []string {
+				return tc.outOfTreeFuncRet
+			},
+			ModuleInTreeFunc: func(moduleName string, ctx clientsholder.Context) bool {
+				return tc.modulesInTreeFuncRet
+			},
+			SetTestingResultFunc: func(result bool) {
+				sharedResult = result
```
edcdavid, line 214: Rededine ginkgo commands here as well

cnf-certification-test/platform/nodetainted/nodetainted.go
```diff
@@ -29,22 +28,35 @@ import (
...
+	ClientHolder      *clientsholder.ClientsHolder
+	OCPContext        clientsholder.Context
+	unitTestingResult bool
 }
 
-// NewNodeTainted creates a new NodeTainted tnf.Test.
-func NewNodeTaintedTester(timeout time.Duration, client *clientsholder.ClientsHolder, ctx clientsholder.Context) *NodeTainted {
+//go:generate moq -out nodetainted_moq.go . TaintedFuncs
+type TaintedFuncs interface {
+	runCommand(cmd string, ctx clientsholder.Context) (string, error)
```
hamadise, line 38: I believe it would make more sense to have the context as the first argument in all the functions and keep it as that in the rest of the code

cnf-certification-test/platform/suite.go
```diff
@@ -176,8 +183,16 @@ func testTainted(env *provider.TestEnvironment) {
 	// We are expecting tainted nodes to be Nil, but only if:
 	// 1) The reason for the tainted node is contains(`module was loaded`)
 	// 2) The modules loaded are all whitelisted.
-	gomega.Expect(taintedNodes).To(gomega.BeNil())
-	gomega.Expect(errNodes).To(gomega.BeNil())
+	tnf.GomegaExpectSliceBeNil(taintedNodes)
+	tnf.GomegaExpectSliceBeNil(errNodes)
+
+	if tnf.IsUnitTest() {
```
greyerof, line 189: Honestly, I don't like these `if tnf.IsUnitTest() {` blocks for two reasons:
1. It's adding unit-testing related code into the function's normal/production code. I believe this "IsUnitTest()" or whatever variables for controlling the function's execution flow just for the sake of unit-testing should never exist inside the production code. That's precisely one of the advantages of mocking functions: this kind of tricks are not needed.
2. As you're mocking every external function here, it seems more natural to mock the tnf.GinkoFail or the tnf.GomegaExpect* functions to keep track of the failing scenarios of this tested function, as the normal execution flow for a failing tc should "die" on line 186 or 187.

sebrandon1:
Marked WIP again while I address some concerns.

sebrandon1:
Depends on: https://github.com/test-network-function/cnf-certification-test/pull/72 

Waiting on #72 to merge with functions needed per our design conversations.

cnf-certification-test/platform/suite_test.go
```diff
@@ -15,3 +15,226 @@
...
+			expectedGetOutOfTreeModules:     0,
+
+			expectedResult: true,
+		},
+	}
+
+	for _, tc := range testCases {
+		sharedResult := true
+
+		generateEnv := func(acceptedModule string) *provider.TestEnvironment {
```
jmontesi, line 176: I guess this function definition can be hoisted out of the loop as it does not depend on the specific test case.

sebrandon1: Well, the only reason I moved it here was to redefine/reset `sharedResult`.  Since `sharedResult` gets reset to `true` every loop I wanted to be able to set it to false again if needed.

cnf-certification-test/accesscontrol/suite.go
```diff
@@ -133,7 +133,7 @@ func TestSecConCapabilities(env *provider.TestEnvironment) {
 		}
 	}
 	tnf.ClaimFilePrintf("bad containers: %v", badContainers)
-	tnf.GomegaExpectSliceBeNil(badContainers)
+	env.GomegaExpectSliceBeNil(badContainers)
```
hamadise, line 136: one last comment, this should be ExpectSliceBeEmpty 
the slice badContainers is not nil.

hamadise: never mind. just saw the declaration and it's var badContainers []string not :=[]string{}

hamadise (approved):

cnf-certification-test/platform/nodetainted/nodetainted.go
sebrandon1: Done!

Makefile
```diff
@@ -132,4 +132,5 @@ install-moq:
 	go install github.com/matryer/moq@latest
 
 generate: install-moq
+	find . | grep _moq.go | xargs rm
```
sebrandon1, line 135: `moq` isn't smart enough to overwrite the current _moq.go files prior to running `go generate` so this manually clears them out.

edcdavid (approved):

PR #75, opened 2022-03-18
Create more tests for platform/fsdiff test

sebrandon1:
Similar to: #63 and others.

Adding an interface and required testing for the TestFsDiff functionality.


hamadise (approved):

cnf-certification-test/platform/suite_test.go
```diff
@@ -236,3 +238,85 @@ func TestTestTainted(t *testing.T) {
...
+		}
+
+		mockFuncs := &cnffsdiff.FsDiffFuncsMock{
+			RunTestFunc: func(ctx clientsholder.Context) {},
+			GetResultsFunc: func() int {
+				return tc.runTestReturn
+			},
+		}
+
+		testContainersFsDiff(generateEnv(), mockFuncs)
```
edcdavid, line 317: I am having some doubts about writing these suite level unit test cases. Seems to me like a lot of efforts for not too much benefits. The interface makes the code more difficult to understand in my opinion.
For instance, here we just return true/false in the mocked GetResultsFunc and the RunTestFunc does nothing (because it is already tested in the helper test file). This example, might be an extreme case, but in general I would expect the helper to do most of the work and have the test case describe the workflow of the test somaybe less testing needed there?

sebrandon1:
Closing this.  Will re-open after removing suite tests.

PR #77, opened 2022-03-21
Fsdiff refactor; no moq or suite_test.go changes

sebrandon1:
Notice I kept the `testerFuncs cnffsdiff.FsDiffFuncs` parameter in the call to the function, but I removed the `_moq.go` from my original PR #75.  I figure this will provide flexibility but since we aren't worry about testing don't have anything in the suite_test.go.

Also, this func `testContainersFsDiff` is now using direct gomega calls again.


cnf-certification-test/platform/suite.go
```diff
@@ -69,23 +69,18 @@ var _ = ginkgo.Describe(common.PlatformAlterationTestKey, func() {
 })
 
 // testContainersFsDiff test that all CUT didn't install new packages are starting
-func testContainersFsDiff(env *provider.TestEnvironment) {
+func testContainersFsDiff(env *provider.TestEnvironment, testerFuncs cnffsdiff.FsDiffFuncs) {
```
sebrandon1, line 72: Just for note, we decided we didn't want to unit test this func because it's simple enough to read and we wanted to keep the direct calls to gomega.

hamadise (approved):

edcdavid (approved):

PR #81, opened 2022-03-22
Clean up automount test with helper func and tests

sebrandon1:
Using #61 as inspiration, I added all of the logic from the suite.go test into a separate `EvaluateTokens` func that can be more easily tested without modifying any of the suite.go code for tests specifically.  Also it makes the suite.go code easier to read.


hamadise (approved):

cnf-certification-test/accesscontrol/rbac/automount.go
```diff
@@ -45,3 +48,45 @@ func (at *AutomountToken) AutomountServiceAccountSetOnSA(serviceAccountName, pod
 	}
 	return sa.AutomountServiceAccountToken, nil
 }
+
```
edcdavid, line 51: Rename argument "ch" argument to something else, to distinguish from struct "ch" in:
func NewAutomountTokenTester(ch *clientsholder.ClientsHolder) *AutomountToken {
	return &AutomountToken{
		ch: ch,
	}
}

edcdavid (approved):

cnf-certification-test/accesscontrol/rbac/automount.go
sebrandon1: Sounds good to me.  I'll clarify that one.

sebrandon1: Fixed.

PR #82, opened 2022-03-23
add the dumps of the env pramters and the results to the calim.json

aabughosh:


pkg/provider/provider.go
```diff
@@ -50,6 +50,18 @@ const (
 	skipMultusConnectivityTestsLabel = "test-network-function.com/skip_multus_connectivity_tests"
 )
 
+type PodStruct v1.Pod
```
hamadise, line 53: can you try using this 
```
type PodStruct v1.Pod
```
it will simplify the casting you're doing in suite.go

edcdavid: yes, this is needed, otherwise we get the object copy where we need the pointer to the object. Right now the skip functionality no longer works

cnf-certification-test/networking/suite.go
```diff
@@ -153,16 +156,16 @@ func testMultusNetworkConnectivity(env *provider.TestEnvironment, count int, aIP
 	for _, put := range env.Pods {
 		// The first container is used to get the network namespace
 		aContainerInPod := &put.Spec.Containers[0]
-
-		if _, ok := env.SkipNetTests[put]; ok {
+		testpod := provider.PodStruct(*put)
```
edcdavid, line 159: This copy operation creates a new address which will not be matched in the skip object

pkg/provider/provider.go
edcdavid: just tried, seems like we cannot have a pointer or interface type as a method receiver :-(
I think we should just skip these maps

edcdavid: Just tried it, simply using the pointer lead to an error in the following function:
func (p PodStruct) MarshalText() ([]byte, error) {
this is because the receiver object p type should not be a pointer type :-(
I think that we should just skip serializing those 3 maps:
MultusIPs          map[*PodStruct]map[string][]string
	SkipNetTests       map[*PodStruct]bool
	SkipMultusNetTests map[*PodStruct]bool

edcdavid: Got it working! we can use:
type PodStruct struct{
BasePod	*v1.Pod
}

edcdavid (commented):
Access control and certification suites do not have the ReportResult

PR #84, opened 2022-03-23
Add TestPodTainsTolerationIdentifier to check if pod

hamadise:
has any tolerations.
Add UT for the test function

Release Note:
Add TestPodTainsTolerationIdentifier

### DIFF
--- a/cnf-certification-test/lifecycle/suite.go
+++ b/cnf-certification-test/lifecycle/suite.go
@@ -17,6 +17,7 @@
 package lifecycle
 
 import (
+	"errors"
 	"fmt"
 	"time"
 
@@ -59,6 +60,7 @@ var _ = ginkgo.Describe(common.LifecycleTestKey, func() {
 	testContainersLivenessProbe(&env)
 	testPodsOwnerReference(&env)
 	testHighAvailability(&env)
+	testPodTainsToleration(&env)
 
 	testID := identifiers.XformToGinkgoItIdentifier(identifiers.TestPodNodeSelectorAndAffinityBestPractices)
 	ginkgo.It(testID, ginkgo.Label(testID), func() {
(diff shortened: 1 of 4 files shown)


cnf-certification-test/lifecycle/suite.go
```diff
@@ -57,6 +58,7 @@ var _ = ginkgo.Describe(common.LifecycleTestKey, func() {
 	testContainersLivenessProbe(&env)
 	testPodsOwnerReference(&env)
 	testHighAvailability(&env)
+	testPodTainsToleration(&env)
```
sebrandon1, line 61:
```suggestion
	testPodTaintsToleration(&env)
```

sebrandon1 (changes requested):
One typo

cnf-certification-test/lifecycle/suite.go
```diff
@@ -85,13 +87,44 @@ var _ = ginkgo.Describe(common.LifecycleTestKey, func() {
...
 
+func testPodTaintsTolerationHelper(put *v1.Pod) (bool, error) {
+	if put == nil {
+		return false, errors.New("invalid input, pod under test is nil")
+	}
+	for _, t := range put.Spec.Tolerations {
+		if t.Effect == v1.TaintEffectNoSchedule || t.Effect == v1.TaintEffectPreferNoSchedule ||
+			t.Effect == v1.TaintEffectNoExecute {
+			logrus.Error(fmt.Sprintf("pod %s:%s has toleration %s\n", put.Namespace, put.Name, t.Effect))
+			return false, nil
```
greyerof, line 98: You're returning on the first bad toleration found. Unless they're all mutually exclusive, I believe you should iterate through all of them, so you can show all the "bad" ones in the claim file so the customer can fix them altogether. I suggest to change the function to something like:
```
getBadTolerationsTaintEffectFromPod(pod *v1.Pod) []string {
    badEffects := []string{}
    for _, t := range put.Spec.Tolerations {
          if t.Effect == v1.TaintEffectNoSchedule || t.Effect == v1.TaintEffectPreferNoSchedule ||
	          t.Effect == v1.TaintEffectNoExecute {
	          badEffects = append(badEffects , t.Effect)
          }
    }
    return badEffects 
}
```
Then, in the pods loops, you can use check that the output slice is empty:
```
if badEffects := getBadTaintTolerationsEffectsFromPod(put); len(badEffects) > 0 {
    badPods = append...
    tnf.ClaimFilePrintf("%s has invalid taint tolerations effects %v", provider.PodToString(put), badEffects)
}
```

greyerof (commented):
You can use provider.PodToString(put) helper function whenever you want to add the pod's name and namespace to a trace or similar.

cnf-certification-test/lifecycle/suite.go
hamadise: will fix it

PR #93, opened 2022-03-28
Clean up the .golangci config

sebrandon1:
Seeing what I can remove out of this configuration without breaking anything 😄


pkg/tnf/identifier/identifier.go
```diff
@@ -79,7 +79,7 @@ func (i *Identifier) UnmarshalJSON(b []byte) error {
 		return err
 	}
 
-	if err = i.unmarshalURL(objMap); err != nil {
+	if err := i.unmarshalURL(objMap); err != nil {
```
hamadise, line 82: why this change ? err is already declared in line 77. you can just reuse it ?

greyerof (approved):

PR #97, opened 2022-03-29
Removed not needed interface fields of the clientsHolder struct.

greyerof:
All the interfaces for k8s can be found under the same k8sClient, so
there's no need to add separate fields for them. Also, OClient renamed to OcpClient.


edcdavid (approved):

sebrandon1 (approved):
I like it.  👍

PR #100, opened 2022-03-29
Refactoring of the  pods env variable to simplify the code

edcdavid:
Extending the PodWrapper struct to completely replace the pods environment variable and simplify the code. Renaming container's StringShort() stringer to String()


pkg/provider/provider.go
```diff
@@ -51,44 +51,65 @@ const (
 )
 
 type PodWrapper struct {
```
greyerof, line 53: As our custom container wrapper is called simply "Container", cannot we just simply name this one "Pod" ?

greyerof (approved):
Minor detail for all traces: there's no need to explicitly use put/cut.String() for our custom wrappers if they have a String() method implemented, as `String()` is the method of the stringer interface that will be actually used when golang libs need to format any object using "%s" or "%v". So you could do this:
```
logrus.Infof("Test pod: %s - Test container: %s", put, cut)
```

sebrandon1 (approved):

pkg/provider/provider.go
edcdavid: fixed

edcdavid:
> Minor detail for all traces: there's no need to explicitly use put/cut.String() for our custom wrappers if they have a String() method implemented, as `String()` is the method of the stringer interface that will be actually used when golang libs need to format any object using "%s" or "%v". So you could do this:
> 
> ```
> logrus.Infof("Test pod: %s - Test container: %s", put, cut)
> ```

removed all calls to String() in logs, good catch !

PR #107, opened 2022-03-31
Switch back to using singleton clientsHolder

sebrandon1:
Similar to: #81 

Removed all of the remaining structs for access-control's tests that were still building their own structs/copies of the clientsHolder and switched them to using the clientsHolder singleton per suggestion from @greyerof.

### DIFF
--- a/cnf-certification-test/accesscontrol/rbac/clusterrolebinding.go
+++ b/cnf-certification-test/accesscontrol/rbac/clusterrolebinding.go
@@ -24,23 +24,10 @@ import (
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-type ClusterRoleBinding struct {
-	serviceAccountName string
-	podNamespace       string
-	ClientsHolder      *clientsholder.ClientsHolder
-}
-
-func NewClusterRoleBindingTester(serviceAccountName, podNamespace string, ch *clientsholder.ClientsHolder) *ClusterRoleBinding {
-	return &ClusterRoleBinding{
-		serviceAccountName: serviceAccountName,
-		podNamespace:       podNamespace,
-		ClientsHolder:      ch,
-	}
-}
-
-func (crb *ClusterRoleBinding) GetClusterRoleBindings() ([]string, error) {
+func GetClusterRoleBindings(serviceAccountName, podNamespace string) ([]string, error) {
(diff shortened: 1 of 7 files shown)


edcdavid (approved):

cnf-certification-test/accesscontrol/rbac/clusterrolebinding.go
```diff
@@ -24,21 +24,21 @@ import (
...
+	ClientsHolder *clientsholder.ClientsHolder
 }
 
-func NewClusterRoleBindingTester(serviceAccountName, podNamespace string, ch *clientsholder.ClientsHolder) *ClusterRoleBinding {
+func NewClusterRoleBindingTester(ch *clientsholder.ClientsHolder) *ClusterRoleBinding {
 	return &ClusterRoleBinding{
-		serviceAccountName: serviceAccountName,
-		podNamespace:       podNamespace,
-		ClientsHolder:      ch,
+		ClientsHolder: ch,
```
greyerof, line 37: Sorry, Brandon, I missed this one in your previous PRs: why do you need to create this wrapper struct just to save the clientsholder? As it's a singleton, you can just call `clients := clientsholder.GetClientsHolder()` inside the GetClusterRoleBindings() function.
```
func GetClusterRoleBindings(serviceAccountName, podNamespace string) ([]string, error) {
	// Get all of the clusterrolebindings from all namespaces.
	clientsHolder := clientsholder.GetClientsHolder()
	crbList, crbErr := clientsHolder.K8sClient.RbacV1().ClusterRoles().List(context.TODO(), v1.ListOptions{})
```

That way you wouldn't need neither the struct, the NewClusterRoleBindingTester func nor the interface for the actual test case. For the unit test, the singleton will have been mocked, so it should also work fine like this:
```
func TestGetClusterRoleBinding(t *testing.T) {
	_ = clientsholder.GetTestClientsHolder(buildTestObjects())
	gatheredCRBs, err := GetClusterRoleBindings("testCR", "podNS")
	assert.Nil(t, err)
	assert.Equal(t, "testNS:testCR", gatheredCRBs[0])
}
```

sebrandon1: Oh true, I forgot it was a singleton.  I'll re-adjust.  Thanks!

greyerof (approved):

PR #114, opened 2022-04-05
Allow setting the log level with the TNF_LOG_LEVEL env variable

jmontesi:
* So far, the log level was always defaulting to "trace" regardless of
  the env variable value.

* Also changed the env variable name to TNF_LOG_LEVEL from LOG_LEVEL,
  as it seemed a too generic name to be specific for the tnf test suite.


cnf-certification-test/suite_test.go
```diff
@@ -116,9 +116,32 @@ func loadJUnitXMLIntoMap(result map[string]interface{}, junitFilename, key strin
 	}
 }
 
-// SetLogFormat sets the log format for logrus
-func SetLogFormat() {
-	log.Info("debug format initialization: start")
+// logLevel retrieves the TNF_LOG_LEVEL environment variable
+func logLevel() string {
+	logLevel := os.Getenv("TNF_LOG_LEVEL")
```
greyerof, line 121: I think this env var's value was actually read and saved by the configuration.LoadEnvironmentVariables(). I've [refactored ](https://github.com/test-network-function/cnf-certification-test/pull/115/files#diff-6a029f2cd4126be02ace9b8e7e8c25c999fadc6aa23fb78304b414078f0a1ae9)it a little bit in #115, but it's still pending for review.

jmontesi: Done.

greyerof (approved):

greyerof:
Added the "non backward compatible" label due to the change of the name of the LOG_LEVEL environment variable.

PR #115, opened 2022-04-05
Wait for all the debug pods to be ready before autodiscovery starts.

greyerof:
When no ginkgo focus was provided, the diagnostic information retrieval
could start before the debug pods were ready, resulting in errors when
executing the commands inside the debug pods, so I've added the wait
right before the autodiscovery.

Calling WaitDebugPodsReady in the BeforeEach is no longer needed.

Other minor refactors:
- The k8s clients holder is initialized as part of the initial sequence
  of the program.
- Ginkgo/Git info traces improvements.
- Added command to trace in ExecCommandContainer().
- WaitDebugPodsReady refactored to return error.
- Removed explicit function to load/process the environment variables.
  Instead, the package's init() function is used, as they never need to
  be reloaed/reparsed.

### DIFF
--- a/cnf-certification-test/suite_test.go
+++ b/cnf-certification-test/suite_test.go
@@ -32,6 +32,8 @@ import (
 	"github.com/onsi/gomega"
 	log "github.com/sirupsen/logrus"
 	"github.com/test-network-function/cnf-certification-test/cnf-certification-test/results"
+	"github.com/test-network-function/cnf-certification-test/internal/clientsholder"
+	"github.com/test-network-function/cnf-certification-test/pkg/configuration"
 	"github.com/test-network-function/cnf-certification-test/pkg/junit"
 	"github.com/test-network-function/cnf-certification-test/pkg/provider"
 	"github.com/test-network-function/test-network-function-claim/pkg/claim"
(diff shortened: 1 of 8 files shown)


pkg/autodiscover/autodiscover.go
```diff
@@ -90,23 +89,16 @@ func buildLabelKeyValue(label configuration.Label) (key, value string) {
 // DoAutoDiscover finds objects under test
 func DoAutoDiscover() DiscoveredTestData {
 	var err error
-	data.Env, err = configuration.LoadEnvironmentVariables()
+	data.Env = *configuration.GetTestParameters()
 	if err != nil {
 		logrus.Fatalln("can't load environment variable")
```
jmontesi, line 94: The function above does not return error anymore, so err here will always be nil. Remove this log message.

greyerof: I forgot that leftover. Good catch! Done.

cnf-certification-test/suite_test.go
```diff
@@ -150,17 +180,12 @@ func TestTest(t *testing.T) {
...
-	if GitRelease == "" {
-		gitDisplayRelease = "Unreleased build post " + GitPreviousRelease
-	} else {
-		gitDisplayRelease = GitRelease
-	}
-	log.Info("Version: ", gitDisplayRelease, " ( ", GitCommit, " )")
+
 	SetLogFormat()
+
+	// Set clientsholder singletone with the filenames from the env vars.
```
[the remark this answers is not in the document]
greyerof: Fixed in the new commit. Thanks.

jmontesi (approved):

PR #116, opened 2022-04-05
Unit tests for isredhat package

sebrandon1:
Taking advantage of work done previously in #109 for the `Command` interface.

This brings the level of coverage for this package to 100%.

### DIFF
--- a/cnf-certification-test/platform/isredhat/isredhat.go
+++ b/cnf-certification-test/platform/isredhat/isredhat.go
@@ -19,7 +19,6 @@ package isredhat
 import (
 	"errors"
 	"regexp"
-	"time"
 
 	"github.com/sirupsen/logrus"
 	"github.com/test-network-function/cnf-certification-test/internal/clientsholder"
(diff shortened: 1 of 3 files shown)


jmontesi (approved):

cnf-certification-test/platform/isredhat/isredhat.go
```diff
@@ -33,14 +32,12 @@ const (
 )
 
 type BaseImageInfo struct {
-	timeout      time.Duration
-	ClientHolder *clientsholder.ClientsHolder
+	ClientHolder clientsholder.Command
```
greyerof, line 35: ClientHolder seems a bit misleading name for this field, as now it's just the Command interface. What about:
```
CmdExecutor clientsholder.Command
```
Also, the clientholder's "Command" interface name is not following the [go-lang naming convention for interfaces](https://go.dev/doc/effective_go#:~:text=By%20convention%2C%20one%2Dmethod%20interfaces,the%20function%20names%20they%20capture.). I believe we should come up with a better name for that interface like "Commander" or "CmdExecutor", but it can be done in further PRs.

greyerof (approved):
As with #107, I don't think you need to save neither the clientsholder's Command interface nor the context in a custom struct, as you can get/create them on-the-fly. It can be done in further PRs, though.

PR #117, opened 2022-04-05
No need to recommend the timeout value

sebrandon1:
I don't think any of the tests now pass around the `DefaultTimeout` value with the exception of #116 which is removing it.  The README should reflect the best practice now.

### DIFF
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -543,7 +543,7 @@ following:
 
 ```go
 // 1. Create the Test.
-pingTester := ping.NewPing(defaultTimeout, targetPodIPAddress, count)
+pingTester := ping.NewPing(targetPodIPAddress, count)
 test, err := tnf.NewTest(oc.GetExpecter(), pingTester, []reel.Handler{pingTester}, oc.GetErrorChannel())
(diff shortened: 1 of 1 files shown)


greyerof (approved):

DEVELOPING.md
```diff
@@ -543,7 +543,7 @@ following:
 
 ```go
 // 1. Create the Test.
-pingTester := ping.NewPing(defaultTimeout, targetPodIPAddress, count)
+pingTester := ping.NewPing(targetPodIPAddress, count)
```
edcdavid, line 546: The whole file needs an update or be deleted for now. The whole code changed

edcdavid, line 546: The whole file needs an update or be deleted for now. The whole code changed

sebrandon1: Yes, I agree.  Do we want to do that in this PR or a subsequent change?

edcdavid: I would rather just delete this file now and create a new one in a future PR as needed. Current content is not applicable to the new project

sebrandon1: Sounds good, I'll just remove this file instead.

PR #119, opened 2022-04-06
Add online tags to tests requiring internet connectivity

hamadise:


sebrandon1 (approved):
Should we also have an "Offline" label as well?

hamadise:
> Should we also have an "offline" label as well?

there's no need to add offline, online==!offline and we can actually supply !online to the suite.

edcdavid (approved):

PR #122, opened 2022-04-07
Add unit test for diagnostics package

sebrandon1:
Added some more unit test coverage for the `diagnostics` package.  I skipped all of the functions that require a `GetTestEnvironment()` object because at this time I'm not sure how to spoof the singleton.

Notable changes:
- Removed `getLscpu`, `getIPConfig`, and `getLsblk` functions in favor of `getHWJsonOutput`.  These functions were essentially the same thing and needed a single parameter passed into them.
- Utilized changes from #109`clientsHolder.Command` interface for mocking out the `execCommandContainer` function.
- BONUS: Added a test in the `provider` package for `ConvertArrayPods` function.

### DIFF
--- a/pkg/diagnostics/diagnostics.go
+++ b/pkg/diagnostics/diagnostics.go
@@ -96,7 +96,7 @@ func GetHwInfoAllNodes() (out map[string]NodeHwInfo) {
 		if err != nil {
 			logrus.Errorf("problem getting lsblk for node %s", debugPod.Spec.NodeName)
 		}
-		hw.Lspci, err = getLspci(debugPod, o)
+		hw.Lspci, err = getHWTextOutput(debugPod, o, lspciCommand)
 		if err != nil {
 			logrus.Errorf("problem getting lspci for node %s", debugPod.Spec.NodeName)
 		}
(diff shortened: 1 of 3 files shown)


pkg/diagnostics/diagnostics.go
```diff
@@ -148,7 +120,7 @@ func getLsblk(debugPod *v1.Pod, o *clientsholder.ClientsHolder) (out interface{}
 }
 
 // getLspci gets lspci in Json format for a given node
-func getLspci(debugPod *v1.Pod, o *clientsholder.ClientsHolder) (out []string, err error) {
+func getLspci(debugPod *v1.Pod, o clientsholder.Command) (out []string, err error) {
```
edcdavid, line 123: What about renaming getLspci to getHWTextOutput and passing the command to it to follow the pattern for  getHWJsonOutput ? This would be the way to store the output of any command that is not natively outputing Json.

edcdavid (approved):

pkg/diagnostics/diagnostics.go
sebrandon1: That works, I'll do that.

hamadise (approved):

PR #123, opened 2022-04-07
Skipping test cases if resources they rely on are not detected

edcdavid:
Verified each test to make sure that they would skip if not finding a required resources (pods, containers, operators, etc)
Added 2 generic function:
- SkipIfEmptyAny: takes a list of resources to check for and skip the test if any of them are empty
- SkipIfEmptyAll: takes a list of resources to check for and skip the test if all of them are empty

### DIFF
--- a/cnf-certification-test/accesscontrol/suite.go
+++ b/cnf-certification-test/accesscontrol/suite.go
@@ -31,6 +31,7 @@ import (
 	"github.com/test-network-function/cnf-certification-test/internal/clientsholder"
 	"github.com/test-network-function/cnf-certification-test/internal/crclient"
 	"github.com/test-network-function/cnf-certification-test/pkg/provider"
+	"github.com/test-network-function/cnf-certification-test/pkg/testhelper"
 	"github.com/test-network-function/cnf-certification-test/pkg/tnf"
 )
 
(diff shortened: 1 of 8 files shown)


pkg/testhelper/testhelper.go
```diff
@@ -33,3 +38,36 @@ func ResultToString(result int) (str string) {
...
+	}
+}
+
+func SkipIfEmptyAll(skip func(string, ...int), object ...interface{}) {
+	countLenZero := 0
+	allTypes := ""
+	for _, o := range object {
+		s := reflect.ValueOf(o)
+		if s.Kind() != reflect.Slice && s.Kind() != reflect.Map {
+			panic("SkipIfEmpty given a non-slice type")
```
hamadise, line 61: non-slice and non-map.
once we get familiar with 1.18,  we can change this.
you can actually add constraint of the interface

cnf-certification-test/accesscontrol/suite.go
```diff
@@ -56,82 +57,92 @@ var _ = ginkgo.Describe(common.AccessControlTestKey, func() {
 	// Security Context: non-compliant capabilities
 	testID := identifiers.XformToGinkgoItIdentifier(identifiers.TestSecConCapabilitiesIdentifier)
 	ginkgo.It(testID, ginkgo.Label(testID), func() {
+		testhelper.SkipIfEmptyAny(ginkgo.Skip, env.Containers)
```
hamadise, line 60: will not a beforeEach work best in this scenario ?
inside the beforeEach skip if containers are empty ?

pkg/testhelper/testhelper.go
edcdavid: fixed

pkg/testhelper/testhelper.go
```diff
@@ -33,3 +38,36 @@ func ResultToString(result int) (str string) {
...
+
+func SkipIfEmptyAny(skip func(string, ...int), object ...interface{}) {
+	for _, o := range object {
+		s := reflect.ValueOf(o)
+		if s.Kind() != reflect.Slice && s.Kind() != reflect.Map {
+			panic("SkipIfEmpty was given a non slice/map type")
+		}
+
+		if s.Len() == 0 {
+			skip(fmt.Sprintf("Test failed because there are no %s to test, please check under test labels", reflect.TypeOf(o)))
```
greyerof, line 50: All the traces int these two Skip* functions use the`"Test failed ..."` pattern? Shouldn't it be `"Test skipped ..."?`

pkg/testhelper/testhelper.go
```diff
@@ -33,3 +38,36 @@ func ResultToString(result int) (str string) {
...
+func SkipIfEmptyAll(skip func(string, ...int), object ...interface{}) {
+	countLenZero := 0
+	allTypes := ""
+	for _, o := range object {
+		s := reflect.ValueOf(o)
+		if s.Kind() != reflect.Slice && s.Kind() != reflect.Map {
+			panic("SkipIfEmpty was given a non slice/map type")
+		}
+
+		if s.Len() == 0 {
```
greyerof, line 64: Minor suggestion: you can get rid of the "countLenZero" variable if you invert the logic returning as soon as a non-zero length object was found:
```
    ...
    allTypes = allTypes + reflect.TypeOf(o).String() + ", "
    if s.Len() != 0 {
        return
    }
  }
  skip(fmt.Sprintf("Test failed because..."))
}
```

pkg/testhelper/testhelper.go
edcdavid: fixed

pkg/testhelper/testhelper.go
edcdavid: using countLenZero to go through all the resources since the ones that are zero length are listed in the end

greyerof (approved):

PR #124, opened 2022-04-12
Remove DEVELOPING.md references

sebrandon1:
Removing the DEVELOPING.md and references to it.

Originally I was editing the markdown in #117 but we decided to just remove it.


schemas/generic-test.schema.json
```diff
@@ -5,7 +5,7 @@
   "definitions": {
```
greyerof, line 5: Can't we remove this file as well? I don't think it's actually being used in the new arch.

sebrandon1: Yes I'll remove.

sebrandon1: Done.

edcdavid (approved):

PR #125, opened 2022-04-13
Add code to download all certified containers/operators

hamadise:


cmd/tnf/fetch/fetch.go
```diff
@@ -0,0 +1,334 @@
...
+	"time"
+
+	log "github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+	"github.com/test-network-function/cnf-certification-test/internal/registry"
+	"gopkg.in/yaml.v2"
+)
+
+var (
+	filterCsi              = "&filter=csv_description=iregex=CSI;organization==certified-operators"
```
edcdavid, line 18: We should not filter based on CSI:
"&filter=csv_description=iregex=CSI;organization==certified-operators" => "&filter=organization==certified-operators"
CSI is only one type of operator (for storage)

sebrandon1:
@hamadise Might I suggest we create a separate repo for these DB files that we could keep separate until our build process and bring them in? Since we'll be developing in this repo we probably don't want/need to download these db files every time we clone?

cmd/tnf/fetch/fetch.go
```diff
@@ -0,0 +1,334 @@
...
+	defer resp.Body.Close()
+	body, err := io.ReadAll(resp.Body)
+
+	if err != nil {
+		log.Error("Error reading body ", err)
+	}
+	return body
+}
+
+func getCertifiedNumbers() CertifiedCatalog {
```
edcdavid, line 105: Rename to getCertifiedCatalogOnDisk

internal/registry/operator.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+// isOperatorCertified check the presence of operator name in certified operators db
+// the operator name is the csv
+// ocpVersion is Major.Minor OCP version
+func IsOperatorCertified(operatorName, ocpVersion string) bool {
+	name, operatorVersion := extractNameVersionFromName(operatorName)
+	data := OperatorOcpVersionMatch{ocpVersion: ocpVersion, operatorVersion: operatorVersion}
+	if v, ok := operatordb[name]; ok {
+		if ocpVersion == "" {
```
edcdavid, line 107: if ocpVersion=="", this is the non-ocp cluster case, but we still need to check the operator version

hamadise:
> @hamadise Might I suggest we create a separate repo for these DB files that we could keep separate until our build process and bring them in? Since we'll be developing in this repo we probably don't want/need to download these db files every time we clone?
> 
> We could call it something like `test-network-function/offline-dbs`.

you'll have to clone that repo anyway. Right ? those files will not change frequently. so I guess, it's not an issue. @edcdavid WHAT DO YOU THINK ?

edcdavid:
> > @hamadise Might I suggest we create a separate repo for these DB files that we could keep separate until our build process and bring them in? Since we'll be developing in this repo we probably don't want/need to download these db files every time we clone?
> > We could call it something like `test-network-function/offline-dbs`.
> 
> you'll have to clone that repo anyway. Right ? those files will not change frequently. so I guess, it's not an issue. @edcdavid WHAT DO YOU THINK ?

Since we have 4.0 coming soon, I would rather release like this and improve it in the future.

edcdavid (approved):

.github/workflows/update-certification.yml
```diff
@@ -0,0 +1,42 @@
...
+      - name: Disable default go problem matcher
+        run: echo "::remove-matcher owner=go::"
+
+      - name: Check out code into the Go module directory
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ github.sha }}
+
+      - name: Install ginkgo
+        run: go install github.com/onsi/ginkgo/v2/ginkgo@v2.1.3
```
sebrandon1, line 29: #160 already manually bumped this to 2.1.4.  We can either get in a subsequent PR or fix now no big deal.

sebrandon1 (approved):

cmd/tnf/fetch/fetch.go
```diff
@@ -0,0 +1,334 @@
...
+	"time"
+
+	log "github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+	"github.com/test-network-function/cnf-certification-test/internal/registry"
+	"gopkg.in/yaml.v2"
+)
+
+var (
+	filterCsi              = "&filter=organization==certified-operators"
```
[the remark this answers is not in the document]
hamadise: fixed

.github/workflows/update-certification.yml
hamadise: fixed

PR #136, opened 2022-04-20
Check for pods using UID 1337

sebrandon1:
Relates to issue CNFCERT-278.

Makes sure that none of the pods are using `securityContext.RunAsUser` == 1337.

TODO: Still need to find the section in the document for the `BestPracticeReference`.

### DIFF
--- a/cnf-certification-test/accesscontrol/suite.go
+++ b/cnf-certification-test/accesscontrol/suite.go
@@ -138,6 +138,12 @@ var _ = ginkgo.Describe(common.AccessControlTestKey, func() {
 		testhelper.SkipIfEmptyAny(ginkgo.Skip, env.Containers)
 		TestOneProcessPerContainer(&env)
 	})
+	// no 1337 UID's being used by pods
+	testID = identifiers.XformToGinkgoItIdentifier(identifiers.Test1337UIDIdentifier)
+	ginkgo.It(testID, ginkgo.Label(testID), func() {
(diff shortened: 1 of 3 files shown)


sebrandon1:
One question I have about this, is should we also be checking for the `RunAsGroup` to not be using 1337.

cnf-certification-test/accesscontrol/suite.go
```diff
@@ -444,3 +450,21 @@ func TestOneProcessPerContainer(env *provider.TestEnvironment) {
 		ginkgo.Fail(errMsg)
 	}
 }
+
+func Test1337UIDs(env *provider.TestEnvironment) {
+	const leetNum = 1337
+	var badPods []string
+	for _, put := range env.Pods {
+		ginkgo.By(fmt.Sprintf("checking if pod %s has a securityContext RunAsUser 1337 (ns= %s)", put.Data.Name, put.Data.Namespace))
+		if put.Data.Spec.SecurityContext.RunAsUser != nil && *put.Data.Spec.SecurityContext.RunAsUser == int64(leetNum) {
```
jmontesi, line 459: I wonder if it wouldn't be better to check the actual UID of the process launched inside the pod as this _PodSecurityContext_ field can be left unspecified. According to the documentation "defaults to user specified in image metadata if unspecified".

Also, there is a similar field inside _container.securityContext_ that takes precedence over the one in _PodSecurityContext_.

sebrandon1: Cool I'll play around with that.

PR #144, opened 2022-04-22
Use pluto in github actions

sebrandon1:
https://github.com/FairwindsOps/pluto

Following the README from the pluto repo, this checks through any possible k8s manifests for deprecated APIVersions.

image

### DIFF
--- a/.github/workflows/pre-main.yaml
+++ b/.github/workflows/pre-main.yaml
@@ -55,6 +55,16 @@ jobs:
 
       - name: make vet
         run: make vet
+  
+  deprecated-k8s-check:
+    name: Pluto Deprecated Version Check
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Download Pluto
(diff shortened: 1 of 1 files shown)


.github/workflows/pre-main.yaml
```diff
@@ -55,6 +55,16 @@ jobs:
...
+  
+  deprecated-k8s-check:
+    name: Pluto Deprecated Version Check
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Download Pluto
+        uses: FairwindsOps/pluto/github-action@master
+      - name: Use pluto
+        run: |
+          pluto detect-files
```
edcdavid, line 67: Is it useful to run pluto in this repo, since we do not maintain yaml files or helm chart releases?

sebrandon1: Probably not.  I just figured it was a good place to test the integration.  I'll close.

PR #145, opened 2022-04-22
Add tests for autodiscover_crds

sebrandon1:
Switched `APIExtClient` to use `k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset.Interface` as the type which allowed for unit testing using the `fake` clientset for api extension.

Documentation:
https://pkg.go.dev/k8s.io/apiextensions-apiserver@v0.23.6/pkg/client/clientset/clientset/fake


internal/clientsholder/clientsholder.go
```diff
@@ -64,6 +66,13 @@ func GetTestClientsHolder(k8sMockObjects []runtime.Object, filenames ...string)
 	return &clientsHolder
 }
 
+// GetTestClientsHolderAPIExt Overwrites the existing clientholders with a mocked version for unit testing.
+func GetTestClientsHolderAPIExt(k8sMockObjects []runtime.Object, filenames ...string) *ClientsHolder {
```
jmontesi, line 70: The name is a bit misleading as what the function returns is the _clientsHolder_ itself, not the _APIExtClient_. If possible it may be better to unify the definition of all the fake clients for the test _clientsHolder_ in the funcion above (_GetTestClientsHolder()_) as it is done for the real one.

jmontesi: What about passing a slice `mockObjects` with all the resources to the `GetTestClientsHolder()` function and then creating specific slices with the resources required/allowed for each client to be passed to them?

sebrandon1: I'll play around with the `reflect` library to see if there's something I can do there.

sebrandon1: Complete!  I used `reflect` to parse the incoming list of runtime objects and build individual slices for each client expecting them.

internal/clientsholder/clientsholder.go
```diff
@@ -58,7 +63,41 @@ func SetupFakeOlmClient(olmMockObjects []runtime.Object) {
 // For other (OLM, )
 // runtime mocking objects loading, use the proper clientset mocking function.
 func GetTestClientsHolder(k8sMockObjects []runtime.Object, filenames ...string) *ClientsHolder {
-	clientsHolder.K8sClient = k8sFakeClient.NewSimpleClientset(k8sMockObjects...)
+	// Build slices of different objects depending on what client
```
jmontesi, line 66: Looks good! Would it be possible to also include the setup of the `clientsHolder.OlmClient`? This way we can also get rid of the `SetupFakeOlmClient()` function.

sebrandon1: I can do this in a separate PR to keep the changes separate.

edcdavid (approved):
Nice!

hamadise (approved):

PR #148, opened 2022-04-25
Removing duplicate test suite name at the beginning of the test in claims file

edcdavid:
for instance:
platform-alteration-platform-alteration-isredhat-release -> platform-alteration-isredhat-release


sebrandon1 (approved):

greyerof (approved):
You should add the "non backward compatible" label to this PR, as this is changing the behavior of the claim/junit file.

PR #161, opened 2022-04-28
Fix High Availability Test

hamadise:
In 3.3.0, to pass the high availability check, the CNF should define pod anti affinity to avoid deploying Pods in the same node.
The number of replica should be higher than 1 as well.
In the current code, we check for replicas number only, this PT fixes that.


sebrandon1:
Can you add more description to this?

sebrandon1 (approved):

edcdavid (approved):

PR #164, opened 2022-04-29
wait for pod to be deleted to complete the drain

edcdavid:
Added code to wait for the pods to be completely deleted during the pod recreation node drain. Otherwise the deployment/statefulset api will not pickup the that the pods have been deleted.
Used mutithreaded approach to speed up deletion

### DIFF
--- a/cnf-certification-test/lifecycle/podrecreation/podrecreation.go
+++ b/cnf-certification-test/lifecycle/podrecreation/podrecreation.go
@@ -39,6 +39,9 @@ const (
 	DefaultGracePeriodInSeconds = 30
 	Cordon                      = "cordon"
 	Uncordon                    = "uncordon"
+	DeleteBackground            = "deleteBackground"
+	DeleteForeground            = "deleteForeground"
+	NoDelete                    = "noDelete"
 )
 
 func CordonHelper(name, operation string) error {
@@ -68,7 +71,7 @@ func CordonHelper(name, operation string) error {
 	return retryErr
 }
 
-func CountPodsWithDelete(nodeName string, isDelete, waitForDeletion bool) (count int, err error) {
+func CountPodsWithDelete(nodeName, mode string) (count int, err error) {
 	clients := clientsholder.GetClientsHolder()
 	pods, err := clients.K8sClient.CoreV1().Pods("").List(context.TODO(), metav1.ListOptions{
 		FieldSelector: "spec.nodeName=" + nodeName, LabelSelector: "pod-template-hash",
@@ -84,10 +87,10 @@ func CountPodsWithDelete(nodeName string, isDelete, waitForDeletion bool) (count
 			continue
 		}
 		count++
-		if !isDelete {
+		if mode == NoDelete {
 			continue
 		}
-		err := deletePod(&pods.Items[idx], waitForDeletion, &wg)
+		err := deletePod(&pods.Items[idx], mode, &wg)
 		if err != nil {
 			logrus.Errorf("error deleting %s", &pods.Items[idx])
 		}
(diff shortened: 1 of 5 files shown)


cnf-certification-test/lifecycle/podrecreation/podrecreation.go
```diff
@@ -75,33 +78,89 @@ func CountPodsWithDelete(nodeName string, isDelete bool) (count int, err error)
...
+		return err
+	}
+	if !waitForDeletion {
+		return nil
+	}
+	wg.Add(1)
+	podName := pod.Name
+	namespace := pod.Namespace
+	go func() {
+		time.Sleep(1 * time.Second)
```
greyerof, line 127: Why is this sleep needed here?

cnf-certification-test/lifecycle/podrecreation/podrecreation.go
```diff
@@ -75,33 +78,89 @@ func CountPodsWithDelete(nodeName string, isDelete bool) (count int, err error)
...
+		FieldSelector: "metadata.name=" + podName + ",metadata.namespace=" + ns,
+	})
+	defer watcher.Stop()
+	if err != nil {
+		return fmt.Errorf("waitPodDeleted ns=%s pod=%s, err=%s", ns, podName, err)
+	}
+	for {
+		select {
+		case event := <-watcher.ResultChan():
+			if event.Type == watch.Deleted || event.Type == "" {
```
greyerof, line 157: I like this new way to detect the deleted pods but... what if the cluster managed to re-deploy the pod so fast that this watcher is linked to the newly created pod instead of the old one? In that case, you will never receive the "Deleted" event, right? I think we should create the watcher before deleting the pod in the `deletePod()` function and pass it to this `waitPodDeleted` function.

cnf-certification-test/lifecycle/podrecreation/podrecreation.go
```diff
@@ -75,33 +78,89 @@ func CountPodsWithDelete(nodeName string, isDelete bool) (count int, err error)
...
+		}
+	}
+	return false
+}
+
+func deletePod(pod *v1.Pod, waitForDeletion bool, wg *sync.WaitGroup) error {
+	clients := clientsholder.GetClientsHolder()
+	logrus.Tracef("deleting %s", pod.String())
+	deleteOptions := metav1.DeleteOptions{}
+	gracePeriodSeconds := int64(DefaultGracePeriodInSeconds + time.Duration(*pod.Spec.TerminationGracePeriodSeconds))
```
greyerof, line 112: Maybe I missed this one from previous reviews but, why are you adding this DefaultGracePeriodInSeconds to the  pod's one? I believe doing this should be enough:
```
gracePeriodSeconds := time.Seconds * time.Duration(*pod.Spec.TerminationGracePeriodSeconds)
```

greyerof (commented):
I love this solution using k8s' watchers, good job :)

cnf-certification-test/lifecycle/suite.go
```diff
@@ -335,13 +335,13 @@ func testPodsRecreation(env *provider.TestEnvironment) { //nolint:funlen
...
 		logrus.Debugf("node: %s cordoned", n)
-		count, err := podrecreation.CountPodsWithDelete(n, false)
+		count, err := podrecreation.CountPodsWithDelete(n, false, true)
 		if err != nil {
 			ginkgo.Fail(fmt.Sprintf("Getting pods list to drain in node %s failed with err: %s. Test inconclusive.", n, err))
 		}
 		nodeTimeout := timeoutPodSetReady + timeoutPodRecreationPerPod*time.Duration(count)
 		logrus.Debugf("draining node: %s with timeout: %s", n, nodeTimeout.String())
-		_, err = podrecreation.CountPodsWithDelete(n, true)
+		_, err = podrecreation.CountPodsWithDelete(n, true, true)
```
jmontesi, line 344: Same here

cnf-certification-test/lifecycle/podrecreation/podrecreation.go
edcdavid: nope, it was from previous testing. Removed

cnf-certification-test/lifecycle/podrecreation/podrecreation.go
edcdavid: fixed. Actually I did it like this before while testing but I thought it was not needed. I think it makes sense. Fixed it.

cnf-certification-test/lifecycle/suite.go
```diff
@@ -335,13 +335,13 @@ func testPodsRecreation(env *provider.TestEnvironment) { //nolint:funlen
 		}
 		ginkgo.By(fmt.Sprintf("Draining and Cordoning node %s: ", n))
 		logrus.Debugf("node: %s cordoned", n)
-		count, err := podrecreation.CountPodsWithDelete(n, false)
+		count, err := podrecreation.CountPodsWithDelete(n, false, true)
```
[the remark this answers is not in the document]
edcdavid: fixed

cnf-certification-test/lifecycle/suite.go
edcdavid: fixed

sebrandon1 (approved):

cnf-certification-test/lifecycle/podrecreation/podrecreation.go
greyerof: I still see the sleep here :o

cnf-certification-test/lifecycle/podrecreation/podrecreation.go
```diff
@@ -65,7 +71,7 @@ func CordonHelper(name, operation string) error {
 	return retryErr
 }
 
-func CountPodsWithDelete(nodeName string, isDelete bool) (count int, err error) {
+func CountPodsWithDelete(nodeName, mode string) (count int, err error) {
```
greyerof, line 74: I believe this function is doing too many things, which ends up with a bit convoluted code and some code for UT only (DeleteBackground mode). I suggest breaking it down into two different but simpler functions to be called separately in the test case:
```
GetPodsToDeleteFromNode(node string) ([]*v1.Pods, error)
DeletePods([]v1.Pods, watchForDeletion bool) error
```
As this is just a refactor suggestion, this could be done in further PRs, though.

greyerof (approved):

cnf-certification-test/lifecycle/podrecreation/podrecreation.go
edcdavid: ok will refactor this in another PR

PR #165, opened 2022-04-29
Add tests for SkipIfEmpty funcs

sebrandon1:
New coverage level: `coverage: 90.9% of statements`.

To be honest these two functions seem very similar and there might be room to merge the two in a subsequent PR.

### DIFF
--- a/pkg/testhelper/testhelper_test.go
+++ b/pkg/testhelper/testhelper_test.go
@@ -56,6 +56,10 @@ func TestSkipIfEmptyFuncs(t *testing.T) {
 			objects: []string{"test"},
 			skipped: false,
 		},
+		{ // Test Case #4 - Multiple objects
+			objects: []string{"test1", "test2"},
+			skipped: false,
+		},
 		// Note: Cannot test calls to panic
 	}
 
--- a/pkg/testhelper/testhelper_test.go
+++ b/pkg/testhelper/testhelper_test.go
@@ -17,6 +17,7 @@
 package testhelper
 
 import (
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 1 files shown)


pkg/testhelper/testhelper_test.go
```diff
@@ -37,3 +38,47 @@ func TestResultToString(t *testing.T) {
...
+		{ // Test Case #1 - Skip because objects is empty, no panic because []string type
+			objects: []string{},
+			skipped: true,
+		},
+		{ // Test Case #2 - Skip because objects is empty, no panic because map type
+			objects: make(map[string]string),
+			skipped: true,
+		},
+		{ // Test Case #3 - No skip because objects is populated, no panic because []string type
+			objects: []string{"test"},
```
edcdavid, line 56: add a TCs with multiple objects

sebrandon1: Done.

edcdavid: Example single test for multiple objects (just to explain what I meant):
func TestSkipIfEmptyFuncs(t *testing.T) {
	testCases := []struct {
		map1, slice1 interface{}
		skippedAll, skippedAny bool
	}{
		{ // Test Case #4 - Multiple objects
			map1: []string{"test1", "test2"},
			slice1: make(map[string]string),
			skippedAny: true,
			skippedAll: false,
		},
		// Note: Cannot test calls to panic
	}

	for _, tc := range testCases {
		result := false

		SkipIfEmptyAll(func(s string, i ...int) {
			if strings.Contains(s, "Test skipped") {
				result = true
			} else {
				result = false
			}
		}, tc.map1, tc.slice1)
		assert.Equal(t, tc.skippedAll, result)

		SkipIfEmptyAny(func(s string, i ...int) {
			if strings.Contains(s, "Test skipped") {
				result = true
			} else {
				result = false
			}
		}, tc.map1, tc.slice1)

		assert.Equal(t, tc.skippedAny, result)
	}
}

sebrandon1: Oh I see what you are saying.  I'll update!

edcdavid (approved):

jmontesi (approved):

PR #80, opened 2022-03-22
Push more logic into help funcs for TestNamespace

sebrandon1:
As we are trying to "simplify" the suite.go's in our test suite I started looking for loops that could be unit tested.

PR #170, opened 2022-05-03
Fix claim file log

sebrandon1:
#80 accidentally broke the way the CI was checking if this test failed.  Using the `loghelper.CuratedLogLines` this will restore this functionality.

### DIFF
--- a/cnf-certification-test/accesscontrol/suite.go
+++ b/cnf-certification-test/accesscontrol/suite.go
@@ -294,9 +294,9 @@ func TestNamespace(env *provider.TestEnvironment) {
 	}
 
 	invalidCrsNum, claimsLog := namespace.GetInvalidCRsNum(invalidCrs)
-	tnf.ClaimFilePrintf("%s", claimsLog)
 	if invalidCrsNum > 0 {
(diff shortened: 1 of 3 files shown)


edcdavid (approved):

cnf-certification-test/accesscontrol/suite.go
```diff
@@ -293,7 +293,8 @@ func TestNamespace(env *provider.TestEnvironment) {
 		ginkgo.Fail("error retrieving CRs")
 	}
 
-	invalidCrsNum := namespace.GetInvalidCRsNum(invalidCrs)
+	invalidCrsNum, claimsLog := namespace.GetInvalidCRsNum(invalidCrs)
+	tnf.ClaimFilePrintf("%s", claimsLog)
```
greyerof, line 297: Normally, the claim file log output is only helpful when the tc fails. I'd move this printf inside the `if invalidCrsNum > 0` block.

greyerof (approved):

cnf-certification-test/accesscontrol/suite.go
sebrandon1: Fixed!

PR #172, opened 2022-05-03
Use strings.Join

sebrandon1:
Found some code where we could utilize `strings.Join` to build our badContainers/badPods output.


hamadise (approved):

hamadise:
it's failing Lint. goLint prefer the old approach. I think this is better tbh.

sebrandon1:
Hmmm it only seems to not like places where there are two items to be concatenated.  The places where it is >2 it seems to like it.

cnf-certification-test/lifecycle/suite.go
```diff
@@ -113,8 +114,7 @@ func testContainersImagePolicy(env *provider.TestEnvironment) {
 		for _, cut := range env.Containers {
 			logrus.Debugln("check container ", cut.Namespace, " ", cut.Podname, " ", cut.Data.Name, " pull policy, should be ", v1.PullIfNotPresent)
 			if cut.Data.ImagePullPolicy != v1.PullIfNotPresent {
-				s := cut.Namespace + ":" + cut.Podname + ":" + cut.Data.Name
-				badcontainers = append(badcontainers, s)
+				badcontainers = append(badcontainers, strings.Join([]string{cut.Namespace, cut.Podname, cut.Data.Name}, ":"))
```
edcdavid, line 117: Shouldn't we use cut.String() instead? same comment for the logrus lines for instance @ line 115
badcontainers = append(badcontainers, cut.String())

sebrandon1: Yes that works too.  I'll make this use cut.String() instead.

cnf-certification-test/lifecycle/suite.go
```diff
@@ -178,8 +175,7 @@ func testPodsOwnerReference(env *provider.TestEnvironment) {
 			o := ownerreference.NewOwnerReference(put.Data)
 			o.RunTest()
 			if o.GetResults() != testhelper.SUCCESS {
-				s := put.Data.Namespace + ":" + put.Data.Name
-				badPods = append(badPods, s)
+				badPods = append(badPods, put.Data.Namespace+":"+put.Data.Name)
```
edcdavid, line 178: use put.String() here as well

edcdavid (approved):

PR #188, opened 2022-05-06
Make suite.go more consistent

sebrandon1:
The diff makes this look more complicated than it is.  All I am doing is moving the:

```
testID := identifiers.XformToGinkgoItIdentifier(identifiers.XXX)
	ginkgo.It(testID, ginkgo.Label(Online, testID), func() {
```

To wrap the calling functions like it is done in other test suites.  I also moved the `testhelper` skip checks out of the func if applicable.

### DIFF
--- a/cnf-certification-test/certification/suite.go
+++ b/cnf-certification-test/certification/suite.go
@@ -49,9 +49,22 @@ var _ = ginkgo.Describe(common.AffiliatedCertTestKey, func() {
 	})
 	ginkgo.ReportAfterEach(results.RecordResult)
 
-	testContainerCertificationStatus(&env)
-	testAllOperatorCertified(&env)
-	testHelmCertified(&env)
+	// Query API for certification status of listed containers
+	testID := identifiers.XformToGinkgoItIdentifier(identifiers.TestContainerIsCertifiedIdentifier)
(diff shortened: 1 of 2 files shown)


cnf-certification-test/operator/suite.go
```diff
@@ -42,24 +42,22 @@ var _ = ginkgo.Describe(common.OperatorTestKey, func() {
 	var env provider.TestEnvironment
 	ginkgo.BeforeEach(func() {
 		env = provider.GetTestEnvironment()
+		testhelper.SkipIfEmptyAny(ginkgo.Skip, env.Operators)
```
edcdavid, line 45: I would revert the change here and add this under each it statement. Otherwise, just modifying one TC will require to check all others (when adding new TCs or modifying). It is way simpler to make the skip relevent to a single TC always.

sebrandon1: Yeah I was 50/50 about moving this originally.  Sounds good I'll switch it back.

sebrandon1: Fixed!

edcdavid (approved):

hamadise (approved):

PR #191, opened 2022-05-06
Add autodiscover unit tests

sebrandon1:
`coverage: 20.8% of statements`

I didn't include tests for `getOpenshiftVersion` or `DoAutoDiscover` in this.

### DIFF
--- a/pkg/autodiscover/autodiscover_test.go
+++ b/pkg/autodiscover/autodiscover_test.go
@@ -15,3 +15,171 @@
 // 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 
 package autodiscover
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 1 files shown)


pkg/autodiscover/autodiscover_test.go
```diff
@@ -15,3 +15,171 @@
...
+		{
+			testList:       []configuration.Namespace{},
+			expectedOutput: nil,
+		},
+		{
+			testList: []configuration.Namespace{
+				{Name: "name1"},
+				{Name: "name1"},
+			},
+			expectedOutput: []string{"name1", "name1"},
```
jmontesi, line 178: Isn't this test case equivalent to the first one?

jmontesi (approved):

pkg/autodiscover/autodiscover_test.go
sebrandon1: I was trying to do two different and two same strings for the different testcases.

hamadise (approved):

PR #192, opened 2022-05-06
Update code for fetching catalog

hamadise:
Add UT for operator certification
Update DB with the latest catalog


edcdavid (approved):

cmd/tnf/fetch/fetch.go
```diff
@@ -156,7 +157,7 @@ func getOperatorCatalogSize() (size, pagesize uint) {
 	if err != nil {
 		log.Fatalf("Error in unmarshaling body: %v", err)
 	}
-	return aCatalog.Total, aCatalog.PageSize
+	return aCatalog.Total, operatorCatalogPageSize
 }
 
 func getOperatorCatalogPage(page, size uint) {
```
greyerof, line 163: You are not using the "size" value in the URL for the http GET, so you'll have only the default size (100) on each page, instead of the expected (500). Line 169 should be:
```
	url := fmt.Sprintf("%spage=%d&page_size=%d%s", operatorcatalogURL, page, size, filterCertifiedOperators)
```
Also, adding a log line would be nice, like:
```
	log.Infof("Getting operators page %d (size %d), url: %s", page, size, url)
```

internal/registry/operator_test.go
```diff
@@ -24,8 +24,8 @@ import (
 )
 
 func TestIsOperatorCertified(t *testing.T) {
-	name := "zoperator.v0.3.6"
```
greyerof, line 27: zoperator can be still used, as it will appear in the catalog as soon the api is fixed, so this change is not needed.

hamadise:
@greyerof can you approve this ?

greyerof (approved):

PR #194, opened 2022-05-09
Workaround for operators catalog downloader.

greyerof:
The previous way to get operators catalog pages wasn't retrieving
all the pages. It can be tested manually in the browser that there are
many more pages than the calculated ones.

As we cannot rely on the total value to calculate the number of
operator pages, the only option is reading until we found a page without any
operator.

Includes index file renaming and page_size increase to 500 from #192


sebrandon1 (approved):
Approved but this should probably be some sort of archive/tarball.

greyerof:
This is not a fix, but a workaround for the bug in the catalog api.

greyerof:
Closing this PR as it will not be needed.

PR #203, opened 2022-05-11
Surface more information about tainted kernel failures

sebrandon1:
If there are "other" tainted kernel failures than the `module was loaded` failures, report them as part of the claim file printout.

### DIFF
--- a/cnf-certification-test/platform/suite.go
+++ b/cnf-certification-test/platform/suite.go
@@ -199,9 +199,11 @@ func testTainted(env *provider.TestEnvironment, testerFuncs nodetainted.TaintedF
 			nodeTaintsAccepted = false
 
 			// Surface more information about tainted kernel failures that have nothing to do with modules.
-			tnf.ClaimFilePrintf("Please note that taints other than 'module was loaded' were found.")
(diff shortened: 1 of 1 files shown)


hamadise (approved):

cnf-certification-test/platform/suite.go
```diff
@@ -185,16 +185,24 @@ func testTainted(env *provider.TestEnvironment, testerFuncs nodetainted.TaintedF
...
 				otherTaintsFound = true
+				otherTaints = append(otherTaints, it)
 			}
 		}
 
 		if otherTaintsFound {
 			nodeTaintsAccepted = false
+
+			// Surface more information about tainted kernel failures that have nothing to do with modules.
+			tnf.ClaimFilePrintf("Please note that taints other than 'module was loaded' were found.")
```
edcdavid, line 202: also add a similar line in the claim file for the tainted modules, see line 213:
logrus.Debug("Collected all of the tainted modules: ", taintedModules)

greyerof: I've run manually this test case and I saw we're not printing the name of the node we're testing... may you add it to the traces, please? I don't think the name of the debug pod is actually helpful for the customer.

greyerof (approved):

PR #207, opened 2022-05-12
Enhanced catalog display

jmontesi:
Porting of PR#561, 582 and 586 from the test-network-function repo.


cmd/tnf/generate/catalog/catalog.go
```diff
@@ -159,10 +238,13 @@ func outputTestCaseBuildingBlocks() {
 
 	// Iterating the map by sorted identifier URL
 	for _, k := range keys {
```
greyerof, line 240: You can remove this outputTestCaseBuildingBlocks() entirely.

greyerof (commented):
Please @hamadise , @edcdavid , @sebrandon1 , correct me if I'm wrong, but this building blocks sections in the catalog does not apply any more in the new repo, as it only had a meaning when we used the "handlers" in the old repo. The entire `pkg/tnf/identifier/identifiers.go` can be removed safely, along with the related functions/parts in the catalog generator tool that used it.

sebrandon1:
I believe you are correct @greyerof

greyerof (approved):

sebrandon1 (approved):

PR #223, opened 2022-05-19
Consolidate release logic

sebrandon1:
The functions `CompareVersion` and `IsReleaseCertified` were duplicated in the code so I created the `releasehelper` package to hold the `IsReleaseCertified` func.  `CompareVersion` now lives in the `stringhelper` package since it was just comparing some strings but I could be persuaded to move it elsewhere.

The struct `ChartStruct` already existed in `internal/api/api.go` so I just left it there as well.

### DIFF
--- a/cnf-certification-test/certification/certtool/certtool.go
+++ b/cnf-certification-test/certification/certtool/certtool.go
@@ -17,14 +17,11 @@
 package certtool
 
 import (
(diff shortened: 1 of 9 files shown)


sebrandon1:
@hamadise will be fixing the code to actually use `chartsdb`.

PR #232, opened 2022-05-26
initial commit to normalize offline and online code

hamadise:


internal/api/onlinecheck/onlinecheck.go
```diff
@@ -0,0 +1,317 @@
...
+	Client HTTPClient
+}
+
+// NewOnlineValidator return an online implementation of the CertificationValidator interface
+func NewOnlineValidator() OnlineValidator {
+	return OnlineValidator{Client: &http.Client{}}
+}
+
+// IsServiceReachable check if redhat catalog is reachable and its database is available to query
+func (checker OnlineValidator) IsServiceReachable() bool {
```
edcdavid, line 74: not required: not sure what the is the best. Do the check every time or just once during discovery ? Might be better to run all the tests with the same config and save it in the claim.yaml with the rest of the config

internal/api/offlinecheck/container.go
```diff
@@ -89,7 +89,7 @@ func LoadBinary(bytes []byte, db map[string]*ContainerCatalogEntry) {
 	}
 }
 
-func IsCertified(registry, repository, tag, digest string) bool {
+func (checker OfflineChecker) IsContainerCertified(registry, repository, tag, digest string) bool {
```
sebrandon1, line 92: You don't use `checker` here.  It probably doesn't need to be a function implemented in the struct?

internal/api/offlinecheck/helm.go
```diff
@@ -72,7 +92,7 @@ func CompareVersion(ver1, ver2 string) bool {
 	return false
 }
 
-func IsReleaseCertified(helm *release.Release, ourKubeVersion string) bool {
+func (checker OfflineChecker) IsReleaseCertified(helm *release.Release, ourKubeVersion string) bool {
```
sebrandon1, line 95: Same here `checker` isn't used.

internal/api/offlinecheck/operator.go
```diff
@@ -120,8 +120,8 @@ func loadOperatorsCatalog(pathToRoot string) {
 // isOperatorCertified check the presence of operator name in certified operators db
 // the operator name is the csv
 // ocpVersion is Major.Minor OCP version
-func IsOperatorCertified(operatorName, ocpVersion, channel string) bool {
-	name, operatorVersion := extractNameVersionFromName(operatorName)
+func (checker OfflineChecker) IsOperatorCertified(operatorName, ocpVersion, channel string) bool {
```
sebrandon1, line 123: Same here `checker` isn't used.

internal/api/offlinecheck/operator_test.go
```diff
@@ -33,10 +34,10 @@ func TestIsOperatorCertified(t *testing.T) {
 	if err != nil {
 		log.Println(err)
 	}
-	loadOperatorsCatalog(path + "/../")
-	ans := IsOperatorCertified(name, ocpversion, channel)
+	loadOperatorsCatalog(path + "/../../")
+	ans := checker.IsOperatorCertified(name, ocpversion, channel)
 	assert.Equal(t, ans, true)
```
sebrandon1, line 39:
```suggestion
	assert.True(t, checker.IsOperatorCertified(name, ocpversion, channel))
```

sebrandon1 (changes requested):
Just some initial notes.

internal/api/offlinecheck/container.go
hamadise: in offline one yes. but in the online version I do need it.

internal/api/offlinecheck/helm.go
hamadise: yep, there's no need for checker in offline mode. To use it, I need to remove registry package and replace it with this one

edcdavid (approved):

internal/api/offlinecheck/container.go
sebrandon1: I don't understand?

sebrandon1: Okay I think I understand now.  It's just organizational.  Sounds good.

sebrandon1 (approved):

PR #242, opened 2022-05-30
Fix for platform-alteration-base-image test case.

greyerof:
The current podman diff --json output won't match the expected regex, as
the output won't come as in the original UT, which added indentation
before the folder and it was expected by the regex.

The actual output currently comes as a single line, like this:
{"changed":["/root","/usr","/usr/lib"],"added":["/root/.bash_history","/usr/lib/blablah"]}

I've changed the implementation to unmarshall the output in order to be more
clear and verbose in the log/claim file when the test case fails.

Also, added lib64 folders to be tested.

### DIFF
--- a/cnf-certification-test/platform/cnffsdiff/fsdiff.go
+++ b/cnf-certification-test/platform/cnffsdiff/fsdiff.go
@@ -17,28 +17,48 @@
 package cnffsdiff
 
 import (
+	"encoding/json"
 	"fmt"
-	"regexp"
 
 	"github.com/sirupsen/logrus"
 	"github.com/test-network-function/cnf-certification-test/internal/clientsholder"
 	"github.com/test-network-function/cnf-certification-test/pkg/testhelper"
 )
 
-const (
-	varlibrpm  = `(?m)[\t|\s]\/var\/lib\/rpm[.]*`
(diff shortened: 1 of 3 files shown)


cnf-certification-test/platform/cnffsdiff/fsdiff.go
```diff
@@ -53,32 +73,52 @@ func NewFsDiffTester(client clientsholder.Command) *FsDiff {
...
+	// Check for deleted folders.
+	for _, deletedFolder := range diff.Deleted {
+		if _, exist := targetFolders[deletedFolder]; exist {
+			logrus.Tracef("Container's folder %s has been deleted.", deletedFolder)
+			f.DeletedFolders = append(f.DeletedFolders, deletedFolder)
+		}
+	}
+
+	// Check for changed folders.
+	for _, changedFolder := range diff.Changed {
```
hamadise, line 110: you should check if any of the target folders is a prefix of the changed folder.
All our target folder were terminated by * before

hamadise: I mean what if a user change something under /bin/go
then the modified folder would be '/bin/go' which is not in our list.

sebrandon1 (approved):

PR #260, opened 2022-06-08
Remove gomega; add more context to failures

sebrandon1:
Instead of seeing a failure like:

```
tnf.ClaimFilePrintf("bad containers: %v", badContainers)
gomega.Expect(badContainers).To(gomega.BeNil())
```

I wanted to add more context to the failures instead of just seeing a list of "bad containers" in DCI output.

### DIFF
--- a/cnf-certification-test/accesscontrol/suite.go
+++ b/cnf-certification-test/accesscontrol/suite.go
@@ -21,7 +21,6 @@ import (
 	"strings"
 
 	"github.com/onsi/ginkgo/v2"
-	"github.com/onsi/gomega"
 	"github.com/sirupsen/logrus"
 	"github.com/test-network-function/cnf-certification-test/cnf-certification-test/accesscontrol/namespace"
 	"github.com/test-network-function/cnf-certification-test/cnf-certification-test/accesscontrol/rbac"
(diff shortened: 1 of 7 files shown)


edcdavid (approved):

cnf-certification-test/accesscontrol/suite.go
```diff
@@ -384,7 +410,9 @@ func TestAutomountServiceToken(env *provider.TestEnvironment) {
 	failedPods := []string{}
 	for _, put := range env.Pods {
 		ginkgo.By(fmt.Sprintf("check the existence of pod service account %s (ns= %s )", put.Data.Namespace, put.Data.Name))
-		gomega.Expect(put.Data.Spec.ServiceAccountName).ToNot(gomega.BeEmpty())
+		if put.Data.Spec.ServiceAccountName == "" {
```
hamadise, line 413: can you add this to tnf file as well ?

hamadise (approved):

cnf-certification-test/accesscontrol/suite.go
```diff
@@ -153,8 +152,11 @@ func TestSecConCapabilities(env *provider.TestEnvironment) {
 			}
 		}
 	}
-	tnf.ClaimFilePrintf("bad containers: %v", badContainers)
-	gomega.Expect(badContainers).To(gomega.BeNil())
```
greyerof, line 157: Although I prefer your approach (without gomega), it is worth noting that gomega assertions can also accept a text that will be printed only if the assertion fails. In this case it could be like this:
```
tnf.ClaimFilePrintf("bad containers: %v", badContainers)
gomega.Expect(badContainers).To(gomega.BeNil(), "Containers were found with incorrect security context capabilities.")
```

greyerof (approved):

cnf-certification-test/accesscontrol/suite.go
sebrandon1: Done.

PR #266, opened 2022-06-09
Add safeguard for SNO clusters

sebrandon1:
Adding a safeguard against running the pod-recreation test case against a cluster with less than two workers.

Bonus, I also added the same check around the `testDeploymentScaling` and `testStatefulSetScaling` tests because even if we are attempting to scale these up/down, if they are properly using anti-affinity rules they'll never scale correctly anyways.

### DIFF
--- a/cnf-certification-test/lifecycle/suite.go
+++ b/cnf-certification-test/lifecycle/suite.go
@@ -39,7 +39,7 @@ const (
 	timeout                    = 60 * time.Second
 	timeoutPodRecreationPerPod = time.Minute
 	timeoutPodSetReady         = 7 * time.Minute
-	minNodesForLifecycle       = 2
+	minWorkerNodesForLifecycle = 2
 )
 
 //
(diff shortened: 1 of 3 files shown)


cnf-certification-test/lifecycle/suite.go
```diff
@@ -39,6 +39,7 @@ const (
 	timeout                    = 60 * time.Second
 	timeoutPodRecreationPerPod = time.Minute
 	timeoutPodSetReady         = 7 * time.Minute
+	minNodesForLifecycle       = 2
```
greyerof, line 42: Shouldn't this be min**Worker**NodesForLifecycle instead? Clusters with one master and one worker node would also make these TCs pointless. We should implement a way to count the worker nodes of env.Nodes, is it contains both types of nodes right now.

sebrandon1: Good idea.  I'll see if I can separate that out.

greyerof (approved):

jmontesi (approved):

PR #269, opened 2022-06-10
Add nil check for deployment object

sebrandon1:
Trace from a recent DCI run:

```
[37mTRACE  [0m[Jun  9 21:19:22.820][podsets.go: 35] check if deployment cert-manager:cert-manager is ready  
[31mERROR  [0m[Jun  9 21:21:12.893][autodiscover_podset.go: 34] Can't retrieve deployment in ns=cert-manager name=cert-manager 
[31mERROR  [0m[Jun  9 21:21:12.893][autodiscover_podset.go: 34] Can't retrieve deployment in ns=cert-manager name=cert-manager 
[38;5;243m------------------------------[0m
[38;5;13m•! [PANICKED] [131.638 seconds][0m
lifecycle
[38;5;243m/usr/tnf/tnf-src/cnf-certification-test/lifecycle/suite.go:47[0m
  [38;5;13m[1m[It] lifecycle-pod-recreation[0m [38;5;204m[lifecycle-pod-recreation][0m
  [38;5;243m/usr/tnf/tnf-src/cnf-certification-test/lifecycle/suite.go:97[0m

  [38;5;243mBegin Captured GinkgoWriter Output >>[0m
    [1mSTEP:[0m Testing node draining effect of deployment [38;5;243m06/09/22 21:19:04.006[0m
    [1mSTEP:[0m Testing initial state for deployments [38;5;243m06/09/22 21:19:04.006[0m
    [1mSTEP:[0m Draining and Cordoning node master-2:  [38;5;243m06/09/22 21:19:04.022[0m
  [38;5;243m<< End Captured GinkgoWriter Output[0m

  [38;5;13mTest Panicked[0m
  [38;5;13mIn [1m[It][0m[38;5;13m at: [1m/usr/local/go/src/runtime/panic.go:220[0m

  [38;5;13mruntime error: invalid memory address or nil pointer dereference[0m
```

Adds a simple check for the `deployment` object itself being nil and logs an error.  Previously, the `provider.DeploymentToString` was throwing a panic because `dp` was nil.


cnf-certification-test/lifecycle/podsets/podsets.go
```diff
@@ -40,6 +40,8 @@ var WaitForDeploymentSetReady = func(ns, name string, timeout time.Duration) boo
 		if err == nil && IsDeploymentReady(dp) {
 			logrus.Tracef("%s is ready", provider.DeploymentToString(dp))
 			return true
+		} else if dp == nil {
+			logrus.Error("error: deployment object found to be nil")
 		} else if err != nil {
 			logrus.Errorf("Error while getting the %s, err: %s", provider.DeploymentToString(dp), err)
```
greyerof, line 46: Suggestion: dp will be nil only when err != nil, so instead of adding a new "else if" branch, you can just use "ns" & "name" here from this function's params:
```
  } else if err != nil {
    logrus.Errorf("Error while getting deployment %s (ns: %s), err: %s", name, ns, err)
  }
```

jmontesi: I think the check for a `dp != nil`/`err != nil` should happen right after the call to `provider.GetUpdatedDeployment()` before  `dp` is used in `IsDeploymentReady()`

sebrandon1: Fixed.  Added string recommended by @greyerof

cnf-certification-test/lifecycle/podsets/podsets.go
```diff
@@ -37,7 +37,9 @@ var WaitForDeploymentSetReady = func(ns, name string, timeout time.Duration) boo
 	start := time.Now()
 	for time.Since(start) < timeout {
 		dp, err := provider.GetUpdatedDeployment(clients.K8sClient.AppsV1(), ns, name)
-		if err == nil && IsDeploymentReady(dp) {
+		if dp == nil {
```
greyerof, line 40: Sorry, Brandon, but I still don't see the need for this check `dp == nil`. Usually, after a function call in go, the first thing to check is the error with a `if err != nil`, and that's precisely the only moment where dp will be nil, so there's no need for another check for that. I suggest this way of checking the output:
```
	for time.Since(start) < timeout {
		dp, err := provider.GetUpdatedDeployment(clients.K8sClient.AppsV1(), ns, name)
		if err != nil {
			logrus.Errorf("Error while getting deployment %s (ns: %s), err: %s", name, ns, err)
		} else if !IsDeploymentReady(dp) {
			logrus.Errorf("%s is not ready yet", provider.DeploymentToString(dp))
		} else {
			logrus.Tracef("%s is ready!", provider.DeploymentToString(dp))
			return true
		}
	    
		time.Sleep(time.Second)
	}
```

sebrandon1: Okay, I agree this works.  I will update.

jmontesi (approved):

greyerof (approved):

PR #272, opened 2022-06-13
Copy chaostesting files in via Dockerfile

sebrandon1:
Okay I set the folder where to copy the files as `/usr/tnf/chaostesting/chaos-test-files/`.

I believe the binary lives in `/usr/tnf` as well:
```
ENV TNF_DIR=/usr/tnf
ENV TNF_SRC_DIR=${TNF_DIR}/tnf-src
ENV TNF_BIN_DIR=${TNF_DIR}/cnf-certification-test
```

This should be the correct path but I'll need some additional reviews.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -76,7 +76,10 @@ RUN mkdir ${TNF_BIN_DIR} && \
     cp --parents `find -name \*.db*` ${TNF_DIR} && \
 	# copy all JSON files to allow tests to run
 	cp --parents `find -name \*.json*` ${TNF_DIR} && \
-	cp cnf-certification-test/cnf-certification-test.test ${TNF_BIN_DIR}
+	cp cnf-certification-test/cnf-certification-test.test ${TNF_BIN_DIR} && \
+    # copy all of the chaos-test-files
+    mkdir -p ${TNF_DIR}/chaostesting/chaos-test-files && \
+    cp /cnf-certification-test/chaostesting/chaos-test-files/ ${TNF_DIR}/chaostesting/chaos-test-files/
 
 WORKDIR ${TNF_DIR}
 
@@ -96,7 +99,6 @@ RUN yum remove -y gcc git wget && \
 FROM scratch
 ARG TNF_PARTNER_DIR=/usr/tnf-partner
 COPY --from=build / /
-COPY /cnf-certification-test/chaostesting/chaos-test-files/ /usr/tnf/chaostesting/chaos-test-files/
 ENV TNF_CONFIGURATION_PATH=/usr/tnf/config/tnf_config.yml
 ENV KUBECONFIG=/usr/tnf/kubeconfig/config
 ENV TNF_PARTNER_SRC_DIR=$TNF_PARTNER_DIR/src
--- a/Dockerfile
+++ b/Dockerfile
@@ -78,8 +78,8 @@ RUN mkdir ${TNF_BIN_DIR} && \
 	cp --parents `find -name \*.json*` ${TNF_DIR} && \
 	cp cnf-certification-test/cnf-certification-test.test ${TNF_BIN_DIR} && \
     # copy all of the chaos-test-files
-    mkdir -p ${TNF_DIR}/chaostesting/chaos-test-files && \
-    cp /cnf-certification-test/chaostesting/chaos-test-files/ ${TNF_DIR}/chaostesting/chaos-test-files/
+    mkdir -p ${TNF_DIR}/cnf-certification-test/chaostesting && \
+    cp -a cnf-certification-test/chaostesting/chaos-test-files ${TNF_DIR}/cnf-certification-test/chaostesting
 
 WORKDIR ${TNF_DIR}
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -96,6 +96,7 @@ RUN yum remove -y gcc git wget && \
 FROM scratch
 ARG TNF_PARTNER_DIR=/usr/tnf-partner
 COPY --from=build / /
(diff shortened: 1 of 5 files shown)


sebrandon1:
The files are hardcoded in the code to be located at:
```
serviceAccountFile = "chaostesting/chaos-test-files/service-account.yaml"
experimentFile     = "chaostesting/chaos-test-files/experiment-delete.yaml"
chaosEngineFile    = "chaostesting/chaos-test-files/chaos-engine.yaml"
```

aabughosh (approved):

jmontesi:
I do not see where the Litmus operator itself is installed in the partner cluster, so this fix might not work.

aabughosh:
> 

we have this in the parnter repo 
install-litmus:
	./scripts/install-litmus-operator.sh

aabughosh:
> @aabughosh I believe that `./scripts/install-litmus-operator.sh` needs to be ran as part of the build process here as well then if that's required to run these YAMLs.

its required  its installing the litmus

Dockerfile
```diff
@@ -76,7 +76,10 @@ RUN mkdir ${TNF_BIN_DIR} && \
     cp --parents `find -name \*.db*` ${TNF_DIR} && \
 	# copy all JSON files to allow tests to run
 	cp --parents `find -name \*.json*` ${TNF_DIR} && \
-	cp cnf-certification-test/cnf-certification-test.test ${TNF_BIN_DIR}
+	cp cnf-certification-test/cnf-certification-test.test ${TNF_BIN_DIR} && \
+    # copy all of the chaos-test-files
+    mkdir -p ${TNF_DIR}/chaostesting/chaos-test-files && \
+    cp /cnf-certification-test/chaostesting/chaos-test-files/ ${TNF_DIR}/chaostesting/chaos-test-files/
```
greyerof, line 82: This copy probably won't work. Files need to be copied into the "${TNF_DIR}/cnf-certification-test". Also, a recursive copy is needed as it has subfolders. What about this:
```
    mkdir -p ${TNF_DIR}/cnf-certification-test/chaostesting && \
    cp -a cnf-certification-test/chaostesting/chaos-test-files ${TNF_DIR}/cnf-certification-test/chaostesting
```

pkg/provider/provider.go
```diff
@@ -289,6 +289,11 @@ func isSkipHelmChart(helmName string, skipHelmChartList []configuration.SkipHelm
 	return false
 }
 
+func ReloadTestEnvironment() TestEnvironment {
```
greyerof, line 292: I'm not sure whether this function is actually needed as we already have the `env.SetNeedsRefresh()`. We already use it in some of the lifecycle TS's test cases.

sebrandon1: Oh nice, I'll remove my func then and use the existing.

Dockerfile
sebrandon1: Fixed!

run-cnf-suites.sh
```diff
@@ -104,6 +104,7 @@ if [[ -z "${TNF_PARTNER_SRC_DIR}" ]]; then
 	echo "env var \"TNF_PARTNER_SRC_DIR\" not set, running the script without updating infra"
 else
 	make -C $TNF_PARTNER_SRC_DIR install-partner-pods
+	make -C $TNF_PARTNER_SRC_DIR install-litmus
```
greyerof, line 107: Shouldn't we remove the litmus operator after TNF has finished, after line 128?
```
if [[ ! -z "${TNF_PARTNER_SRC_DIR}" ]]; then
	make -C $TNF_PARTNER_SRC_DIR delete-litmus
fi
```

sebrandon1: Good suggestion, I'll add that.

sebrandon1: Added the removal after the call to the test.

run-cnf-suites.sh
```diff
@@ -126,3 +127,8 @@ else
 fi
 
 cd ./cnf-certification-test && ./cnf-certification-test.test $FOCUS_STRING $SKIP_STRING $LABEL_STRING ${GINKGO_ARGS}
+
+if [ -d ${TNF_PARTNER_DIR}/src ] &&  [[ ! -z "${TNF_PARTNER_SRC_DIR}" ]]; then
```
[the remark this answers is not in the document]
greyerof: I don't think we need this extra check. I believe we should remove these two env vars [from the github workflow](https://github.com/test-network-function/cnf-certification-test/blob/cad0513bbe1103b5fac5a2301d928d0f34c90121/.github/workflows/pre-main.yaml#L23). Both env vars (`TNF_PARTNER_DIR` and `TNF_PARNER_SRC_DIR`) are used [by the tnf Dockerfile](https://github.com/test-network-function/cnf-certification-test/blob/cad0513bbe1103b5fac5a2301d928d0f34c90121/Dockerfile#L2) to download the partner repo into a specific container folder. Thus, `TNF_PARNER_SRC_DIR` is checked because when `run-cnf-suites.sh` is running from the container, we want to call the script to create the daemonset.
So, if I'm not wrong, [your original commit](https://github.com/test-network-function/cnf-certification-test/commit/9b0fbd5c83024defb46898de3195f323b708ef37) will work well if those two env vars are removed from the workflow parameters.
Other than that, I don't know why there's two env vars... just `TNF_PARTNER_DIR` would do it, but that could be addressed in a different PR.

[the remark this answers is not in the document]
sebrandon1: Okay I have removed the ENV vars.

greyerof (approved):

PR #275, opened 2022-06-14
Use CSV name to search in online catalog.

greyerof:
The URL to get the operator info was using the "package" field to filter, but there's also a "csv_name" on each entry that can be used as a filter, as that's exactly the field we're using in our TC. Also, the offline checker uses the csv name by default (see `buildOperatorKey()` so now both checkers online/offline are aligned.

### DIFF
--- a/internal/api/api.go
+++ b/internal/api/api.go
@@ -10,7 +10,7 @@ import (
 //go:generate moq -out api_moq.go . CertAPIClientFuncs
 type CertificationValidator interface {
 	IsContainerCertified(registry, repository, tag, digest string) bool
-	IsOperatorCertified(operatorName, ocpVersion, channel string) bool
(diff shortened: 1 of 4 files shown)


sebrandon1 (approved):

hamadise:
@greyerof is the offline check fine ? 
the function
```
func (checker OfflineChecker) IsOperatorCertified(operatorName, ocpVersion, channel string) bool {
```
should have the same variable names

greyerof:
@hamadise I've uploaded [a commit](https://github.com/test-network-function/cnf-certification-test/pull/275/commits/d6714ac7a97bbb4b672c93de17ab37b90f987b12) to update that argument name of the offline checker function so it's more accurate now.

hamadise (approved):

PR #291, opened 2022-06-21
updated service mesh

shimritproj:
Verifies if service mesh is exist by using in Istio.

### DIFF
--- a/pkg/autodiscover/autodiscover_operators.go
+++ b/pkg/autodiscover/autodiscover_operators.go
@@ -34,7 +34,7 @@ const (
 	istioNamespace = "istio-system"
 )
 
-func findnamespace(oc corev1client.CoreV1Interface) bool {
+func findIstioNamespace(oc corev1client.CoreV1Interface) bool {
 	nsList, err := oc.Namespaces().List(context.TODO(), metav1.ListOptions{})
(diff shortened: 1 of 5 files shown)


pkg/autodiscover/autodiscover_operators.go
```diff
@@ -31,16 +31,17 @@ import (
 )
 
 const (
-	istio = "istio-system"
+	istioNamespace = "istio-system"
 )
 
 func findnamespace(oc corev1client.CoreV1Interface) bool {
```
greyerof, line 37: I also added a comment about [the name of this function](https://github.com/test-network-function/cnf-certification-test/pull/228#discussion_r898209585). Can you review it, please?

greyerof (commented):
There's only one comment left. By the way, you should be more specific with the PR subjects. "updated service mesh" is too generic.  Doesn't "Service mesh test case updated." look a bit better? Also note it's a good idea to start it with a capital letter.

pkg/autodiscover/autodiscover_operators.go
shimritproj: sure, tnx for your attention.

greyerof (approved):

PR #294, opened 2022-06-21
Refactor Access Control

rdavid:
- Unite common flows into cycles.
- The logic is not changed.

### DIFF
--- a/cnf-certification-test/accesscontrol/suite.go
+++ b/cnf-certification-test/accesscontrol/suite.go
@@ -32,111 +32,119 @@ import (
 	"github.com/test-network-function/cnf-certification-test/pkg/provider"
 	"github.com/test-network-function/cnf-certification-test/pkg/testhelper"
 	"github.com/test-network-function/cnf-certification-test/pkg/tnf"
+	"github.com/test-network-function/test-network-function-claim/pkg/claim"
 )
 
+type Test struct {
+	id  claim.Identifier
+	run func(*provider.TestEnvironment)
+}
+
 var (
-	nonCompliantCapabilities = []string{"NET_ADMIN", "SYS_ADMIN", "NET_RAW", "IPC_LOCK"}
+	nonCompliantCapabilities = []string{
+		"NET_ADMIN",
+		"SYS_ADMIN",
+		"NET_RAW",
+		"IPC_LOCK",
+	}
 	invalidNamespacePrefixes = []string{
 		"default",
 		"openshift-",
 		"istio-",
 		"aspenmesh-",
 	}
+	containerTests = []Test{
+		{
+			id:  identifiers.TestSecConCapabilitiesIdentifier,
+			run: TestSecConCapabilities,
+		},
+		{
+			id:  identifiers.TestSecConNonRootUserIdentifier,
+			run: TestSecConRootUser,
+		},
+		{
+			id:  identifiers.TestSecConPrivilegeEscalation,
+			run: TestSecConPrivilegeEscalation,
+		},
+		{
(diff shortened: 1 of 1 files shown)


cnf-certification-test/accesscontrol/suite.go
```diff
@@ -32,111 +32,119 @@ import (
...
+		"NET_RAW",
+		"IPC_LOCK",
+	}
 	invalidNamespacePrefixes = []string{
 		"default",
 		"openshift-",
 		"istio-",
 		"aspenmesh-",
 	}
+	testContainer = []Test{
```
greyerof, line 56: I would change these new slice names to `containerTests`, `podTests` and `namespaceTests`.

cnf-certification-test/accesscontrol/suite.go
```diff
@@ -32,111 +32,119 @@ import (
...
+		{
+			id: identifiers.TestSecConCapabilitiesIdentifier,
+			do: TestSecConCapabilities,
```
greyerof, line 59: Suggestion: `run` instead of `do`.

cnf-certification-test/accesscontrol/suite.go
```diff
@@ -32,111 +32,119 @@ import (
...
-		testhelper.SkipIfEmptyAny(ginkgo.Skip, env.Pods)
-		TestAutomountServiceToken(&env)
-	})
-	// one process per container
-	testID = identifiers.XformToGinkgoItIdentifier(identifiers.TestOneProcessPerContainerIdentifier)
-	ginkgo.It(testID, ginkgo.Label(testID), func() {
-		testhelper.SkipIfEmptyAny(ginkgo.Skip, env.Containers)
-		TestOneProcessPerContainer(&env)
-	})
+	for _, t := range testContainer {
```
greyerof, line 127: Another suggestion: t --> test.
```
for _, test := range containerTests {
   ...
   ginkgo.It(..., func() {
     ...
     test.Run(&env)
  })
}
```
Doesn't it look a bit better?

greyerof (approved):
Just minor comments/suggestions.

sebrandon1 (approved):
Approved pending @greyerof's comments.

cnf-certification-test/accesscontrol/suite.go
```diff
@@ -32,111 +32,119 @@ import (
...
+		tid := identifiers.XformToGinkgoItIdentifier(t.id)
+		ginkgo.It(tid, ginkgo.Label(tid), func() {
+			testhelper.SkipIfEmptyAny(ginkgo.Skip, env.Containers)
```
edcdavid, line 130: To be honest, I prefer the current pattern. We are loosing the ability to have a custom skip function per test case ( the testhelper.SkipIfEmptyAny functions). It works in this case but not all suites use the same. Also, modifying the common code will need regression test for all tests. I feel this part should not be optimized.

sebrandon1: I agree that it's more uniform across all of the test suites to keep it the way it is.

rdavid: To have the ability I can add skip function to the Test structure. The current function is too long, the actions are out of order and looks messy.

I would add Containers/Pods/Namespaces to the structure as well and have only one cycle, but I did not find a way to do in Go, it will be a pointer on a data member in C++, which is to complex for Go :-)

rdavid: If we are agree on this I will fix the other suites as well.

PR #305, opened 2022-06-28
Use set for target folders in platform tests

rdavid:
- Use mapset from https://github.com/deckarep/golang-set.
- The logic is not changed.

### DIFF
--- a/cnf-certification-test/platform/cnffsdiff/fsdiff.go
+++ b/cnf-certification-test/platform/cnffsdiff/fsdiff.go
@@ -85,7 +85,6 @@ func intersectTargetFolders(src []string) []string {
 	return dst
 }
 
-//nolint:funlen
 func (f *FsDiff) RunTest(ctx clientsholder.Context, containerUID string) {
 	output, outerr, err := f.ClientHolder.ExecCommandContainer(ctx, fmt.Sprintf("chroot /host podman diff --format json %s", containerUID))
 	if err != nil {


cnf-certification-test/platform/cnffsdiff/fsdiff.go
```diff
@@ -73,6 +74,17 @@ func NewFsDiffTester(client clientsholder.Command) *FsDiff {
 	}
 }
 
+func intersectTargetFolders(src []string) []string {
+	var dst []string
+	for _, folder := range src {
+		if targetFolders.Contains(folder) {
```
sebrandon1, line 80: Instead of this, I believe we could just use the existing `StringInSlice` func found [here](https://github.com/test-network-function/cnf-certification-test/blob/main/pkg/tnf/utils/utils.go#L25).  That way we don't need the MapSet library at all.

cnf-certification-test/platform/cnffsdiff/fsdiff.go
```diff
@@ -20,26 +20,27 @@ import (
...
+		"/lib",
+		"/lib64",
+		"/sbin",
+		"/usr/bin",
+		"/usr/lib",
+		"/usr/lib64",
+		"/usr/sbin",
+		"/var/lib/rpm",
+		"/var/lib/dpkg",
+	)
```
sebrandon1, line 43:
```suggestion
	targetFolders = []string{
		"/bin",
		"/lib",
		"/lib64",
		"/sbin",
		"/usr/bin",
		"/usr/lib",
		"/usr/lib64",
		"/usr/sbin",
		"/var/lib/rpm",
		"/var/lib/dpkg",
	}
```

Just switch this to a `[]string` and then we can use `StringInSlice`.

go.mod
```diff
@@ -49,6 +49,7 @@ require (
 	github.com/containerd/containerd v1.6.6 // indirect
 	github.com/cyphar/filepath-securejoin v0.2.3 // indirect
 	github.com/davecgh/go-spew v1.1.1 // indirect
+	github.com/deckarep/golang-set/v2 v2.1.0 // indirect
```
sebrandon1, line 52: I'm curious why this says //indirect as we are directly referencing in the code?

cnf-certification-test/platform/cnffsdiff/fsdiff.go
sebrandon1: Marking as resolved.

cnf-certification-test/platform/cnffsdiff/fsdiff.go
sebrandon1: Decided against switching this to use StringInSlice to take advantage of the O(1) lookup instead of the O(n).

sebrandon1: Also yes, we know that the old Go maps were O(1) as well.  😄

go.mod
rdavid: I guess it is because we say v2, not v2.1.0. Actually I prefer to set the exact version, I didn't find a way to do it.

sebrandon1 (approved):
We discussed the benefits of using this new Set library.  I think it could be a beneficial improvement to places in the code.

PR #309, opened 2022-06-29
use Goclient to dynamically deploy the partner repo daemonset

shimritproj:
use Goclient to dynamically deploy the partner repo daemonset.


cnf-certification-test/daemonset/daemonset.go
```diff
@@ -0,0 +1,231 @@
...
+	var ptpPods *v1core.PodList
+	ptpPods, err = client.CoreV1().Pods(namespace).List(context.TODO(), metav1.ListOptions{LabelSelector: "name=" + daemonSetName})
+	if err != nil {
+		return ptpPods, err
+	}
+	fmt.Printf("Successfully created daemon set %s\n", daemonSetName)
+	return ptpPods, nil
+}
+
+func partnerRepoDaemonset(daemonSetName, namespace, containerName, imageWithVersion string) {
```
sebrandon1, line 221: This function doesn't have a return value or handle potential errors.

cnf-certification-test/daemonset/daemonset.go
```diff
@@ -0,0 +1,231 @@
...
+	fmt.Printf("Successfully cleaned up daemon set %s\n", daemonSetName)
+	return nil
+}
+
+// Check if the daemon set exists
+func doesDaemonSetExist(daemonSetName, namespace string) bool {
+	client := clientsholder.GetClientsHolder().K8sClient.AppsV1()
+	_, err := client.DaemonSets(namespace).Get(context.TODO(), daemonSetName, metav1.GetOptions{})
+	if err != nil {
+		fmt.Println("Error ocurred for" + err.Error())
```
sebrandon1, line 188:
```suggestion
		fmt.Println("Error occurred checking for Daemonset to exist: " + err.Error())
```

cnf-certification-test/daemonset/daemonset.go
```diff
@@ -0,0 +1,231 @@
...
+	"context"
+	"errors"
+	"fmt"
+	"time"
+
+	"github.com/sirupsen/logrus"
+	"github.com/test-network-function/cnf-certification-test/internal/clientsholder"
+	appsv1 "k8s.io/api/apps/v1"
+	v1 "k8s.io/api/apps/v1"
+	v1core "k8s.io/api/core/v1"
```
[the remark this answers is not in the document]
shimritproj: I agree.

bnshr (changes requested):

internal/daemonset/daemonset.go
```diff
@@ -0,0 +1,216 @@
...
+	ginkgo.AfterEach(func() {
+		env.SetNeedsRefresh()
+	})
+
+	testID := identifiers.XformToGinkgoItIdentifier(identifiers.TestICMPv4ConnectivityIdentifier)
+	ginkgo.It(testID, ginkgo.Label(testID), func() {
+		partnerRepoDaemonset()
+	})
+
+})
```
sebrandon1, line 46: I am not sure I understand this part.  This isn't a test, we are just trying to deploy the daemonset in the code versus the call to `install-partner-pods` [here](https://github.com/test-network-function/cnf-certification-test/blob/968feaa74a036b4893d6560114e8199817d371c0/run-cnf-suites.sh#L105).

I would think we would call to create the daemonset somewhere around [here](https://github.com/test-network-function/cnf-certification-test/blob/main/cnf-certification-test/suite_test.go#L162) in TestTest() that would deploy the daemonset first before the call to `ginkgo.RunSpecs(t, CnfCertificationTestSuiteName)`.

cnf-certification-test/suite_test.go
```diff
@@ -159,6 +160,7 @@ func TestTest(t *testing.T) {
 
 	// Run tests specs only if not in diagnostic mode, otherwise all TSs would run.
 	if !diagnosticMode {
+		damonset.PartnerRepoDaemonset()
```
sebrandon1, line 163:
```suggestion
		daemonset.PartnerRepoDaemonset()
```

internal/daemonset/daemonset.go
```diff
@@ -0,0 +1,194 @@
...
+	_, err := client.AppsV1().DaemonSets(namespace).Create(context.TODO(), rebootDaemonSet, metav1.CreateOptions{})
+	if err != nil {
+		return nil, err
+	}
+	err = provider.WaitDebugPodsReady()
+	if err != nil {
+		return nil, err
+	}
+
+	tnf.ClaimFilePrintf("DeamonSet is ready")
```
sebrandon1, line 172:
```suggestion
	tnf.ClaimFilePrintf("Daemonset is ready")
```

internal/daemonset/daemonset.go
```diff
@@ -0,0 +1,194 @@
...
+	daemonSetName    = "partner-repo"
+	namespace        = "default"
+	containerName    = "container-00"
+	imageWithVersion = "quay.io/testnetworkfunction/debug-partner:latest"
+	timeout          = 5 * time.Minute
+	nodeExporter     = "node-exporter"
+)
+
+//nolint:funlen
+func CreateDaemonSetsTemplate(dsName, namespace, containerName, imageWithVersion string) *v1.DaemonSet {
```
bnshr, line 27: You could configure the security context here by passing a bool parameter to this function.
`func CreateDaemonSetsTemplate(dsName, namespace, containerName, imageWithVersion string, runAsPrivileged bool)` and when runAsPrivileged is true, then use make `** Privileged: true**.

internal/daemonset/daemonset.go
```diff
@@ -0,0 +1,194 @@
...
+								},
+							},
+						},
+					},
+				},
+			},
+		},
+	}
+}
+func DeleteDaemonSet(daemonSetName, namespace string) error {
```
bnshr, line 114: Please add a comment before this function as it is to be exported.

bnshr (changes requested):

internal/daemonset/daemonset.go
shimritproj: you're right. I did it.

internal/daemonset/daemonset.go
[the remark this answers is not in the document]
shimritproj: you're right. I did it.

sebrandon1: I was looking at our existing debug daemonset YAML [here](https://github.com/test-network-function/cnf-certification-test-partner/blob/main/test-partner/debugpartner.yaml#L30) and the privileged pod setting is always true.

shimritproj: I changed the 'runAsPrivileged' to be 'True' always.

cnf-certification-test/daemonset/daemonset.go
shimritproj: I fixed it.

internal/daemonset/daemonset.go
```diff
@@ -0,0 +1,199 @@
...
+	"github.com/test-network-function/cnf-certification-test/pkg/provider"
+	"github.com/test-network-function/cnf-certification-test/pkg/tnf"
+	v1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	daemonSetName    = "partner-repo"
+	namespace        = "default"
```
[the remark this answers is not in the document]
shimritproj: I think that "default" is ok, Do you think that it is need to be "tnf"? @bnshr

bnshr (approved):

cnf-certification-test/suite_test.go
```diff
@@ -159,6 +160,7 @@ func TestTest(t *testing.T) {
 
 	// Run tests specs only if not in diagnostic mode, otherwise all TSs would run.
 	if !diagnosticMode {
+		daemonset.PartnerRepoDaemonset()
```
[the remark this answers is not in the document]
edcdavid: right, correct. Not sure what is best, in ptp-operator, we delete it at the end.

internal/daemonset/daemonset.go
```diff
@@ -0,0 +1,200 @@
...
+}
+
+// Deploy daemon set on repo partner
+func PartnerRepoDaemonset() map[string]corev1.Pod {
+	dsRunningPods, err := CreateDaemonSet(daemonSetName, namespace, containerName, imageWithVersion, timeout)
+	if err != nil {
+		logrus.Errorf("Error : +%v\n", err.Error())
+	}
+
+	nodeToPodMapping := make(map[string]corev1.Pod)
```
edcdavid, line 194: This is only valid if deploying the daemonset is successful, also handle the case of the daemonset deployment not successfull

run-cnf-suites.sh
```diff
@@ -103,10 +103,8 @@ if [[ ! -f "/proc/1/cgroup" ]] || grep -q init\.scope /proc/1/cgroup; then
 fi
 
 if [[ ! -z "${TNF_PARTNER_SRC_DIR}" ]]; then
-	echo "attempting to install partner pods"
-	make -C $TNF_PARTNER_SRC_DIR install-partner-pods
-	#echo "attempting to install litmus"
-	#make -C $TNF_PARTNER_SRC_DIR install-litmus
+	echo "attempting to install litmus"
+	make -C $TNF_PARTNER_SRC_DIR install-litmus
```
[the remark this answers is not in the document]
shimritproj: Agreed, I did it.

internal/daemonset/daemonset.go
```diff
@@ -0,0 +1,200 @@
...
+	if err != nil {
+		return ptpPods, err
+	}
+	tnf.ClaimFilePrintf("Successfully created daemon set %s\n", daemonSetName)
+	return ptpPods, nil
+}
+
+// Deploy daemon set on repo partner
+func PartnerRepoDaemonset() map[string]corev1.Pod {
+	dsRunningPods, err := CreateDaemonSet(daemonSetName, namespace, containerName, imageWithVersion, timeout)
```
edcdavid, line 189: Use provider.DaemonSetName here, otherwise it will not match the name used by the provider.WaitDebugPodsReady function (debug!=partner-repo)

internal/daemonset/daemonset.go
```diff
@@ -0,0 +1,200 @@
...
+	if err != nil {
+		tnf.ClaimFilePrintf("Error occurred checking for Daemonset to exist: " + err.Error())
+	}
+	// If the error is not found, that means the daemon set exists
+	return err == nil
+}
+
+// Create daemon set
+func CreateDaemonSet(daemonSetName, namespace, containerName, imageWithVersion string, timeout time.Duration) (*corev1.PodList, error) {
+	rebootDaemonSet := CreateDaemonSetsTemplate(daemonSetName, namespace, containerName, imageWithVersion)
```
edcdavid, line 157: rename rebootDaemonSet to aDaemonSet for instance

internal/daemonset/daemonset.go
```diff
@@ -0,0 +1,200 @@
...
+							},
+						},
+					},
+				},
+			},
+		},
+	}
+}
+
+// Delete dameno set
```
edcdavid, line 115: typo

cnf-certification-test/suite_test.go
jmontesi, line 163: I think I'd use another name for this function. Something like `DeployTestDaemonset()` or similar.

shimritproj: maybe PartnerDeployTestDamenoset()? @jmontesi

shimritproj: I agree.

internal/daemonset/daemonset.go
shimritproj: I fixed it.

cnf-certification-test/common/constant.go
```diff
@@ -22,6 +22,7 @@ const (
 	AccessControlTestKey      = "access-control"
 	LifecycleTestKey          = "lifecycle"
 	ChaosTesting              = "chaostesting"
+	Daemonset                 = "daemonset"
```
sebrandon1, line 25: Is this needed?

shimritproj: No needed, I was forget to remove it, now I removed.

internal/daemonset/daemonset.go
```diff
@@ -0,0 +1,199 @@
...
+			},
+			Template: corev1.PodTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: matchLabels,
+				},
+				Spec: corev1.PodSpec{
+					Containers:       []corev1.Container{container},
+					PreemptionPolicy: &preempt,
+					Priority:         &zeroInt32,
+					HostNetwork:      true,
```
greyerof, line 77: Please add this field here:
```
HostPID: true
```

internal/daemonset/daemonset.go
```diff
@@ -0,0 +1,199 @@
...
+		tnf.ClaimFilePrintf("The daemonset (%s) deletion is unsuccessful due to %s", daemonSetName, err)
+	}
+	doneCleanUp := false
+	for start := time.Now(); !doneCleanUp && time.Since(start) < timeout; {
+		client := clientsholder.GetClientsHolder().K8sClient
+		pods, err := client.CoreV1().Pods(namespace).List(context.TODO(), metav1.ListOptions{LabelSelector: "name=" + daemonSetName})
+		if err != nil {
+			return fmt.Errorf("failed to get pods, err: %s", err)
+		}
+		//nolint:ineffassign
```
greyerof, line 130: Normally, you should try to fix the linter errors that you get instead of hiding them with a "nolint" tag. Here, the linter complains because the assignment `doneCleanUp = true` (line 132) doesn't make any sense, as you're doing a `break` later. You don't need that extra doneCleanUp variable at all. Just a break:
```
for start := time.Now(); time.Since(start) < timeout; {
  ...
  ...
  if len(pods.Items) == 0 {
     break
  }
```

internal/daemonset/daemonset.go
```diff
@@ -0,0 +1,199 @@
...
+					},
+				},
+			},
+		},
+	}
+}
+
+// Delete daemon set
+func DeleteDaemonSet(daemonSetName, namespace string) error {
+	tnf.ClaimFilePrintf("Deleting daemon set %s\n", daemonSetName)
```
greyerof, line 115: Do not use tnf.ClaimFilePrintf() here, as this code is not running inside any ginkgo TC. Please use the proper logrus logging function `logrus.Infof/logrus.Warnf/logrus.Errorf`. The "\n" is not needed either, by the way.

internal/daemonset/daemonset.go
```diff
@@ -0,0 +1,199 @@
...
+	deletePolicy := metav1.DeletePropagationForeground
+	client := clientsholder.GetClientsHolder().K8sClient.AppsV1()
+	if err := client.DaemonSets(namespace).Delete(context.TODO(), daemonSetName, metav1.DeleteOptions{
```
greyerof, line 118: Instead of breaking this `if err ...` into multiple lines. Why not just a simple assignment and a following check?
```
	err := client.DaemonSets(namespace).Delete(context.TODO(), daemonSetName, metav1.DeleteOptions{
		PropagationPolicy: &deletePolicy,})
	if err != nil {
		tnf.ClaimFilePrintf("The daemonset (%s) deletion is unsuccessful due to %s", daemonSetName, err)
	}
```

internal/daemonset/daemonset.go
```diff
@@ -0,0 +1,199 @@
...
+		PropagationPolicy: &deletePolicy,
+	}); err != nil {
+		tnf.ClaimFilePrintf("The daemonset (%s) deletion is unsuccessful due to %s", daemonSetName, err)
```
greyerof, line 121: You should return with the error right here, as it doesn't make any sense to continue after the trace. You can just return a custom error like this:
```
return fmt.Errorf("daemonset %s deletion failed: %w", daemonSetName, err)
```

internal/daemonset/daemonset.go
```diff
@@ -0,0 +1,199 @@
...
+		pods, err := client.CoreV1().Pods(namespace).List(context.TODO(), metav1.ListOptions{LabelSelector: "name=" + daemonSetName})
+		if err != nil {
+			return fmt.Errorf("failed to get pods, err: %s", err)
+		}
+		//nolint:ineffassign
+		if len(pods.Items) == 0 {
+			doneCleanUp = true
+			break
+		}
+		time.Sleep(time.Duration(timeout.Minutes()))
```
greyerof, line 135: You shouldn't use the same `timeout` var here, as this is a "retryInterval" time for sleeping, not a timeout. You can call it `daemonsetDeletionCheckRetryInteval` and make its value to be 2-5 secs as much. 

Also, do not use the `.Minutes()` for this time.Sleep. A time.Duration is a nanoseconds value. So here you're passing 5 nanoseconds to the time.Sleep function here.

internal/daemonset/daemonset.go
```diff
@@ -0,0 +1,199 @@
...
+
+// Check if the daemon set exists
+func doesDaemonSetExist(daemonSetName, namespace string) bool {
+	client := clientsholder.GetClientsHolder().K8sClient.AppsV1()
+	_, err := client.DaemonSets(namespace).Get(context.TODO(), daemonSetName, metav1.GetOptions{})
+	if err != nil {
+		tnf.ClaimFilePrintf("Error occurred checking for Daemonset to exist: " + err.Error())
+	}
+	// If the error is not found, that means the daemon set exists
+	return err == nil
```
greyerof, line 150: Here, you already know that err is not nil, because you tested it in line 146, so this check should be redundant. This is my suggestion for this code:
```
if err != nil {
  logrus.Errof("Error occurred checking for Daemonset to exist: %v", err)
  return false
}
return true
```

internal/daemonset/daemonset.go
```diff
@@ -0,0 +1,199 @@
...
+}
+
+// Create daemon set
+func CreateDaemonSet(daemonSetName, namespace, containerName, imageWithVersion string, timeout time.Duration) (*corev1.PodList, error) {
+	aDaemonSet := CreateDaemonSetsTemplate(daemonSetName, namespace, containerName, imageWithVersion)
+	if doesDaemonSetExist(daemonSetName, namespace) {
+		err := DeleteDaemonSet(daemonSetName, namespace)
+		if err != nil {
+			tnf.ClaimFilePrintf("Failed to delete debug daemonset because: %s", err)
```
greyerof, line 159: You should return the error right after the log trace here, or create a new error using that trace as in my previous comment.

internal/daemonset/daemonset.go
```diff
@@ -0,0 +1,199 @@
...
+		return nil, err
+	}
+	err = provider.WaitDebugPodsReady()
+	if err != nil {
+		return nil, err
+	}
+
+	tnf.ClaimFilePrintf("Daemonset is ready")
+
+	var ptpPods *corev1.PodList
```
greyerof, line 176: ptpPods? This seems like a copy/paste typo. Please change to `daemonsetPods` or maybe just `pods`.

internal/daemonset/daemonset.go
```diff
@@ -0,0 +1,199 @@
...
+	ptpPods, err = client.CoreV1().Pods(namespace).List(context.TODO(), metav1.ListOptions{LabelSelector: "name=" + daemonSetName})
+	if err != nil {
+		return ptpPods, err
```
greyerof, line 179: Normally, when an error is detected, you want to return empty objects explicitly: 
```
return nil, err
```

internal/daemonset/daemonset.go
```diff
@@ -0,0 +1,199 @@
...
+	}
+	tnf.ClaimFilePrintf("Successfully created daemon set %s\n", daemonSetName)
+	return ptpPods, nil
+}
+
+// Deploy daemon set on repo partner
+func DeployPartnerTestDaemonset() map[string]corev1.Pod {
```
greyerof, line 186: Why do you need to return the list of daemonset pods here? You're not using them. I believe you should only return an `error` here, in case something went wrong inside the function.

internal/daemonset/daemonset.go
```diff
@@ -0,0 +1,199 @@
...
+	containerName    = "container-00"
+)
+
+//nolint:funlen
+func CreateDaemonSetsTemplate(dsName, namespace, containerName, imageWithVersion string) *v1.DaemonSet {
+	dsAnnotations := make(map[string]string)
+	dsAnnotations["debug.openshift.io/source-container"] = containerName
+	dsAnnotations["openshift.io/scc"] = nodeExporter
+	matchLabels := make(map[string]string)
+	matchLabels["name"] = dsName
```
greyerof, line 30: As @edcdavid pointed out, you need the tnf's debug label here:
```
matchLabels["test-network-function.com/app"] = "debug"

```

internal/daemonset/daemonset.go
shimritproj: I changed.

internal/daemonset/daemonset.go
shimritproj: you're right, tnx for your comment.

internal/daemonset/daemonset.go
shimritproj: I agree.

internal/daemonset/daemonset.go
shimritproj: right, I did it.

internal/daemonset/daemonset.go
shimritproj: I agree with you that now I don't use with this information but, Are we don't need to return the list of daemonset? or maybe print it? someone need to know this information. what do you think? @greyerof

internal/daemonset/daemonset.go
shimritproj: you'r right, my mistake!
tnx for your comment.

internal/daemonset/daemonset.go
shimritproj: I agree, tnx Gonzalo for your review!

greyerof (commented):
There are still a couple of comments pending from my previous review.

internal/daemonset/daemonset.go
```diff
@@ -0,0 +1,195 @@
...
+	deletePolicy := metav1.DeletePropagationForeground
+	client := clientsholder.GetClientsHolder().K8sClient.AppsV1()
+	err := client.DaemonSets(namespace).Delete(context.TODO(), daemonSetName, metav1.DeleteOptions{
+		PropagationPolicy: &deletePolicy})
+	if err != nil {
+		return fmt.Errorf("daemonset %s deletion failed: %w", daemonSetName, err)
+	}
+
+	for start := time.Now(); time.Since(start) < timeout; {
+		client := clientsholder.GetClientsHolder().K8sClient
```
greyerof, line 127: Please move this client retrieval outside this `for` loop (line 125). There's no need for it to be to be retrieved on each iteration.

internal/daemonset/daemonset.go
```diff
@@ -0,0 +1,195 @@
...
+		logrus.Infof("Error occurred checking for Daemonset to exist: %s", err.Error())
+		return false
+	}
+	// If the error is not found, that means the daemon set exists
+	return err == nil
+}
+
+// Create daemon set
+func CreateDaemonSet(daemonSetName, namespace, containerName, imageWithVersion string, timeout time.Duration) (*corev1.PodList, error) {
+	aDaemonSet := CreateDaemonSetsTemplate(daemonSetName, namespace, containerName, imageWithVersion)
```
greyerof, line 157: Just a minor suggestion: move this template creation to line 165 (right after the trace), closer to where it's going to be actually used (líne 166).

internal/daemonset/daemonset.go
```diff
@@ -0,0 +1,195 @@
...
+	client := clientsholder.GetClientsHolder().K8sClient.AppsV1()
+	err := client.DaemonSets(namespace).Delete(context.TODO(), daemonSetName, metav1.DeleteOptions{
+		PropagationPolicy: &deletePolicy})
+	if err != nil {
+		return fmt.Errorf("daemonset %s deletion failed: %w", daemonSetName, err)
+	}
+
+	for start := time.Now(); time.Since(start) < timeout; {
+		client := clientsholder.GetClientsHolder().K8sClient
+		pods, err := client.CoreV1().Pods(namespace).List(context.TODO(), metav1.ListOptions{LabelSelector: "name=" + daemonSetName})
```
greyerof, line 128: The label selector won't work for the current partner repo's daemonset. as it doesn't have the "name" label. Try using the debug label:
`LabelSelector: "test-network-function.com/app=debug"`
This should make the github's CI to pass.
You should also change this in line 179.

internal/daemonset/daemonset.go
shimritproj: I fixed it.

sebrandon1 (approved):

internal/daemonset/daemonset.go
```diff
@@ -0,0 +1,193 @@
...
+	for start := time.Now(); time.Since(start) < timeout; {
+		pods, err := client.CoreV1().Pods(namespace).List(context.TODO(), metav1.ListOptions{LabelSelector: "test-network-function.com/app=debug"})
+		if err != nil {
+			return fmt.Errorf("failed to get pods, err: %s", err)
+		}
+
+		if len(pods.Items) == 0 {
+			break
+		}
+		time.Sleep(time.Duration(timeout.Minutes()))
```
greyerof, line 43: There's a [previous comment](https://github.com/test-network-function/cnf-certification-test/pull/309#discussion_r927568164) from me about this Sleep that you haven't replied/addressed. I'll copy it here:

You shouldn't use the same `timeout` var here, as this is a more a "retry interval" time for sleeping before doing the next iteration, not a timeout. You can call it daemonsetDeletionCheckRetryInteval and make its value to be 2-5 secs as much.

Also, do not use the `.Minutes()` for this time.Sleep. A time.Duration is a nanoseconds value. So here you're passing 5 nanoseconds to the time.Sleep function here, which is pointless.

internal/daemonset/daemonset.go
```diff
@@ -0,0 +1,193 @@
...
+func DeleteDaemonSet(daemonSetName, namespace string) error {
+	logrus.Infof("Deleting daemon set %s", daemonSetName)
+	deletePolicy := metav1.DeletePropagationForeground
+	client := clientsholder.GetClientsHolder().K8sClient
+	err := client.AppsV1().DaemonSets(namespace).Delete(context.TODO(), daemonSetName, metav1.DeleteOptions{
+		PropagationPolicy: &deletePolicy})
+	if err != nil {
+		return fmt.Errorf("daemonset %s deletion failed: %w", daemonSetName, err)
+	}
+	for start := time.Now(); time.Since(start) < timeout; {
```
greyerof, line 34: You should also add some code to handle the "timeout" scenario. Imagine that one of the debug pods do not stop and len(pods.Items) != 0 on each iteration until the timeout.

internal/daemonset/daemonset.go
```diff
@@ -0,0 +1,193 @@
...
+	logrus.Infof("Successfully cleaned up daemon set %s", daemonSetName)
+	return nil
+}
+
+// Check if the daemon set exists
+func doesDaemonSetExist(daemonSetName, namespace string) bool {
+	client := clientsholder.GetClientsHolder().K8sClient.AppsV1()
+	_, err := client.DaemonSets(namespace).Get(context.TODO(), daemonSetName, metav1.GetOptions{})
+	if err != nil {
+		logrus.Infof("Error occurred checking for Daemonset to exist: %s", err.Error())
```
greyerof, line 55: There's no need to use err.Error() in this trace. You can just use "err" here.

internal/daemonset/daemonset.go
```diff
@@ -0,0 +1,193 @@
...
+		return false
+	}
+	// If the error is not found, that means the daemon set exists
+	return err == nil
```
greyerof, line 59: There's no need to check `err`. You can just return `nil` here.

cnf-certification-test/suite_test.go
```diff
@@ -158,7 +160,14 @@ func TestTest(t *testing.T) {
 	claimhelper.UnmarshalConfigurations(configurations, claimData.Configurations)
 
 	// Run tests specs only if not in diagnostic mode, otherwise all TSs would run.
+	var env provider.TestEnvironment
 	if !diagnosticMode {
+		err = daemonset.DeployPartnerTestDaemonset()
+		if err != nil {
+			log.Errorf("Error : +%v\n", err.Error())
```
greyerof, line 167: There's a [previous comment](https://github.com/test-network-function/cnf-certification-test/commit/d66784a8d8440e9353158b9fe1f62af9de84862d#r79443611) from me about this trace:

1. The newline "\n" char is not needed in these logging functions.
2. You can just use "err". There's no need to use err.Error() here.
3. Why the "+" symbol before the %v?
4. Please use a more meaningful error message like "Error deploying partner daemonset" or similar.

internal/daemonset/daemonset.go
```diff
@@ -0,0 +1,193 @@
...
+	}
+	logrus.Infof("Successfully created daemon set %s", daemonSetName)
+	return daemonsetPods, nil
+}
+
+// Deploy daemon set on repo partner
+func DeployPartnerTestDaemonset() error {
+	_, err := CreateDaemonSet(provider.DaemonSetName, provider.DaemonSetNamespace, containerName, imageWithVersion, timeout)
+	if err != nil {
+		logrus.Errorf("Error deploying partner daemonset %s", err.Error())
```
greyerof, line 189: There's no need to use err.Error() in this trace. You can just use "err" here.

internal/daemonset/daemonset.go
shimritproj: right, but I think that I need just return True and no nil.

cnf-certification-test/suite_test.go
shimritproj: It old version , I changed. 
look- 

func DeployPartnerTestDaemonset() error {
	_, err := CreateDaemonSet(provider.DaemonSetName, provider.DaemonSetNamespace, containerName, imageWithVersion, timeout)
	if err != nil {
		logrus.Errorf("Error deploying partner daemonset %s", err.Error())
		return err
	}
	return nil
}

internal/daemonset/daemonset.go
shimritproj: I fixed it.

internal/daemonset/daemonset.go
shimritproj: In this case I will do :

time.Sleep(time.Duration(daemonsetDeletionCheckRetryInteval))

when , daemonsetDeletionCheckRetryInteval = 5 sec

@greyerof

internal/daemonset/daemonset.go
```diff
@@ -0,0 +1,199 @@
...
+		pods, err := client.CoreV1().Pods(namespace).List(context.TODO(), metav1.ListOptions{LabelSelector: "test-network-function.com/app=debug"})
+		if err != nil {
+			return fmt.Errorf("failed to get pods, err: %s", err)
+		}
+
+		if len(pods.Items) == 0 {
+			allPodsRemoved = true
+			break
+		}
+		allPodsRemoved = false
```
greyerof, line 46: You don't need this assignment to false on each iteration, as it's already initialized in line 35 and the only place where it could change is in line 43, right before leaving the loop.

greyerof (commented):
Almost ready. Just a minor comment.

greyerof (approved):

PR #311, opened 2022-06-29
Print more details about containers that fail the fsdiff test

sebrandon1:
`cut.String()` prints out all of the details about the container instead of just the pod name.

### DIFF
--- a/cnf-certification-test/platform/suite.go
+++ b/cnf-certification-test/platform/suite.go
@@ -166,7 +166,7 @@ func testContainersFsDiff(env *provider.TestEnvironment) {
 	var badContainers []string
 	var errContainers []string
 	for _, cut := range env.Containers {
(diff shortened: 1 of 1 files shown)


cnf-certification-test/platform/suite.go
```diff
@@ -166,7 +166,7 @@ func testContainersFsDiff(env *provider.TestEnvironment) {
 	var badContainers []string
 	var errContainers []string
 	for _, cut := range env.Containers {
-		logrus.Debug(fmt.Sprintf("%s(%s) should not install new packages after starting", cut.Podname, cut.Data.Name))
+		logrus.Debug(fmt.Sprintf("%s should not install new packages after starting", cut.String()))
```
rdavid, line 169: I think it is more clear to have pod name and container name.

sebrandon1: This prints all of that via String(): https://github.com/test-network-function/cnf-certification-test/blob/main/pkg/provider/provider.go#L418

rdavid (approved):

PR #333, opened 2022-07-11
Chaos testing pod-delete test case disabled and removed from catalog.

greyerof:
The TC code is still there so it can also be manually run (after
removing the skip), for internal tests, but the catalog entry has been
removed so we can decide whether to keep chaos testing as part of TNF or
just agree on the final list of chaos test cases to be implemented
here.


sebrandon1 (approved):
Approved!

PR #336, opened 2022-07-12
run-cnf-suites.sh: avoid error when listing the specs

jmontesi:


run-cnf-suites.sh
```diff
@@ -59,6 +59,14 @@ while [[ $1 == -* ]]; do
 	shift
 done
 
+# Check if the test cluster runs OCP
+res=`oc version | grep  Server`
+if [ -z "$res" ]
+then
+   echo "Minikube or similar detected"
```
sebrandon1, line 66: As part of this PR, do you want to change `Minikube` to `Kind` since we don't use minikube anymore.

jmontesi: I'll rewrite it in a more generic way.

sebrandon1 (approved):

greyerof (approved):

jmontesi:
Error example:
ERROR  [Jul 13 09:26:25.762][provider.go: 713] Failed to get machineConfig name for node kind-worker

PR #357, opened 2022-07-18
Add logging around compatibility

sebrandon1:
Added some error handling and some more logging to help figure out why we are panicking in QE.


pkg/compatibility/compatibility.go
```diff
@@ -195,8 +196,16 @@ func IsRHCOSCompatible(machineVersion, ocpVersion string) bool {
 	lifecycleInfo := GetLifeCycleDates()
 	if entry, ok := lifecycleInfo[ocpVersion]; ok {
 		// Collect the machine version and the entry version
-		mv, _ := gv.NewVersion(machineVersion)
-		ev, _ := gv.NewVersion(entry.MinRHCOSVersion)
+		mv, err := gv.NewVersion(machineVersion)
+		if err != nil {
+			logrus.Errorf("Error parsing machineVersion: %s", machineVersion)
```
edcdavid, line 201: why not also print the err message ?

edcdavid (approved):

pkg/compatibility/compatibility.go
sebrandon1: Good idea, will add that.

PR #365, opened 2022-07-19
Add toleration check test

sebrandon1:
A new test that will check the existence of the default k8s tolerations on pods and if they are modified or added to this will flag a failure.

This is referenced in section 10.6 of the [v1.3 requirements doc](https://connect.redhat.com/sites/default/files/2022-05/Cloud%20Native%20Network%20Function%20Requirements%201-3.pdf).

### DIFF
--- a/CATALOG.md
+++ b/CATALOG.md
@@ -175,6 +175,18 @@ Description|http://test-network-function.com/testcases/access-control/requests-a
 Result Type|informative
 Suggested Remediation|Add requests and limits to your container spec.  See: https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#requests-and-limits
 Best Practice Reference|https://TODO Section 4.6.11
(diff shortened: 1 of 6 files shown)


sebrandon1:
Marked WIP until I can figure out how to get the CI to pass.

sebrandon1:
One thing I'm wondering about this requirement is pods automatically get a certain set of tolerations:
```
 tolerations:
  - effect: NoExecute
    key: node.kubernetes.io/not-ready
    operator: Exists
    tolerationSeconds: 300
  - effect: NoExecute
    key: node.kubernetes.io/unreachable
    operator: Exists
    tolerationSeconds: 300
    ```

But only for a certain amount of seconds.  How do we want to enforce this check?  Should we only enforce/fail the test when the toleration exists without a `tolerationSeconds` value?

greyerof (approved):

PR #372, opened 2022-07-21
Add test for requests and limits for containers

sebrandon1:
Simple test to check for the existence of `requests` and `limits` on a container spec which corresponds to 4.6.11 in the new v1.4 requirements document.  Note, I don't know if we have a public link yet for the v1.4 document so I marked it as TODO.

### DIFF
--- a/cnf-certification-test/lifecycle/suite.go
+++ b/cnf-certification-test/lifecycle/suite.go
@@ -110,6 +110,12 @@ var _ = ginkgo.Describe(common.LifecycleTestKey, func() {
 		testPodsRecreation(&env)
 	})
 
+	testID = identifiers.XformToGinkgoItIdentifier(identifiers.TestPodRequestsAndLimitsIdentifier)
+	ginkgo.It(testID, ginkgo.Label(testID), func() {
+		testhelper.SkipIfEmptyAny(ginkgo.Skip, env.Pods)
+		testPodRequestsAndLimits(&env)
+	})
+
 	if env.IsIntrusive() {
 		testID = identifiers.XformToGinkgoItIdentifier(identifiers.TestDeploymentScalingIdentifier)
 		ginkgo.It(testID, ginkgo.Label(testID), func() {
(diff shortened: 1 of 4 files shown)


cnf-certification-test/lifecycle/suite.go
```diff
@@ -372,3 +378,23 @@ func testPodsRecreation(env *provider.TestEnvironment) { //nolint:funlen
...
 }
+
+func testPodRequestsAndLimits(env *provider.TestEnvironment) {
+	var containersMissingRequestsOrLimits []string
+	ginkgo.By("Testing container resource requests and limits")
+
+	// Loop through the containers, looking for containers that are missing requests or limits.
+	// These need to be defined in order to pass.
+	for _, cut := range env.Containers {
+		if len(cut.Data.Resources.Limits) == 0 || len(cut.Data.Resources.Requests) == 0 {
```
rdavid, line 389: What is cut?

I think it is more clear to have two different conditions with two different messages. By the way one container could fail both conditions.

rdavid (changes requested):

cnf-certification-test/lifecycle/suite.go
sebrandon1: This is true.  @mwlinca and I have been talking about this as well and I think there is more to add.  I'm going to make this work in progress as I fix.

rdavid (approved):

cnf-certification-test/lifecycle/suite.go
```diff
@@ -372,3 +378,33 @@ func testPodsRecreation(env *provider.TestEnvironment) { //nolint:funlen
...
+
+func testPodRequestsAndLimits(env *provider.TestEnvironment) {
+	var containersMissingRequestsOrLimits []string
+	ginkgo.By("Testing container resource requests and limits")
+
+	// Loop through the containers, looking for containers that are missing requests or limits.
+	// These need to be defined in order to pass.
+	for _, cut := range env.Containers {
+		badContainer := false
+		if len(cut.Data.Resources.Limits) == 0 {
```
jmontesi, line 390: The requirement states that "Pods must define requests and limits values for **CPU** and **memory**" So, shouldn't this condition also check for the existence of the strings "cpu" and "memory" in the map?

cnf-certification-test/lifecycle/suite.go
```diff
@@ -372,3 +378,33 @@ func testPodsRecreation(env *provider.TestEnvironment) { //nolint:funlen
...
+			tnf.ClaimFilePrintf("Container has been found missing resource limits: %s", cut.String())
+			badContainer = true
+		}
+
+		if len(cut.Data.Resources.Requests) == 0 {
```
jmontesi, line 395: Same as above.

cnf-certification-test/lifecycle/suite.go
```diff
@@ -372,3 +378,33 @@ func testPodsRecreation(env *provider.TestEnvironment) { //nolint:funlen
...
+			badContainer = true
+		}
+
+		if badContainer {
+			containersMissingRequestsOrLimits = append(containersMissingRequestsOrLimits, cut.String())
+		}
+	}
+
+	if n := len(containersMissingRequestsOrLimits); n > 0 {
+		errMsg := fmt.Sprintf("Containers found that are missing either resource requests or limits: %d. See logs for more detail.", n)
```
jmontesi, line 406: If the above applies, the string should reference "requests or limits for cpu/memory".

cnf-certification-test/lifecycle/suite.go
sebrandon1: Good call, I can get more granular instead of just checking the top level.  Will add.

sebrandon1: Okay I added checks to make sure that CPU and memory are set.

jmontesi (approved):

PR #377, opened 2022-07-25
Address more lint problems

sebrandon1:
`staticcheck` seems to hate that `debugPod == nil` doesn't return immediately after instead of calling ginkgo.Fail.

Semi related to #369.  Not sure why it didn't flag these originally. The PR [passed the CI](https://github.com/test-network-function/cnf-certification-test/actions/runs/2733634001).

Also, I had to bump the timeout again because of https://github.com/golangci/golangci-lint/issues/2997

### DIFF
--- a/cnf-certification-test/accesscontrol/suite.go
+++ b/cnf-certification-test/accesscontrol/suite.go
@@ -446,30 +446,30 @@ func TestOneProcessPerContainer(env *provider.TestEnvironment) {
 		debugPod := env.DebugPods[cut.NodeName]
 		if debugPod == nil {
 			ginkgo.Fail(fmt.Sprintf("Debug pod not found on Node: %s", cut.NodeName))
-		}
+		} else {
+			ocpContext := clientsholder.Context{
+				Namespace:     debugPod.Namespace,
+				Podname:       debugPod.Name,
+				Containername: debugPod.Spec.Containers[0].Name,
+			}
 
-		ocpContext := clientsholder.Context{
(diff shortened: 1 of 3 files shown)


cnf-certification-test/accesscontrol/suite.go
```diff
@@ -446,30 +446,30 @@ func TestOneProcessPerContainer(env *provider.TestEnvironment) {
 		debugPod := env.DebugPods[cut.NodeName]
 		if debugPod == nil {
 			ginkgo.Fail(fmt.Sprintf("Debug pod not found on Node: %s", cut.NodeName))
```
greyerof, line 448: So here's where gomega would have come handy with something like this:
```
Expect(debugPod).NotTo(BeNil(), fmt.Sprintf("Debug pod not found on Node: %s", cut.NodeName))
```

greyerof (approved):

dominikh:
> staticcheck seems to hate that debugPod == nil doesn't return immediately after instead of calling ginkgo.Fail.

Staticcheck should be able to detect that ginkgo.Fail unconditionally panics and thus not flag future dereferences of debugPod.

I tested the parent of this commit, both with the proper `staticcheck` command and with golangci-lint 1.47.1, and neither of them flagged the code. If this is intermittently failing for you then that's fishy and worth investigating.

sebrandon1:
Thanks @dominikh.  I'll try what you did again just to see if maybe it was a fluke.

PR #386, opened 2022-07-28
Add namespace resource quota test

sebrandon1:
Depends on #372 and https://github.com/test-network-function/cnf-certification-test-partner/pull/140

This is just a branch off of the #372 PR so I didn't have to wait for that to merge.

https://github.com/test-network-function/cnf-certification-test-partner/pull/140 needs to merge first so the CI starts passing because our test environment currently doesn't have a limit range set.

### DIFF
--- a/cnf-certification-test/accesscontrol/suite.go
+++ b/cnf-certification-test/accesscontrol/suite.go
@@ -148,6 +148,11 @@ var _ = ginkgo.Describe(common.AccessControlTestKey, func() {
 		testhelper.SkipIfEmptyAny(ginkgo.Skip, env.Pods)
 		TestSysPtraceCapability(&env)
 	})
+	testID = identifiers.XformToGinkgoItIdentifier(identifiers.TestNamespaceResourceQuotaIdentifier)
+	ginkgo.It(testID, ginkgo.Label(testID), func() {
+		testhelper.SkipIfEmptyAny(ginkgo.Skip, env.Pods)
+		TestNamespaceResourceQuota(&env)
+	})
 })
 
 // TestSecConCapabilities verifies that non compliant capabilities are not present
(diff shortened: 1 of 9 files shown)


pkg/autodiscover/autodiscover_limitrange.go
```diff
@@ -0,0 +1,35 @@
...
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+)
+
+func getLimitRanges(oc corev1client.CoreV1Interface) ([]corev1.LimitRange, error) {
+	limitRanges := []corev1.LimitRange{}
+	lrl, err := oc.LimitRanges("").List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		return nil, err
+	}
+	limitRanges = append(limitRanges, lrl.Items...)
```
greyerof, line 33: Why are you appending here? You can just return the lrl.Items directly here and get rid of the limitRanges temp var:
```
return lrl.Items
```

greyerof (approved):

jmontesi:
Have you considered using the `ResourceQuota` object instead of the `LimitRange`? That first object seems to fit better the requirement as it is more global.

sebrandon1:
I'll take a look at the ResourceQuota objects.

cnf-certification-test/accesscontrol/suite.go
```diff
@@ -531,3 +536,67 @@ func TestSysPtraceCapability(env *provider.TestEnvironment) {
...
+		ginkgo.Fail(errMsg)
+	}
+}
+
+//nolint:dupl
+func TestNamespaceResourceQuota(env *provider.TestEnvironment) {
+	var namespacesMissingQuotas []string
+	ginkgo.By("Testing namespace resource quotas")
+
+	for _, put := range env.Pods {
```
jmontesi, line 577: As _ResourceQuotas_ apply to namespace level, wouldn't it be more convenient to iterate through namespaces instead of pods?

pkg/autodiscover/autodiscover_limitrange.go
sebrandon1: Fixed.

cnf-certification-test/accesscontrol/suite.go
```diff
@@ -531,3 +536,67 @@ func TestSysPtraceCapability(env *provider.TestEnvironment) {
...
+		// Look through all of the pods and compare their namespace to any potential
+		// limit ranges that exist for a certain namespace.
```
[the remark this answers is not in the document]
sebrandon1: Fixed.  I removed LimitRange test, kept the ResourceQuota.

cnf-certification-test/accesscontrol/suite.go
```diff
@@ -531,3 +536,67 @@ func TestSysPtraceCapability(env *provider.TestEnvironment) {
...
+		foundPodNamespaceLimit := false
```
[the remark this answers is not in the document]
sebrandon1: Fixed.  I removed LimitRange test, kept the ResourceQuota.

cnf-certification-test/accesscontrol/suite.go
```diff
@@ -531,3 +536,67 @@ func TestSysPtraceCapability(env *provider.TestEnvironment) {
...
+		for index := range env.ResourceQuotas {
+			// We are just checking for the existence of the limit range as of right now.
+			// Read more about the limit range object here:
+			// https://kubernetes.io/docs/tasks/administer-cluster/manage-resources/memory-default-namespace/
```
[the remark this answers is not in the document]
sebrandon1: Fixed.  I removed LimitRange test, kept the ResourceQuota.

cnf-certification-test/accesscontrol/suite.go
sebrandon1: I think we only care about workload pods that are deployed in their namespaces.  It might be possible there are other namespaces in the cluster that are not holding pods that we are testing?

jmontesi: That's right. I don't think the namespaces are labeled, so there is no way to distinguish the ones for running workloads from those for other purposes.

jmontesi (approved):

PR #390, opened 2022-08-01
Add test for persistent volume reclaim policy

sebrandon1:
Looks through all of the persistent volumes in the cluster and if they are not using `ReclaimPolicy: Delete` they are flagged as a failure.

Refers to section 3.3.4 in the v1.4 requirements document.

Also, I removed the `dupl` and `noctx` linters because I could get them to pass even with `nolint` tags.  🤷

### DIFF
--- a/cnf-certification-test/identifiers/identifiers.go
+++ b/cnf-certification-test/identifiers/identifiers.go
@@ -349,6 +349,10 @@ var (
 		Url:     formTestURL(common.AccessControlTestKey, "pod-toleration-bypass"),
 		Version: versionOne,
 	}
+	TestPersistentVolumeReclaimPolicyIdentifier = claim.Identifier{
+		Url:     formTestURL(common.LifecycleTestKey, "persistent-volume-reclaim-policy"),
+		Version: versionOne,
+	}
 )
 
 func formDescription(identifier claim.Identifier, description string) string {
@@ -955,6 +959,14 @@ that there are no changes to the following directories:
 		Description:           formDescription(TestPodRequestsAndLimitsIdentifier, `Check that containers have resource requests and limits specified in their spec.`),
 		Remediation:           RequestsAndLimitsRemediation,
 		BestPracticeReference: bestPracticeDocV1dot4URL + " Section 4.6.11",
+		ExceptionProcess:      NoDocumentedProcess,
+	},
+	TestPersistentVolumeReclaimPolicyIdentifier: {
+		Identifier:            TestPersistentVolumeReclaimPolicyIdentifier,
+		Type:                  informativeResult,
+		Description:           formDescription(TestPersistentVolumeReclaimPolicyIdentifier, `Check that the persistent volumes the CNF pods are using have a reclaim policy of delete.`),
+		Remediation:           PersistentVolumeReclaimPolicyRemediation,
+		BestPracticeReference: bestPracticeDocV1dot4URL + " Section 3.3.4",
 		ExceptionProcess:      NoDocumentedProcess,
 	},
 	TestNamespaceResourceQuotaIdentifier: {
--- a/cnf-certification-test/lifecycle/suite.go
+++ b/cnf-certification-test/lifecycle/suite.go
@@ -144,6 +144,11 @@ var _ = ginkgo.Describe(common.LifecycleTestKey, func() {
 			testStatefulSetScaling(&env, timeout)
 		})
 	}
+
+	testID = identifiers.XformToGinkgoItIdentifier(identifiers.TestPersistentVolumeReclaimPolicyIdentifier)
+	ginkgo.It(testID, ginkgo.Label(testID), func() {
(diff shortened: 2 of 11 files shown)


cnf-certification-test/identifiers/identifiers.go
```diff
@@ -938,5 +942,14 @@ that there are no changes to the following directories:
 		Description:           formDescription(TestPodRequestsAndLimitsIdentifier, `Check that containers have resource requests and limits specified in their spec.`),
 		Remediation:           RequestsAndLimitsRemediation,
 		BestPracticeReference: bestPracticeDocV1dot4URL + " Section 4.6.11",
+		ExceptionProcess:      NoDocumentedProcess,
+	},
+	TestPersistentVolumeReclaimPolicyIdentifier: {
+		Identifier:            TestPersistentVolumeReclaimPolicyIdentifier,
+		Type:                  informativeResult,
+		Description:           formDescription(TestPersistentVolumeReclaimPolicyIdentifier, `Check that containers have resource requests and limits specified in their spec.`),
```
greyerof, line 950: This description seems to be a copy/paste from `TestPodRequestsAndLimitsIdentifier`.

pkg/autodiscover/autodiscover_pv.go
```diff
@@ -0,0 +1,33 @@
...
+import (
+	"context"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+)
+
+func getPersistentVolumes(oc corev1client.CoreV1Interface) ([]corev1.PersistentVolume, error) {
+	pvs, err := oc.PersistentVolumes().List(context.TODO(), metav1.ListOptions{})
```
greyerof, line 28: This will discover all PVs in the cluster, but shouldn't we test only the PVs that the CNF's pods/containers are actually using through a PVC?

sebrandon1: True, I'll add that in the actual test.  Right now I wasn't filtering it.

cnf-certification-test/identifiers/identifiers.go
sebrandon1: Yes, good catch.  Fixed.

pkg/autodiscover/autodiscover_pv.go
sebrandon1: Test updated.

cnf-certification-test/lifecycle/suite.go
```diff
@@ -436,3 +441,24 @@ func testPodRequestsAndLimits(env *provider.TestEnvironment) {
...
 }
+
+func testPodPersistentVolumeReclaimPolicy(env *provider.TestEnvironment) {
+	ginkgo.By("Testing PersistentVolumes for reclaim policy to be set to delete")
+	var persistentVolumesBadReclaim []string
+
+	// Look through all of the pods, matching their persistent volumes to the list of overall cluster PVs and checking their reclaim status.
+	for _, put := range env.Pods {
+		for index := range env.PersistentVolumes {
+			if put.Data.Name == env.PersistentVolumes[index].Name && env.PersistentVolumes[index].Spec.PersistentVolumeReclaimPolicy != corev1.PersistentVolumeReclaimDelete {
```
greyerof, line 452: I don't understand this part of the check: `put.Data.Name == env.PersistentVolumes[index].Name`. If I understood PVs correctly, they are separate resources/objects with their own names, so this `if` probably won't match for any PV in the cluster, unless the customer is forced to use the same name for pods and PVs. Am I missing something?

sebrandon1: Ah yes!  You are correct.  I had to add a loop to look at the `Volumes` and their names.

greyerof (approved):

pkg/provider/provider.go
```diff
@@ -100,10 +100,11 @@ func ConvertArrayPods(pods []*corev1.Pod) (out []*Pod) {
...
 type TestEnvironment struct { // rename this with testTarget
-	Namespaces        []string               `json:"testNamespaces"`
-	Pods              []*Pod                 `json:"testPods"`
-	Containers        []*Container           `json:"testContainers"`
-	Operators         []Operator             `json:"testOperators"`
+	Namespaces        []string     `json:"testNamespaces"`
+	Pods              []*Pod       `json:"testPods"`
+	Containers        []*Container `json:"testContainers"`
+	Operators         []Operator   `json:"testOperators"`
+	PersistentVolumes []corev1.PersistentVolume
```
jmontesi, line 107: Don't you need to get the `PersistentVolumes` from the autodiscover package inside `buildTestEnvironment()`?

sebrandon1: Good catch.  Fixed!

jmontesi (approved):

PR #402, opened 2022-08-04
Change daemonset image to use partner repo if provided

ramperher:
build-depends: 25728

**Context**

This change merged after releasing tnf v4.0.2 changed the way debug daemonset is created. In particular, the image to be used is fixed, pointing to quay.io: https://github.com/test-network-function/cnf-certification-test/commit/5e4ef828e22fc6cff0b6b9b49d27426d8a4f9617#diff-75ec911f14b9ea61000ce3840f3deee935c88181bef148b6cbfdabc4833b9c4eR17

This is breaking deployments on disconnected environments. 

**Change**

- Use the `TNF_PARTNER_REPO` and `SUPPORT_IMAGE` variables if provided to update the `imageWithVersion` variable.
- Update `run-container.sh` and `README.md` to include `SUPPORT_NAME` environment variable.
- Update `README.md` with the usage of `OPENSHIFT_VERSION` variable to build the container, because this command is failing due to its absence.


sebrandon1:
Good catch on the disconnected environments.  Looks like its having trouble with the logic of setting the variable.

internal/daemonset/daemonset.go
```diff
@@ -14,7 +15,8 @@ import (
 )
 
 const (
-	imageWithVersion                   = "quay.io/testnetworkfunction/debug-partner:latest"
+	imageWithVersion                   = os.Getenv("TNF_PARTNER_REPO") != "" ?
```
greyerof, line 18: Good catch! A couple of comments: there's no ternary operator in golang "?", so you'll need to create or use an existing function to set this package variable. Also, The image name should also be optionally set by another env var,[ like in the old yaml](https://github.com/test-network-function/cnf-certification-test-partner/blob/d250d1acf1091e82efd6e36e2100594eabc1975c/test-partner/debugpartner.yaml#L25) (SUPPORT_IMAGE).

ramperher: Sure, I'll check it now, I just started the change yesterday to review it with more depth today. I'll keep you posted

ramperher: Done, I've included this in my last commit. I'll start testing that on DCI.

ramperher:
I'm having issues with the testing of this change in our labs. It seems to be a problem when cloning the repository in our labs in case of having merged a change in the main repo. I'm checking in parallel

ramperher:
This change is ready for review, even though I will continue testing it on DCI to have a SUCCESS job next Monday.

sebrandon1 (approved):

internal/daemonset/daemonset.go
```diff
@@ -187,8 +189,28 @@ func CreateDaemonSet(daemonSetName, namespace, containerName, imageWithVersion s
...
+	if os.Getenv("TNF_PARTNER_REPO") != "" {
+		tnfPartnerRepo = os.Getenv("TNF_PARTNER_REPO")
+	} else {
+		tnfPartnerRepo = tnfPartnerRepoDef
+	}
+	if os.Getenv("SUPPORT_IMAGE") != "" {
+		supportImage = os.Getenv("SUPPORT_IMAGE")
+	} else {
+		supportImage = supportImageDef
+	}
```
rdavid, line 206: I'd write less strings:
```
tnfPartnerRepo := os.Getenv("TNF_PARTNER_REPO")
if tnfPartnerRepo == "" {
  tnfPartnerRepo = tnfPartnerRepoDef
}
supportImage := os.Getenv("SUPPORT_IMAGE")
if supportImage == "" {
  supportImage = supportImageDef
}
```

rdavid (approved):

internal/daemonset/daemonset.go
ramperher: I tried in previous commits but it failed on lint check, I'll try it again to see if it works, also checking that is not breaking our jobs

greyerof (approved):

internal/daemonset/daemonset.go
ramperher: Included, and seems that lint is working, thanks!

ramperher:
Thanks for the review! I'll launch a last DCI job to confirm that the daemonset is working fine, then I'll merge this.

PR #404, opened 2022-08-04
Add scenario-based labeling

sebrandon1:
We have been discussing how we could potentially break up some of the test cases to only run in certain scenarios where they are deemed applicable.  For example, there could be two separate CNF workloads "xyz" and "abc" that might have certain requirements as far as tests but they also might have tests that overlap.  We should be able to label individual tests to belong to certain scenarios/CNFs (DU, CU, etc) and let them be configurable.

Marked WIP for discussion.

See https://github.com/test-network-function/test-network-function-claim/pull/57

I am updating the -claim repo to add a `tags` section into the `Identifier` struct.  This will let us be able to define certain tags/scenarios for when each test is ran.

### DIFF
--- a/pkg/configuration/configuration.go
+++ b/pkg/configuration/configuration.go
@@ -48,6 +48,11 @@ type AcceptedKernelTaintsInfo struct {
 	Module string `yaml:"module" json:"module"`
 }
 
+// TestScenarioLabels will help determine which set of tests we want to run for different types of scenarios
+type TestScenarioLabels struct {
+	Scenario string `yaml:"scenario" json:"scenario"`
+}
+
 // Label ns/name/value for resource lookup
 type Label struct {
 	Prefix string `yaml:"prefix" json:"prefix"`
@@ -98,6 +103,8 @@ type TestConfiguration struct {
 	SkipHelmChartList    []SkipHelmChartList        `yaml:"skipHelmChartList" json:"skipHelmChartList"`
 	// CheckDiscoveredContainerCertificationStatus controls whether the container certification test will validate images used by autodiscovered containers, in addition to the configured image list
 	CheckDiscoveredContainerCertificationStatus bool `yaml:"checkDiscoveredContainerCertificationStatus" json:"checkDiscoveredContainerCertificationStatus"`
+	// Scenarios is a configurable set of labels that are applied to certain tests that only desired ran in certain scenarios
+	Scenarios []TestScenarioLabels `yaml:"scenarioLabels,omitempty" json:"scenarioLabels,omitempty"`
 }
 
 type TestParameters struct {


greyerof:
This could be a simple implementation of using tags (@jmontesi 's original idea):

We need a tags list to be passed to the tnf, either by command line arg "--tags tag1,tag2,tag3" as we're currently doing with ginkgo's focus/skip/labels options, or by parsing a new field in the config yaml file (testTags: -tag1 -tag2).

1. Add a new hardcoded list of tags per TC. On each identifier, add a new (corresponding) tags slice. E.g:
```
	TestHugepagesNotManuallyManipulated = claim.Identifier{
		Url:     formTestURL(common.PlatformAlterationTestKey, "hugepages-config"),
		Version: versionOne,
		Tags: map[testTag]bool{testTagCommon: true, testTagRanFarEdge: true}  // map type just for convenience
	}
```
2. For each TC, just skip unless the proper tag was passed to the cli:
```
		testhelper.SkipIfEmptyAny(ginkgo.Skip, env.Containers)
		testhelper.SkipIfNotTagged(identifier, env.TargetTags, ginkgo.Skip)
		testTerminationMessagePolicy(&env)
	})
```
Where `SkipIfNotTagged` could be something similar to this:
```
SkipIfNotTagged(id claim.Identifier, targetTags []testTags, skipFn func(string, ...int)) {
	if len(testTags) == 0 {
		return   // All TCs can be executed: backward compatibility.
	}
	
	for testTag := range testTags {
		if id[testTag] {
			return 
		}
	}
	
	skipFn("TC not tagged.")
}

```	
So, in the configuration.go file, we only need a list of tags like:
``` 
type testTag string

const (
	testTagCommon:      "common"
	testTagRan:         "ran-far"
	testTagRanFarEdge:  "ran-far-edge"
)
```
What do you think?

jmontesi:
I think @greyerof's proposal reflects quite well what I had in mind about using tags for splitting tests into different areas. I like that there is no need to modify how ginkgo is called by using these "private tags" (from ginkgo point of view) but at the same time there are some other things I do like that much, such as:

- Increases the "code density" due to the fact that every test needs to call the `SkipIfNotTagged()` function.
- When the tests for a particular "private" tag (such as "ran-far") are run, there will be some tests which will be focused but that will not run because they don't have the "private" tag. But they still will appear as skipped, producing an unnecessary log output.
- The fact that we are introducing the distinction between a "private" tag and a ginkgo label when they are very similar.

Another proposal could be to keep the step 1 from @greyerof's (maybe using a slice instead of a map). And then write the test calls like this:

 ```
tags := convertToLabelString(<tags from step 1 struct>)
ginkgo.It(testID, ginkgo.Label(tags), func() {
	testhelper.SkipIfEmptyAny(ginkgo.Skip, env.Containers)
	testTerminationMessagePolicy(&env)
})
```
This requires modifying the way ginkgo is called to add a more precise label combination (see [Spec Labels](https://onsi.github.io/ginkgo/#spec-labels)) but offers an unified labeling/tagging system for suites and technical domains.

Let me know what you think.

sebrandon1:
Good suggestions.
1) I like having the tags listed in the configuration file.  I'm guessing that the partners will like that as well.
2) For the `Add a new hardcoded list of tags per TC.` I believe I started that in [PR#57](https://github.com/test-network-function/test-network-function-claim/pull/57).  To make it an `array` type in the JSON scheme it turns the Go translation into []interface{}.
3) If we go with the ginkgo labeling scheme, does that clash with how we are kicking off the test suite in general?  Don't you have to pass it in as a command line variable like we are doing [here](https://github.com/test-network-function/cnf-certification-test/blob/main/run-cnf-suites.sh#L80)?  I think I like @greyerof's suggestion to use `SkipIfNotTagged` to handle it all in the code to skip certain tests.

greyerof:
@jmontesi The idea of re-using ginkgo's labels looks way more cleaner, but that couples us (again) even more to ginkgo, which I don't like too much. As the project grows, I'm more inclined to believe that the yaml config file is a better place to add these kind of config switches. However, **your way can be a good and easy/quick starting point right now**. Regarding your concern about skipped TCs, using ginkgo labels as you propose will also mark as skipped all the TCs focused that don't match the labels. The -f is mandatory. If I'm not wrong, ginkgo labels only work as a filter for TCs that ginkgo found with the "-f" flag.
@sebrandon1 regarding your concern, for CI jobs it won't be necessary to update the CLI args, as ginkgo labels are a way to filter TCs found with the -f (focus) flag, as I said before. But we might want to add more workflows to test single or multiple labels...

edcdavid (commented):
Why not do this in the ./run-cnf-suites.sh script? Don't we already have the capability to add and remove TCs from every run ? We could pass a "profile" name that would simply paste the right combination of focus and tags, for instance -ran, -du, etc.

sebrandon1:
Closing in favor of: https://github.com/test-network-function/cnf-certification-test/pull/444

PR #407, opened 2022-08-05
Add QE failure to unit tests for tolerations

sebrandon1:
QE has been running the new toleration test added in #365 and has been hitting a failure because pods have:

```
// - effect: NoSchedule
//   key: node.kubernetes.io/memory-pressure
//   operator: Exists
```

### DIFF
--- a/cnf-certification-test/accesscontrol/tolerations/tolerations_test.go
+++ b/cnf-certification-test/accesscontrol/tolerations/tolerations_test.go
@@ -78,6 +78,14 @@ func TestIsTolerationModified(t *testing.T) {
 			},
 			expectedOutput: false,
 		},
+		{ // Test Case #6 - example from QE
+			testToleration: v1.Toleration{
+				Key:      "node.kubernetes.io/memory-pressure",
+				Operator: v1.TolerationOpExists,
+				Effect:   v1.TaintEffectNoSchedule,
+			},
+			expectedOutput: true,
+		},
 	}
 
 	for _, tc := range testCases {


sebrandon1:
After some discussion, we _might_ have to add an allowlist for certain pods to pass the toleration check... or maybe QE might have to switch to an operator that does not add additional tolerations.

ramperher:
Hi @sebrandon1, in fact I am hitting the same issue in our current DCI jobs. If you see [this last job](https://www.distributed-ci.io/jobs/8ce96e71-dd43-478a-b204-545d5ba509fd/tests) that comes from https://github.com/test-network-function/cnf-certification-test/pull/402, you will see this is failing because of `NoSchedule` toleration:

```
Pod: NoSchedule has been found with non-default pod: test-85bcb64d66-7dkxz ns: test-cnf toleration NoSchedule which is not allowed.
Pod: NoSchedule has been found with non-default pod: test-85bcb64d66-8snqs ns: test-cnf toleration NoSchedule which is not allowed.
Pod: NoSchedule has been found with non-default pod: test-6dbbcfd84c-9zlpt ns: production-cnf toleration NoSchedule which is not allowed.
Pod: NoSchedule has been found with non-default pod: test-6dbbcfd84c-knq8t ns: production-cnf toleration NoSchedule which is not allowed.
Number of pods found with non-compliant tolerations: 4
```

Checking the pods, you can see we are hitting exactly the same error:

```
[kni@provisioner.cluster5.dfwt5g.lab ~]$ oc describe pod -n test-cnf test-85bcb64d66-jw6vk
...
Tolerations:        node.kubernetes.io/memory-pressure:NoSchedule op=Exists
```

However, we are using the same pod under test your team have defined [here](https://github.com/test-network-function/cnf-certification-test-partner/blob/main/test-target/local-pod-under-test.yaml), so we are not defining any extra toleration to the pods. It seems that k8s include them automatically depending on the cluster status.

Would you mind if I can test some DCI jobs using your change to confirm that this change makes the test to pass in our labs? You will see some dcibot messages in the change

sebrandon1:
Interesting, yes it seems like `node.kubernetes.io/memory-pressure:NoSchedule` must get added automatically or something by both your cluster and QE.  Reviewing [the tolerations doc](https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/#taint-nodes-by-condition) it seems like there's a way for pods with a QoS class to get an automatic memory toleration added?  Does that apply in your cluster?

sebrandon1:
Yes go ahead and run tests against this PR that's fine.

ramperher:
> Interesting, yes it seems like `node.kubernetes.io/memory-pressure:NoSchedule` must get added automatically or something by both your cluster and QE. Reviewing [the tolerations doc](https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/#taint-nodes-by-condition) it seems like there's a way for pods with a QoS class to get an automatic memory toleration added? Does that apply in your cluster?

I'll check this if I can add this kind of toleration to avoid the taint, thanks!

ramperher:
I've finally created this change to handle this toleration issue: https://github.com/test-network-function/cnf-certification-test/pull/417, I didn't know if I could make changes in your PR, @sebrandon1 , or create a new, so I did the second finally, just in case :)

sebrandon1:
@ramperher thank you for the PR!  I'll close mine and point to #417

PR #409, opened 2022-08-08
Fix toleration err message

sebrandon1:
Whoops, I accidentally wasn't printing the pod name where it should have been.

### DIFF
--- a/cnf-certification-test/accesscontrol/suite.go
+++ b/cnf-certification-test/accesscontrol/suite.go
@@ -581,7 +581,7 @@ func TestPodTolerationBypass(env *provider.TestEnvironment) {
 			// Check if the tolerations fall outside the 'default' and are modified versions
 			if tolerations.IsTolerationModified(t) {
 				podsWithRestrictedTolerationsNotDefault = append(podsWithRestrictedTolerationsNotDefault, put.String())
-				tnf.ClaimFilePrintf("Pod: %s has been found with non-default %s toleration %s which is not allowed.", string(t.Effect), put.String(), t.Effect)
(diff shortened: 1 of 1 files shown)


jmontesi (approved):

greyerof (approved):

cnf-certification-test/accesscontrol/suite.go
```diff
@@ -581,7 +581,7 @@ func TestPodTolerationBypass(env *provider.TestEnvironment) {
 			// Check if the tolerations fall outside the 'default' and are modified versions
 			if tolerations.IsTolerationModified(t) {
 				podsWithRestrictedTolerationsNotDefault = append(podsWithRestrictedTolerationsNotDefault, put.String())
-				tnf.ClaimFilePrintf("Pod: %s has been found with non-default %s toleration %s which is not allowed.", string(t.Effect), put.String(), t.Effect)
+				tnf.ClaimFilePrintf("Pod: %s has been found with non-default %s toleration %s which is not allowed.", put.Data.Name, put.String(), t.Effect)
```
greyerof, line 584: The stringer for put (put.String()) is already showing the name of the pod, so that first "put.Data.Name" is not needed.
```
func (p *Pod) String() string {
	return fmt.Sprintf("pod: %s ns: %s",
		p.Data.Name,
		p.Data.Namespace,
	)
}
```
Are you sure this new version of the trace is ok?

sebrandon1: I'll refactor since we can use the .String()

rdavid (approved):

PR #417, opened 2022-08-10
Do not fail if memory-pressure toleration is present and qosClass is different than BestEffort

ramperher:
**Context**

In [this change](https://github.com/test-network-function/cnf-certification-test/pull/407), it was discussed that there is a toleration that seems to be added by k8s and is not being treated by the `access-control-pod-toleration-bypass`, so that it is not considered as a modified toleration.

The reason why this memory-issue toleration is appearing is because QoS class applied to the pods under test. If resource limits (for CPU and RAM) are defined, according to this [k8s docs](https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/#taint-nodes-by-condition), it is said that "The control plane also adds the node.kubernetes.io/memory-pressure toleration on pods that have a [QoS class](https://kubernetes.io/docs/reference/glossary/?all=true#term-qos-class) other than BestEffort".

So, if the QoS class applied to the pod is different than BestEffort, this toleration should not be considered as modified and the access-control test must pass in that case.

**Changes included**

- Include QoSClass as a parameter to be interpreted by toleration check
- Also consider that a toleration is not modified if we have NoSchedule - memory-pressure toleration with QoS class different than BestEffort, following k8s docs
- New unit tests

### DIFF
--- a/cnf-certification-test/accesscontrol/tolerations/tolerations.go
+++ b/cnf-certification-test/accesscontrol/tolerations/tolerations.go
@@ -55,7 +55,7 @@ func IsTolerationModified(t v1.Toleration, qosClass v1.PodQOSClass) bool {
 		// If toleration is NoSchedule - node.kubernetes.io/memory-pressure - Exists and the QoS class for
 		// the pod is different than BestEffort, it is also a default toleration added by k8s
 		if (t.Key == memoryPressureStr) &&
-			(t.Operator == v1.TolerationOpExists && t.TolerationSeconds == nil) &&
+			(t.Operator == v1.TolerationOpExists) &&
 			(qosClass != v1.PodQOSBestEffort) {
 			return false
 		}
--- a/cnf-certification-test/accesscontrol/tolerations/tolerations.go
+++ b/cnf-certification-test/accesscontrol/tolerations/tolerations.go
@@ -50,11 +50,13 @@ func IsTolerationModified(t v1.Toleration, qosClass v1.PodQOSClass) bool {
 		if t.Key == notReadyStr || t.Key == unreachableStr &&
 			(t.Operator == v1.TolerationOpExists && t.TolerationSeconds != nil && *t.TolerationSeconds == int64(tolerationSecondsDefault)) {
 			return false
-		} else if (t.Key == memoryPressureStr) &&
+		}
+	} else if t.Effect == v1.TaintEffectNoSchedule {
+		// If toleration is NoSchedule - node.kubernetes.io/memory-pressure - Exists and the QoS class for
+		// the pod is different than BestEffort, it is also a default toleration added by k8s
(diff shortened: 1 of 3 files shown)


ramperher:
> * SUCCESS https://www.distributed-ci.io/jobs/32157cb6-a08e-4661-b114-db7afd5def5b/jobStates
>   from [Do not fail if memory-pressure toleration is present and qosClass is different than BestEffort #417](https://github.com/test-network-function/cnf-certification-test/pull/417)

This change makes our job to work again :) because, for our tnf testing pipeline, we were making the job to fail if any tnf unit test was failing, and the only one that was still failing was `access-control-pod-toleration-bypass`. In this last DCI job, we can see it is passing thanks to this change.

I'd say this is ready for review, CC @sebrandon1 @greyerof

sebrandon1 (approved):

cnf-certification-test/accesscontrol/tolerations/tolerations.go
```diff
@@ -40,12 +41,24 @@ func IsTolerationModified(t v1.Toleration) bool {
...
 	if t.Effect == v1.TaintEffectNoExecute {
 		if t.Key == notReadyStr || t.Key == unreachableStr &&
 			(t.Operator == v1.TolerationOpExists && t.TolerationSeconds != nil && *t.TolerationSeconds == int64(tolerationSecondsDefault)) {
 			return false
 		}
+	} else if t.Effect == v1.TaintEffectNoSchedule {
+		// If toleration is NoSchedule - node.kubernetes.io/memory-pressure - Exists and the QoS class for
+		// the pod is different than BestEffort, it is also a default toleration added by k8s
+		if (t.Key == memoryPressureStr) &&
+			(t.Operator == v1.TolerationOpExists && t.TolerationSeconds == nil) &&
```
greyerof, line 58: I don't think the `t.TolerationSeconds == nil` check is needed here. As per k8s docs, that field is only used for NoExecute effects.

ramperher: let's remove that, then, I just included it to make sure it is not defined, but if it's not related to NoSchedule, let's remove it

ramperher:
> Unable to find -e pipeline

Do not pay much attention to this and further messages, because I'll create some other too, I am testing some automations in our labs to be able to catch PRs from this repo, so that we can test them in our labs to verify if they are fine or not. But it's not affecting the change, the change is, in fact, ready for merging.

greyerof (approved):

ramperher:
@sebrandon1 @greyerof If someone of you can merge it, it would be great, because I don't have permissions to do that, I'll ask in your gchat if I can have them too for the future, as I intend to collaborate if finding bugs and issues while testing on DCI. Thanks for your review by the way!

PR #418, opened 2022-08-10
cnfextentions feature

edcdavid:
this is to import the cnfextensions (https://github.com/test-network-function/cnfextensions) project as an example of partner private test case.
The following PR does the following:
- capitalizes a few string constants so that they can be used in the extended repo
- adds a new import line in cnf-certification-test/suite_test.go to run the new cnfextensions example suite
- adds the "cnfextensions" suite in CI


sebrandon1:
As a suggestion we could probably use `cnfextensions` as an external library brought in via go.mod to be able to properly version in rather than just using a locked in git hash.

edcdavid:
> As a suggestion we could probably use `cnfextensions` as an external library brought in via go.mod to be able to properly version in rather than just using a locked in git hash.

@sebrandon1 I removed the submodule, much better this way, it was not needed. Thanks for the suggestion!

sebrandon1:
Perfect!  Yeah I think this will be better than a submodule for sure.

sebrandon1:
As far as this PR goes, do we need to keep it open or do we wait for further partner input as to how we want to handle "extensions"?

greyerof (approved):

sebrandon1 (approved):

PR #419, opened 2022-08-10
Move autodiscover to external library

sebrandon1:
Just as a PoC, I decided to see how much of a problem it would be to move the `autodiscover` package to its own external library.

I had to move the `clientsholder` and `configuration` packages as well.  The thinking is that autodiscover needs access to the clients and the configuration but we can also consume those via the `autodiscover` library.


jmontesi:
I wouldn't move the autodiscovery module to an external library at the moment unless we know that there are several projects interested in using it. Otherwise it just adds extra work in order to have this module in sync with the CNF Certification.

sebrandon1:
Yup agreed.  Just wanted to you to take a look before closing @greyerof.  Thanks!

PR #434, opened 2022-08-18
Add CPU isolation test

sebrandon1:
Adds a test for CPU isolation that only runs when enabled via configuration.  I added the configuration variable to `TestParameters` but I'm not really sure if that's where we want to keep it.  I'll leave that open for discussion.

### DIFF
--- a/cnf-certification-test/lifecycle/suite.go
+++ b/cnf-certification-test/lifecycle/suite.go
@@ -149,7 +149,7 @@ var _ = ginkgo.Describe(common.LifecycleTestKey, func() {
 		testPodPersistentVolumeReclaimPolicy(&env)
 	})
 
-	testID, tags = identifiers.GetGinkgoTestIDAndLabels(identifiers.TestCPUIsolation)
+	testID, tags = identifiers.GetGinkgoTestIDAndLabels(identifiers.TestCPUIsolationIdentifier)
 	ginkgo.It(testID, ginkgo.Label(tags...), func() {
 		testhelper.SkipIfEmptyAny(ginkgo.Skip, env.Pods)
 		testCPUIsolation(&env)
@@ -495,23 +495,23 @@ func testCPUIsolation(env *provider.TestEnvironment) {
 	ginkgo.By("Testing pods for CPU isolation requirements")
 
 	// Individual requirements we are looking for:
-	// - Annotations must be provided disabling CPU and IRQ load-balancing.
+	//  - CPU Requests and Limits must be in the form of whole units
 	// - Resource Requests and Limits must be provided and identical
 
 	// Additional checks if the above pass
-	// - CPU Requests and Limits must be in the form of whole units
 	// - 'runtimeClassName' must be specified
+	// - Annotations must be provided disabling CPU and IRQ load-balancing.
 
 	podsMissingIsolationRequirements := make(map[string]bool)
 
 	for _, put := range env.Pods {
-		if put.IsCPUIsolationEnabled() && !put.IsCPUIsolationCompliant() {
+		if put.IsPodGuaranteed() && !put.IsCPUIsolationCompliant() {
 			podsMissingIsolationRequirements[put.Data.Name] = true
 		}
 	}
 
 	if n := len(podsMissingIsolationRequirements); n > 0 {
-		errMsg := fmt.Sprintf("Number of pods found that are not compliant with CPU isolation requirements: %d. See logs for more detail.", n)
+		errMsg := fmt.Sprintf("Number of guaranteed pods found that are not compliant with CPU isolation requirements: %d. See logs for more detail.", n)
 		tnf.ClaimFilePrintf(errMsg)
 		ginkgo.Fail(errMsg)
 	}
--- a/cnf-certification-test/lifecycle/suite.go
+++ b/cnf-certification-test/lifecycle/suite.go
@@ -24,6 +24,7 @@ import (
 	"github.com/sirupsen/logrus"
 	"github.com/test-network-function/cnf-certification-test/cnf-certification-test/common"
 	"github.com/test-network-function/cnf-certification-test/cnf-certification-test/identifiers"
+	"github.com/test-network-function/cnf-certification-test/cnf-certification-test/lifecycle/isolation"
 	"github.com/test-network-function/cnf-certification-test/cnf-certification-test/lifecycle/ownerreference"
 	"github.com/test-network-function/cnf-certification-test/cnf-certification-test/lifecycle/podrecreation"
 	"github.com/test-network-function/cnf-certification-test/cnf-certification-test/lifecycle/podsets"
@@ -149,6 +150,11 @@ var _ = ginkgo.Describe(common.LifecycleTestKey, func() {
 		testPodPersistentVolumeReclaimPolicy(&env)
 	})
 
+	testID, tags = identifiers.GetGinkgoTestIDAndLabels(identifiers.TestCPUIsolation)
+	ginkgo.It(testID, ginkgo.Label(tags...), func() {
+		testhelper.SkipIfEmptyAny(ginkgo.Skip, env.Pods)
+		testCPUIsolation(&env)
+	})
 })
 
 func testContainersPreStop(env *provider.TestEnvironment) {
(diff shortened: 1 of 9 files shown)


cnf-certification-test/lifecycle/suite.go
```diff
@@ -147,6 +148,13 @@ var _ = ginkgo.Describe(common.LifecycleTestKey, func() {
 	ginkgo.It(testID, ginkgo.Label(testID), func() {
 		testPodPersistentVolumeReclaimPolicy(&env)
 	})
+
+	testID = identifiers.XformToGinkgoItIdentifier(identifiers.TestCPUIsolation)
+	ginkgo.It(testID, ginkgo.Label(testID), func() {
+		if provider.CPUIsolationTestingEnabled() {
```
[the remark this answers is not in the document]
sebrandon1: Fixed.

cnf-certification-test/lifecycle/suite.go
```diff
@@ -147,6 +148,14 @@ var _ = ginkgo.Describe(common.LifecycleTestKey, func() {
 	ginkgo.It(testID, ginkgo.Label(testID), func() {
 		testPodPersistentVolumeReclaimPolicy(&env)
 	})
+
+	testID = identifiers.XformToGinkgoItIdentifier(identifiers.TestCPUIsolation)
+	ginkgo.It(testID, ginkgo.Label(testID), func() {
+		testhelper.SkipIfEmptyAny(ginkgo.Skip, env.Pods)
+		if provider.CPUIsolationTestingEnabled() {
```
edcdavid, line 155: We should detect isolated pods individually by checking if request cpu == limit cpu, instead of making it a test. Not all the pods in a namespace will be isolated. The CNF is indicating isolation by making the request and limit equal.

sebrandon1: Okay I can change it up and detect the isolated pods.

sebrandon1: Okay per the spec again, the last line of the requirement says: `DOORS Requirement: In order to use Isolated CPUs, specific annotations must be defined in the pod spec` so I switched up my logic to check for those annotations first to distinguish what is/isn't an isolated pod.

cnf-certification-test/lifecycle/suite.go
```diff
@@ -484,3 +490,51 @@ func testPodPersistentVolumeReclaimPolicy(env *provider.TestEnvironment) {
...
+	// Individual requirements we are looking for:
+	// - Resource Requests and Limits must be provided and identical. This determines whether or not we run the rest of the tests.
+	// - CPU Requests and Limits must be in the form of whole units
+	// - 'runtimeClassName' must be specified
+	// - Annotations must be provided disabling CPU and IRQ load-balancing
+
+	podsMissingIsolationRequirements := make(map[string]bool)
+
+	for _, put := range env.Pods {
+		it := isolation.NewCPUIsolationTester(put)
```
edcdavid, line 506: I would rename it with something else (I confused it with ginkgo "It"). Does it mean isolation test ?

cnf-certification-test/lifecycle/suite.go
```diff
@@ -484,3 +490,51 @@ func testPodPersistentVolumeReclaimPolicy(env *provider.TestEnvironment) {
...
+
+		// Pods that have identical resources, are considered CPU isolated.
+		if it.LoadBalancingDisabled() {
```
edcdavid, line 509: Shouldn't it be it.AreResourcesIdentical() here ?

cnf-certification-test/lifecycle/suite.go
```diff
@@ -484,3 +490,51 @@ func testPodPersistentVolumeReclaimPolicy(env *provider.TestEnvironment) {
...
+			}
+
+			if !it.IsRuntimeClassNameSpecified() {
+				errMsg := fmt.Sprintf("%s has been found to not have runtimeClassName specified.", put.String())
+				logrus.Debugf(errMsg)
+				tnf.ClaimFilePrintf(errMsg)
+				podsMissingIsolationRequirements[put.Data.Name] = true
+			}
+
+			if !it.AreResourcesIdentical() {
```
edcdavid, line 524: Same comment, shouldn't it be  it.LoadBalancingDisabled() here

cnf-certification-test/lifecycle/suite.go
```diff
@@ -484,3 +490,51 @@ func testPodPersistentVolumeReclaimPolicy(env *provider.TestEnvironment) {
...
+				errMsg := fmt.Sprintf("%s has been found to not have identical resources specified.", put.String())
+				logrus.Debugf(errMsg)
+				tnf.ClaimFilePrintf(errMsg)
+				podsMissingIsolationRequirements[put.Data.Name] = true
+			}
+		} else {
+			logrus.Debugf("%s is not an isolated pod. Skipping.", put.String())
```
edcdavid, line 531: I would keep track of skipped pods and compliant pods using isolated cpus as well. At the end:
- if all the pods are skipped, no pods are compliant isolated, no pods are non compliant isolated -> skip the test because it does not apply for this CNF
- if some pods are skipped some are compliant/isolated and none is non-compliant -> pass the test
- if some pods are non compliant -> fail the test

cnf-certification-test/lifecycle/suite.go
sebrandon1: Yeah I was just thinking I-solation T-ester 😄 I can change it up.

cnf-certification-test/lifecycle/suite.go
sebrandon1: Nope, the DOORS requirement says that it's the annotations that determine the isolation status.

cnf-certification-test/lifecycle/suite.go
sebrandon1: Nope I switched them because the annotations determine the CPU isolated pod.

sebrandon1: I added `put.IsCPUIsolationEnabled() && !put.IsCPUIsolationCompliant()` as the check.

cnf-certification-test/lifecycle/suite.go
sebrandon1: I added `put.IsCPUIsolationEnabled() && !put.IsCPUIsolationCompliant()` as the check.

pkg/provider/isolation.go
```diff
@@ -0,0 +1,128 @@
...
+			return false
+		}
+	}
+
+	return true
+}
+
+func AreCPUResourcesWholeUnits(p *Pod) bool {
+	isInteger := func(val float64) bool {
+		return val == float64(int(val))
```
greyerof, line 62: Just a suggestion: I always have a lot of doubts about number castings and I know this should work, but messing with float <-> int castings might produce weird results in some cases... For this case I'd rather use the available `milli` value (int64) and check against the remainder of the division by 1000.
```
  cpuRequestsIsInteger := cut.Data.Resources.Requests.Cpu().MilliValue() % 1000 == 0
  cpuLimitsIsInteger := cut.Data.Resources.Limits.Cpu().MilliValue() % 1000 == 0
```

pkg/provider/isolation.go
```diff
@@ -0,0 +1,128 @@
...
+// with this program; if not, write to the Free Software Foundation, Inc.,
+// 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+package provider
+
+import (
+	"github.com/sirupsen/logrus"
+)
+
+type TesterFuncs interface {
```
greyerof, line 23: Is this interface really needed? I don't see it being used anywhere.

pkg/provider/provider.go
```diff
@@ -499,6 +500,30 @@ func (p *Pod) String() string {
 	)
 }
 
+func (p *Pod) IsCPUIsolationEnabled() bool {
+	return LoadBalancingDisabled(p) && AreResourcesIdentical(p)
```
edcdavid, line 504: I would change it to the following:
return AreCPUResourcesWholeUnits(p) && AreResourcesIdentical(p)
name of the function IsCPUIsolationEnabled -> IsPodGuaranteed

cnf-certification-test/lifecycle/suite.go
```diff
@@ -484,3 +489,29 @@ func testPodPersistentVolumeReclaimPolicy(env *provider.TestEnvironment) {
...
+	// - Resource Requests and Limits must be provided and identical
+
+	// Additional checks if the above pass
+	// - CPU Requests and Limits must be in the form of whole units
+	// - 'runtimeClassName' must be specified
+
+	podsMissingIsolationRequirements := make(map[string]bool)
+
+	for _, put := range env.Pods {
+		if put.IsCPUIsolationEnabled() && !put.IsCPUIsolationCompliant() {
```
edcdavid, line 507: IsCPUIsolationEnabled -> IsGaranteedPod

cnf-certification-test/lifecycle/suite.go
```diff
@@ -484,3 +489,29 @@ func testPodPersistentVolumeReclaimPolicy(env *provider.TestEnvironment) {
...
+			podsMissingIsolationRequirements[put.Data.Name] = true
+		}
+	}
+
+	if n := len(podsMissingIsolationRequirements); n > 0 {
+		errMsg := fmt.Sprintf("Number of pods found that are not compliant with CPU isolation requirements: %d. See logs for more detail.", n)
```
edcdavid, line 513: Number of pods found -> Number of guaranteed pods found

pkg/provider/isolation.go
sebrandon1: Good call, it's better to be explicit.  I'll update.

pkg/provider/isolation.go
sebrandon1: Nope, not needed.  I had it there originally when I was scaffolding the funcs and I was thinking about testing.   Will remove.

cnf-certification-test/lifecycle/suite.go
sebrandon1: Fixed.

cnf-certification-test/lifecycle/suite.go
sebrandon1: Fixed.

pkg/provider/provider.go
```diff
@@ -499,6 +500,30 @@ func (p *Pod) String() string {
...
+}
+
+func (p *Pod) IsCPUIsolationCompliant() bool {
+	isCPUIsolated := true
+
+	if !AreCPUResourcesWholeUnits(p) {
```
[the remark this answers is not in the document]
sebrandon1: Done.

pkg/provider/provider.go
sebrandon1: Adjusted.

cnf-certification-test/lifecycle/suite.go
```diff
@@ -484,3 +489,29 @@ func testPodPersistentVolumeReclaimPolicy(env *provider.TestEnvironment) {
...
+	// - Resource Requests and Limits must be provided and identical
+
+	// Additional checks if the above pass
+	// - 'runtimeClassName' must be specified
+	// - Annotations must be provided disabling CPU and IRQ load-balancing.
+
+	podsMissingIsolationRequirements := make(map[string]bool)
+
+	for _, put := range env.Pods {
+		if put.IsPodGuaranteed() && !put.IsCPUIsolationCompliant() {
```
edcdavid, line 507: optional: split this if and count the # of guaranteed pods so that if it is zero, the test can be skipped instead of passing (e.g. no isolated pods to test). Something like: 
var guaranteedPods int
if put.IsPodGuaranteed() {
guaranteedPods++
if !put.IsCPUIsolationCompliant() {

edcdavid (approved):

cnf-certification-test/lifecycle/suite.go
sebrandon1: Either way, it'll be evaluating the `IsPodGuaranteed` for each pod correct?  I think I like this better than counters?  What do you think?

sebrandon1: I will be adding a `GuaranteedPods` variable to the provider env to be able to loop through (or skip if empty) the GuranteedPods.  👍

sebrandon1: Completed in: https://github.com/test-network-function/cnf-certification-test/commit/053908124e1c2a9065bec8f165ed27c99669a13c

PR #445, opened 2022-08-23
Test names to be executed or skipped must start any of the test names provided

ramperher:
With the current tnf code, if we introduce a test name to be executed (`-f` argument) or skipped (`-s` argument), what is happening is that tnf is taking all the test names that contain that string provided.

So, for example, if we provide `operator` as argument, it would not only match with all operator unit tests (that is the idea), but it would also match with the `affiliated-certification-operator-is-certified` unit test, and it shouldn't.

With this change, we are ensuring that the name provided would match with unit tests only starting with the string provided, which is more restrictive and would avoid that problem.


sebrandon1 (approved):

run-cnf-suites.sh
```diff
@@ -41,12 +41,12 @@ while [[ $1 == -* ]]; do
 			fi ;;
 		-s|--skip)
 			while (( "$#" >= 2 )) && ! [[ $2 = --* ]] && ! [[ $2 = -* ]] ; do
-				SKIP="$2|$SKIP"
+				SKIP="(^$2)|$SKIP"
```
edcdavid, line 44: Is this working for you? I could not get it to work on my machine. Looks like other regex are working but not ^. Also the tests seem to skip all

fredericlepied:
test comment

run-cnf-suites.sh
sebrandon1: Yeah I'm seeing the skip-all happening as well.

```
Ran 0 of 61 Specs in 0.002 seconds
SUCCESS! -- 0 Passed | 0 Failed | 0 Pending | 61 Skipped
--- PASS: TestTest (22.43s)
PASS
```

I'm just trying to do `make build-cnf-tests && ./run-cnf-suites.sh -f affiliated-certification`

ramperher: Yes, me too, I manually tested the regex beforehand and it "should" work, but in my case, if you see the DCI job below, it's also skipping all the tests. I'll take a look

jmontesi:
Actually, I think it could be a good idea to completely remove the `-f` and `-s` options and use only labels if that's possible. Labels should be flexible enough to support all our use cases. Also, that approach is in line with what Ginkgo's documentation recommends. From the [Description-Based Filtering](https://onsi.github.io/ginkgo/#description-based-filtering) section of Ginkgo's doc:

_The description-based --focus and --skip flags were Ginkgo's original command-line based filtering mechanism and will continue to be supported - however we recommend using labels when possible as the label filter language is more flexible and easier to reason about._

Labels are easier to handle since this recent [PR](https://github.com/test-network-function/cnf-certification-test/pull/444).

ramperher:
> Actually, I think it could be a good idea to completely remove the `-f` and `-s` options and use only labels if that's possible. Labels should be flexible enough to support all our use cases. Also, that approach is in line with what Ginkgo's documentation recommends. From the [Description-Based Filtering](https://onsi.github.io/ginkgo/#description-based-filtering) section of Ginkgo's doc:
> 
> _The description-based --focus and --skip flags were Ginkgo's original command-line based filtering mechanism and will continue to be supported - however we recommend using labels when possible as the label filter language is more flexible and easier to reason about._
> 
> Labels are easier to handle since this recent [PR](https://github.com/test-network-function/cnf-certification-test/pull/444).

The easier way of defining the test cases or specific unit tests to be executed, the better, so if labels help on that purpose, it would be great, at least from my side as an user. However, I think labels are not currently supported with `run-tnf-container.sh`, aren't they? This option would have to be included then

I'll continue testing if it's possible to do the regex correctly for `-f` and `-s` options in the meanwhile.

ramperher:
> * SUCCESS https://www.distributed-ci.io/jobs/8acf3b3d-7755-4f39-9706-9c7314f6afd5/jobStates
>   from [Test names to be executed or skipped must start any of the test names provided #445](https://github.com/test-network-function/cnf-certification-test/pull/445)

I changed the code to use ^... instead of (^...) but the result is the same, all the tests are skipped. It's like ginkgo doesn't like when providing more complex regex, but in fact it's doing regex when defining the test suites as test1|test2|...|testn, so it's strange

Maybe it's better to try to only use labels, as @jmontesi commented before, what do you think? Because, in that case, I can give a try by reusing this PR for doing that.

jmontesi:
> > * SUCCESS https://www.distributed-ci.io/jobs/8acf3b3d-7755-4f39-9706-9c7314f6afd5/jobStates
> >   from [Test names to be executed or skipped must start any of the test names provided #445](https://github.com/test-network-function/cnf-certification-test/pull/445)
> 
> I changed the code to use ^... instead of (^...) but the result is the same, all the tests are skipped. It's like ginkgo doesn't like when providing more complex regex, but in fact it's doing regex when defining the test suites as test1|test2|...|testn, so it's strange
> 
> Maybe it's better to try to only use labels, as @jmontesi commented before, what do you think? Because, in that case, I can give a try by reusing this PR for doing that.

@ramperher, I'm woking on implementing the label-only approach. I still have to adapt the scrips that launch the CNF Certification as a container but it shouldn't take me too long.

ramperher:
> > > * SUCCESS https://www.distributed-ci.io/jobs/8acf3b3d-7755-4f39-9706-9c7314f6afd5/jobStates
> > >   from [Test names to be executed or skipped must start any of the test names provided #445](https://github.com/test-network-function/cnf-certification-test/pull/445)
> > 
> > 
> > I changed the code to use ^... instead of (^...) but the result is the same, all the tests are skipped. It's like ginkgo doesn't like when providing more complex regex, but in fact it's doing regex when defining the test suites as test1|test2|...|testn, so it's strange
> > Maybe it's better to try to only use labels, as @jmontesi commented before, what do you think? Because, in that case, I can give a try by reusing this PR for doing that.
> 
> @ramperher, I'm woking on implementing the label-only approach. I still have to adapt the scrips that launch the CNF Certification as a container but it shouldn't take me too long.

That sounds good, thank you! In the meanwhile, I will close this PR, I think it's better to wait for your approach, it will be cleaner and probably it will work better.

PR #460, opened 2022-08-26
Ignoring nftable and iptables rules set by machine-config operator

edcdavid:
build-depends: 25943

According to this bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1915027
The machine config operator configures iptables for every pods to prevent access to the machine config server. If these rules are seen still pass the test. 
I am not sure how well this is going to work in the future or for various versions of openshift since these rules could change
Also now running the ip6tables-save command. nft list ruleset covers both ip v4/v6 so no need for 2 calls.
https://github.com/test-network-function/cnf-certification-test-partner/pull/155 removes completely iptables and nftables to prevent any configuration

### DIFF
--- a/cnf-certification-test/networking/netutil/netutil.go
+++ b/cnf-certification-test/networking/netutil/netutil.go
@@ -124,9 +124,7 @@ const (
 )
 
 func stripSpaceTabLine(in string) string {
-	s1 := strings.ReplaceAll(in, "\n", "")
-	s2 := strings.ReplaceAll(s1, "\t", "")
-	return strings.ReplaceAll(s2, " ", "")
+	return strings.NewReplacer("\n", "", "\t", "", " ", "").Replace(in)
 }
 func isIPOrNSTablesPresent(cut *provider.Container, command string) (bool, string, error) { //nolint:gocritic
 	outStr, errStr, err := crclient.ExecCommandContainerNSEnter(command, cut)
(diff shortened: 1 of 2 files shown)


cnf-certification-test/networking/netutil/netutil.go
```diff
@@ -81,24 +81,78 @@ func GetListeningPorts(cut *provider.Container) (map[PortInfo]bool, error) {
...
+	-A OUTPUT -d 169.254.169.254/32 -p udp -m udp ! --dport 53 -j REJECT --reject-with icmp-port-unreachable
+	COMMIT
+	`
 )
 
-func isIPOrNSTablesPresent(cut *provider.Container, command string) (bool, error) {
+func stripSpaceTabLine(in string) string {
+	s1 := strings.ReplaceAll(in, "\n", "")
+	s2 := strings.ReplaceAll(s1, "\t", "")
+	return strings.ReplaceAll(s2, " ", "")
```
rdavid, line 129: replacer := strings.NewReplacer("\n", "", "\t", "", " ", "")
replacer.Replace(in)

ramperher:
It is curious, but in the first dcibot job, nftables test was passing, but in the second, it was iptables test the one that was passing. In theory, both must pass to consider this change is fine.

cnf-certification-test/networking/netutil/netutil.go
edcdavid: fixed

rdavid: There is no commit with the code change :-)

sebrandon1 (approved):

rdavid (approved):

PR #472, opened 2022-08-31
run-tnf-contaier.sh: remove the options -f (focus) and -s (skip)

jmontesi:
build-depends: 26051
build-depends: https://github.com/dci-labs/dallas-pipelines/pull/474

This options' functionality overlaps with that of labels, which will be
left as the only method for filtering the test suites/cases.

Some examples:
-l "operator"
-l 'common && !lifecycle'
-l 'observability,access-control'


cnf-certification-test/identifiers/identifiers.go
```diff
@@ -473,15 +462,20 @@ func getGinkgoTestID(identifier claim.Identifier, extra string) string {
...
-		return nil
+		panic("Invalid Identifier Url")
 	}
-	return strings.Split(result[1], "/")
+
+	result = strings.Split(result[1], "/")
+	suiteName = result[0]
+	testName = result[1]
+
+	return suiteName, testName
```
sebrandon1, line 478: Was this just for readability?

sebrandon1 (commented):
An initial comment.  That `while` loop in run-tnf-container.sh looks much better now 😄

cnf-certification-test/identifiers/identifiers.go
jmontesi: I believe the linter was complaining about some requirement requiring the return values to be named. I can recheck this.

ramperher:
I believe you need to update the README.md file to remove the references to focus and skip arguments.

By the way, I've appended to the description of the PR the following lines, to test this functionality in our labs for each update made in this change:

build-depends: 26007
build-depends: https://github.com/dci-labs/dallas-pipelines/pull/465

So, new DCI jobs launched will be based on that code, which makes use of `-l` instead of `-f` and `-s`

run-tnf-container.sh
```diff
@@ -49,8 +48,7 @@ usage() {
 	  -n: set the network mode of the container.
 	  -d: set the DNS resolver address for the test containers started by docker, may be required with 
 	      certain docker version if the kubeconfig contains host names
-    -f: Set the suites that should be tested, multiple suites can be supplied
-    -s: Set the test cases that should be skipped
+    -f: Set the test labels that should be tested
```
ramperher, line 51: this must be -l instead of -f, and properly indented

jmontesi:
> I believe you need to update the README.md file to remove the references to focus and skip arguments.
> 
> By the way, I've appended to the description of the PR the following lines, to test this functionality in our labs for each update made in this change:
> 
> build-depends: 26007 build-depends: https://github.com/dci-labs/dallas-pipelines/pull/465
> 
> So, new DCI jobs launched will be based on that code, which makes use of `-l` instead of `-f` and `-s`. I'll also update these changes to keep sync with this change.

Yes, I still need to update the README.md and the Git workflows. So, from the DCI side is everything ok to proceed with this change?

ramperher:
> > I believe you need to update the README.md file to remove the references to focus and skip arguments.
> > By the way, I've appended to the description of the PR the following lines, to test this functionality in our labs for each update made in this change:
> > build-depends: 26007 build-depends: [dci-labs/dallas-pipelines#465](https://github.com/dci-labs/dallas-pipelines/pull/465)
> > So, new DCI jobs launched will be based on that code, which makes use of `-l` instead of `-f` and `-s`. I'll also update these changes to keep sync with this change.
> 
> Yes, I still need to update the README.md and the Git workflows. So, from the DCI side is everything ok to proceed with this change?

I'm still testing it, as you can see, I need to check some things right now before merging. I'll tell you in the next minutes :)

ramperher:
> * SUCCESS https://www.distributed-ci.io/jobs/dc33ced0-95bd-44cf-905d-b119875a940d/jobStates
>   from [run-tnf-contaier.sh: remove the options -f (focus) and -s (skip) #472](https://github.com/test-network-function/cnf-certification-test/pull/472)

You have my :+1: from my side in terms of execution, but also waiting for the README + the minor comment I left before.

I've used these labels: `-l access-control,(networking && !networking-dual-stack-service),lifecycle,observability,(platform-alteration && !platform-alteration-service-mesh-usage),operator` (so, for example, we test all the networking tests excepting the dual-stack one, or all the platform-alteration tests excepting the service-mesh one. Now, for example, if I tip `operator`, the affiliated-certification test regarding operator tests is not executed, which was the behavior expected :)

ramperher:
@sebrandon1, @rdavid and I were discussing about this change yesterday, and maybe, to guarantee backwards compatibility, it would be better to include `-l` option in `run-tnf-container.sh` but without deleting `-f` and `-s` options, so that they are compatible. What do you think?

jmontesi:
> @sebrandon1, @rdavid and I were discussing about this change yesterday, and maybe, to guarantee backwards compatibility, it would be better to include `-l` option in `run-tnf-container.sh` but without deleting `-f` and `-s` options, so that they are compatible. What do you think?

In my opinion there must be a very strong reason to prevent an improvement to our code base/way of working just because we want to guarantee backwards compatibility. Sometimes it's necessary due to the nature of the change but in general it's an approach which often results in less flexible and suboptimal software to say the least.

If, in any case, you prefer to keep those options then I think it's better to close this PR and open a new one just to add the `-l` option to the `run-tnf-suites.sh` script.

ramperher:
> > @sebrandon1, @rdavid and I were discussing about this change yesterday, and maybe, to guarantee backwards compatibility, it would be better to include `-l` option in `run-tnf-container.sh` but without deleting `-f` and `-s` options, so that they are compatible. What do you think?
> 
> In my opinion there must be a very strong reason to prevent an improvement to our code base/way of working just because we want to guarantee backwards compatibility. Sometimes it's necessary due to the nature of the change but in general it's an approach which often results in less flexible and suboptimal software to say the least.
> 
> If, in any case, you prefer to keep those options then I think it's better to close this PR and open a new one just to add the `-l` option to the `run-tnf-suites.sh` script.

We will have a meeting today to discuss about this, I can forward you the meeting invitation if you want (it is at 17:00h CEST). I agree with you with your first statement because it is the philosophy we have for our DCI agents: we always force to use the latest version possible and to do the updates required in clients' side to make it work. However, the problem here is not related to the software itself, but it's related to compatibility with scripts used with partners, mainly. So, as I said, we'll discuss about this later on and then come up with a decision.

ramperher:
Sorry for the noise, I was testing the changes proposed in the meeting we had before, I've just finished with that last successful job

jmontesi:
Closed in favor of: https://github.com/test-network-function/cnf-certification-test/pull/496

PR #476, opened 2022-09-01
Add fix for the sysctlCommand in sysctl-config tc

shaior:
Added a fix to the GetSysctlSettings func, the sysctlCommand should run with a chroot /host prefix.


sebrandon1 (approved):

sebrandon1:
The DCI failure seems to be coming from other tests.  Will merge.

PR #478, opened 2022-09-01
Add test for LabelsMatch

sebrandon1:
Forgot to add these tests in #423

### DIFF
--- a/cnf-certification-test/networking/policies/policies_test.go
+++ b/cnf-certification-test/networking/policies/policies_test.go
@@ -180,3 +180,38 @@ func TestIsNetworkPolicyCompliant(t *testing.T) {
 		assert.Equal(t, tc.expectedIngressOutput, IsNetworkPolicyCompliant(&tc.testNP, networkingv1.PolicyTypeIngress))
 	}
 }
+
+func TestLabelsMatch(t *testing.T) {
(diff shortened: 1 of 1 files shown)


cnf-certification-test/networking/policies/policies_test.go
```diff
@@ -180,3 +180,38 @@ func TestIsNetworkPolicyCompliant(t *testing.T) {
...
+			expectedOutput: true,
+		},
+		{
+			testPodSelectorLabels: metav1.LabelSelector{
+				MatchLabels: map[string]string{
+					"label1": "value1",
+				},
+			},
+			testPodLabels: map[string]string{
+				"label2": "value2",
```
greyerof, line 208: What about another negative TC with the same label key but different value?
```
			testPodSelectorLabels: metav1.LabelSelector{
				MatchLabels: map[string]string{
					"label1": "value1",
				},
			},
			testPodLabels: map[string]string{
				"label1": "value2",
			},
```

greyerof (approved):

edcdavid (approved):

cnf-certification-test/networking/policies/policies_test.go
sebrandon1: Done!

PR #485, opened 2022-09-05
Fetch tool improvement to minimize changes in operator pages.

greyerof:
This improvement tries to minimize the number of files changed for the offline catalog whenever new operators have been added as certified in the online catalog. The per-page information fields "total", "page_size", and "page" have been removed to avoid this change to appear in the automatic PR for all the pages:
image
Those fields can be removed as they're never used by the test case.

The "total" field will be shown only in the last page, just for debugging/reference.

### DIFF
--- a/cmd/tnf/fetch/fetch.go
+++ b/cmd/tnf/fetch/fetch.go
@@ -155,6 +155,7 @@ func serializeData(data CertifiedCatalog) {
 	}
 	log.Info("serialization time", time.Since(start))
 }
+
 func getOperatorCatalogSize() (size, pagesize uint, err error) {
 	log.Infof("Getting operators catalog size, url: %s", operatorsCatalogSizeURL)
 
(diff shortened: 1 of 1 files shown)


sebrandon1 (approved):

cmd/tnf/fetch/fetch.go
```diff
@@ -173,12 +174,25 @@ func getOperatorCatalogSize() (size, pagesize uint, err error) {
 	return aCatalog.Total, aCatalog.PageSize, nil
 }
 
-func getOperatorCatalogPage(page, size uint) error {
+func getOperatorCatalogPage(page, size uint, isLastPage bool) error {
+	const (
```
rdavid, line 178: Consider:
```
	const (
		excludeFilterForLastPage = "&exclude=page,page_size"
		excludeFilter            = excludeFilterForLastPage+",total"
		
	)
```

rdavid (approved):

PR #506, opened 2022-09-13
Switch Pod and Container to nest the k8s structs

sebrandon1:
Similar to #501 that did the same for `Deployment` and `StatefulSet`.  This removes the "Data" variable from the struct that wraps the k8s structs themselves.


sebrandon1:
Looks like the DCI bot is failing with `No test focus (-f) or label (-l) was set, so only diagnostic functions will run..` @ramperher @jmontesi

ramperher:
> Looks like the DCI bot is failing with `No test focus (-f) or label (-l) was set, so only diagnostic functions will run..` @ramperher @jmontesi

It seemed to be a transient failure, I've rechecked and it's fine now

PR #512, opened 2022-09-15
Add control-plane toleration to daemonset.

greyerof:
Nodes with k8s version v1.24 in kind have a new toleration for the control-plane+worker nodes. We need to add it here, otherwise the pod won't run on that node, making CI to fail.

Signed-off-by: Gonzalo Reyero Ferreras <greyerof@redhat.com>


sebrandon1 (approved):
Nice catch!  It should have probably got carried over from https://github.com/test-network-function/cnf-certification-test-partner/pull/131 originally.

jmontesi (approved):

PR #513, opened 2022-09-15
manageability: new test to check for container ports naming conventions

jmontesi:


cnf-certification-test/identifiers/identifiers.go
```diff
@@ -525,6 +525,11 @@ var (
 		Url:     formTestURL(common.LifecycleTestKey, "cpu-isolation"),
 		Version: versionOne,
 	}
+	TestContainerPortNameFormat = claim.Identifier{
```
edcdavid, line 528: you could also use the AddCatalogEntry to add a new TC

cnf-certification-test/manageability/suite.go
```diff
@@ -57,3 +65,29 @@ func testContainersImageTag(env *provider.TestEnvironment) {
 		ginkgo.Fail("Containers have been found that are missing tag(s) on image.")
 	}
 }
+
+// The name field in the ContainerPort section must be of the form <protocol>[-<suffix>] where <protocol> is one of the following,
+// and the optional <suffix> can be chosen by the application. Allowed protocol names: grpc, grpc-web, http, http2, tcp, udp.
+var allowedProtocolNames = map[string]bool{"grpc": true, "http": true, "http2": true, "tcp": true, "udp": true}
```
edcdavid, line 71: I don't believe this is an exhaustive list, maybe we could scrape the following db for the ip protocol keyword:
https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xhtml

jmontesi: The problem I see is that the requirement is too vague. In the list you propose I haven't found "grpc" or "http2", which are given as valid examples of protocol names. I think this requirement needs some clarification.

edcdavid: yes, I agree. fyi, I found this project that already has a json with the port mappings: https://github.com/silverwind/port-numbers/blob/master/ports.json

cnf-certification-test/manageability/suite.go
```diff
@@ -57,3 +65,29 @@ func testContainersImageTag(env *provider.TestEnvironment) {
...
+
+func containerPortNameFormatCheck(portName string) bool {
```
sebrandon1, line 73: Just for organization, should we move this to something like `manageability/ports.go` or something similar?  To keep the suite.go clear of helper funcs?

cnf-certification-test/identifiers/identifiers.go
greyerof: This can be done in a different PR by someone else.

cnf-certification-test/manageability/suite.go
greyerof: This can be done in a different PR by someone else.

cnf-certification-test/manageability/suite.go
greyerof: As discussed internally, let's keep the current implementation for now.

greyerof (approved):
I've rebased it because Juan is not available. I propose to merge this as it is right now, since this implementation seems to be valid as per current requirements. Improvements by @edcdavid and @sebrandon1 can be done in separate PRs so that no more extra commits are added to this (his) PR by a different author.

edcdavid (approved):

sebrandon1 (approved):

PR #514, opened 2022-09-15
Fix to deploy the daemonset also in diagnostic mode.

greyerof:
Diagnostic mode also needs the daemonset pods so we can get all the node information through them.

Also, TNF will stop execution if the daemonset hasn't been deployed correctly, as a lot of TCs relay on it to work.

### DIFF
--- a/cnf-certification-test/suite_test.go
+++ b/cnf-certification-test/suite_test.go
@@ -142,6 +142,7 @@ func TestTest(t *testing.T) {
 	// Diagnostic functions will run when no focus test suites or labels are provided.
 	var diagnosticMode bool
 	if len(ginkgoConfig.FocusStrings) == 0 && ginkgoConfig.LabelFilter == "" {
+		log.Infof("TNF will run in diagnostic mode so no test case will be launched.")
 		diagnosticMode = true
 	}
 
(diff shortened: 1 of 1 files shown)


cnf-certification-test/suite_test.go
```diff
@@ -142,18 +142,19 @@ func TestTest(t *testing.T) {
...
 
 	// Deploy the daemonset before getting the environment for the first time
-	if !diagnosticMode {
-		err := daemonset.DeployPartnerTestDaemonset()
-		if err != nil {
-			log.Errorf("Error deploying partner daemonset %s", err)
-		}
+	err := daemonset.DeployPartnerTestDaemonset()
+	if err != nil {
+		log.Errorf("Error deploying partner daemonset %s", err)
```
rdavid, line 155: Add vertical space before the comment.

rdavid (approved):

cnf-certification-test/suite_test.go
greyerof: Done, but, please keep in mind we're not enforcing any strict coding/cosmetics guidelines on this project. 

Personally, for the sake of code readability, and as a general rule, also like to add those empty lines, but I don't see how an extra empty line can help readability in this particular code block, (`if {}`), that only takes 3 lines, where one of them is a comment. 

Also, notice that those extra empty-lines do increase the line count, which would potentially result in a linter error (funlen).

rdavid: I understand, personally, I hate vertical spaces. The only vertical space I put is before any comment. If you want separate some code, it is better to write a function.

PR #516, opened 2022-09-15
Update copyright year

rdavid:


sebrandon1:
For these, in past experience the copyright only applies to whenever the file was last touched/modified.  So not _every_ file will be 2022 but I guess there's not really a way of knowing since we just created this repo this year and copied everything over.

sebrandon1 (approved):

rdavid:
@sebrandon1 There are discussions on this topic. I believe that the second year should be a year of publishing, not a year of changing. So mostly people in open source update all files to the last year.

greyerof (approved):

PR #523, opened 2022-09-16
Add AffinityRequired test to lifecycle

sebrandon1:
build-depends: 26152

Adds a test to the `lifecycle` suite that checks for a label `AffinityRequired: true` to be supplied for workloads that require colocation of pods. 

- Tests pods, deployments, and statefulsets.
- Adds "groupings" section(s) into the provider.go.
- Adds `AffinityRequired()` and `IsAffinityCompliant()` funcs to each of the above structs.
- Had to modify `testPodNodeSelectorAndAffinityBestPractices` to avoid pods that have `AffinityRequired` flags.  This func was already looping through NonGuaranteedPods.  Maybe there is a way to do cross-sections of different lists of pods we could implement later.
- The `IsAffinityCompliant` tests are all relatively the same with some copy/pasting.

### DIFF
--- a/CATALOG.md
+++ b/CATALOG.md
@@ -338,6 +338,45 @@ Exception Process|There is no documented exception process for this.
 
 ### lifecycle
 
+#### affinity-required-deployments
+
+Property|Description
+---|---
+Test Case Name|affinity-required-deployments
+Test Case Label|lifecycle-affinity-required-deployments
+Unique ID|http://test-network-function.com/testcases/lifecycle/affinity-required-deployments
+Version|v1.0.0
+Description|http://test-network-function.com/testcases/lifecycle/affinity-required-deployments Checks that affinity rules are in place if AffinityRequired: 'true' labels are set on Deployments.
+Result Type|informative
+Suggested Remediation|If a pod/statefulset/deployment is required to use affinity rules, please add AffinityRequired: 'true' as a label
+Best Practice Reference|https://TODO Section 4.6.24
+Exception Process|There is no documented exception process for this.
+#### affinity-required-pods
+
+Property|Description
+---|---
+Test Case Name|affinity-required-pods
+Test Case Label|lifecycle-affinity-required-pods
+Unique ID|http://test-network-function.com/testcases/lifecycle/affinity-required-pods
+Version|v1.0.0
+Description|http://test-network-function.com/testcases/lifecycle/affinity-required-pods Checks that affinity rules are in place if AffinityRequired: 'true' labels are set on Pods.
+Result Type|informative
+Suggested Remediation|If a pod/statefulset/deployment is required to use affinity rules, please add AffinityRequired: 'true' as a label
+Best Practice Reference|https://TODO Section 4.6.24
+Exception Process|There is no documented exception process for this.
+#### affinity-required-statefulsets
+
+Property|Description
+---|---
+Test Case Name|affinity-required-statefulsets
+Test Case Label|lifecycle-affinity-required-statefulsets
+Unique ID|http://test-network-function.com/testcases/lifecycle/affinity-required-statefulsets
+Version|v1.0.0
+Description|http://test-network-function.com/testcases/lifecycle/affinity-required-statefulsets Checks that affinity rules are in place if AffinityRequired: 'true' labels are set on StatefulSets.
+Result Type|informative
+Suggested Remediation|If a pod/statefulset/deployment is required to use affinity rules, please add AffinityRequired: 'true' as a label
+Best Practice Reference|https://TODO Section 4.6.24
+Exception Process|There is no documented exception process for this.
 #### container-shutdown
 
 Property|Description
--- a/README.md
+++ b/README.md
@@ -75,7 +75,8 @@ For Network Interfaces:
 The label test-network-function.com/skip_connectivity_tests excludes pods from all connectivity tests. The label value is not important, only its presence.
 The label test-network-function.com/skip_multus_connectivity_tests excludes pods from multus connectivity tests. Tests on default interface are still done. The label value is not important, only its presence.
 
-
+### AffinityRequired
+For CNF workloads that require pods to use Pod or Node Affinity rules, the label `AffinityRequired: true` must be included on either the Deployment, StatefulSet, or Pod YAML.  This will prevent any tests for anti-affinity to fail as well as test your workloads for affinity rules that support your CNF's use-case.
 
 ### certifiedcontainerinfo
 
--- a/cnf-certification-test/identifiers/identifiers.go
+++ b/cnf-certification-test/identifiers/identifiers.go
@@ -99,6 +99,9 @@ var (
 	TestContainerIsCertifiedDigestIdentifier claim.Identifier
 	TestPodHugePages2M                       claim.Identifier
 	TestReservedExtendedPartnerPorts         claim.Identifier
+	TestAffinityRequiredPods                 claim.Identifier
+	TestAffinityRequiredDeployments          claim.Identifier
+	TestAffinityRequiredStatefulSets         claim.Identifier
 )
 
 //nolint:funlen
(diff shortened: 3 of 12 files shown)


cnf-certification-test/identifiers/identifiers.go
```diff
@@ -96,6 +96,7 @@ var (
 	TestICMPv4ConnectivityIdentifier   claim.Identifier
 	TestNetworkPolicyDenyAllIdentifier claim.Identifier
 	Test1337UIDIdentifier              claim.Identifier
+	TestColocationAllowed              claim.Identifier
```
edcdavid, line 99: Change across the code ColocationAllowed with ColocationRequired

pkg/provider/provider.go
```diff
@@ -43,7 +43,7 @@ import (
 )
 
 const (
-	AffinityAllowedKey               = "AffinityAllowed"
+	AffinityRequiredKey              = "AffinityRequired"
```
edcdavid, line 46: Add a note in the Readme about the new "AffinityRequired" label

ramperher:
Hi @sebrandon1 , just to let you know, I've added this in the PR description: `build-depends: 26152`, in order to test this change with https://softwarefactory-project.io/r/c/dci-openshift-app-agent/+/26152, so that I can validate that, by including hugepages in our workloads, the test passes

CATALOG.md
```diff
@@ -325,6 +325,19 @@ Exception Process|There is no documented exception process for this.
 
 ### lifecycle
 
+#### colocation-allowed
+
+Property|Description
+---|---
+Test Case Name|colocation-allowed
+Test Case Label|lifecycle-colocation-allowed
```
ramperher, line 333: test name is now `affinity-required`, I believe

ramperher:
Hi again @sebrandon1 , I've been doing some tests and I have a main question here: isn't this new test colliding with `lifecycle-pod-high-availability` workflow?

Because, in theory, in order to pass `lifecycle-pod-high-availability` test suite, apart from having more than 1 replica in the deployment/statefulset, it is also needed to have defined podAntiAffinity rules. However, in this new test suite, even though you set the AffinityRequired label and you have podAffinity rules defined, if a podAntiAffinity rule is found, then it doesn't pass the test.

Does it mean that both `lifecycle-pod-high-availability` and `lifecycle-affinity-required` cannot pass simultaneously?

sebrandon1:
@ramperher Thanks for the comment.  I forgot to change that test to look at the deployments/statefulsets that do not have the AffinityRequired flag only.  See this commit: https://github.com/test-network-function/cnf-certification-test/pull/523/commits/6f54cd25e170555650c5a6a5ac168272ee1e3ecc

ramperher:
> @ramperher Thanks for the comment. I forgot to change that test to look at the deployments/statefulsets that do not have the AffinityRequired flag only. See this commit: [6f54cd2](https://github.com/test-network-function/cnf-certification-test/commit/6f54cd25e170555650c5a6a5ac168272ee1e3ecc)

Cool! Thanks for the update! I'll take a look tomorrow to test both cases then! Don't pay attention if there's any failure in today's jobs, they may fail because of the new affiliated-certification test, by the way

CATALOG.md
sebrandon1: Thanks, just regenerated the CATALOG.md.

pkg/provider/provider.go
sebrandon1: Done.

cnf-certification-test/identifiers/remediation.go
```diff
@@ -195,4 +195,6 @@ const (
 	CPUIsolationRemediation = `CPU isolation testing is enabled.  Please ensure that all pods adhere to the CPU isolation requirements`
 
 	UID1337Remediation = `Use another process UID that is not 1337`
+
+	AffinityRequiredRemediation = `If a pod/statefulset/deployment is required to use affinity rules, please add 'AffinityRequired: true' as a label`
```
ramperher, line 199: Same: I'd emphasise that label value must be string, not boolean, so I'd say `AffinityRequired: 'true'`, maybe.

cnf-certification-test/identifiers/identifiers.go
```diff
@@ -161,6 +162,17 @@ The label value is not important, only its presence.`,
 		bestPracticeDocV1dot4URL+" Section 3.5.4",
 		TagExtended)
 
+	TestAffinityRequired = AddCatalogEntry(
+		"affinity-required",
+		common.LifecycleTestKey,
+		`Checks that affinity rules are in place if AffinityRequired: true labels are set on Pods, Deployments, and StatefulSets.`,
```
ramperher, line 168: Can we remark here what I said about "true" is string but not a boolean?

cnf-certification-test/lifecycle/suite.go
```diff
@@ -148,6 +148,13 @@ var _ = ginkgo.Describe(common.LifecycleTestKey, func() {
 		testhelper.SkipIfEmptyAny(ginkgo.Skip, env.GuaranteedPods)
 		testCPUIsolation(&env)
 	})
+
+	testID, tags = identifiers.GetGinkgoTestIDAndLabels(identifiers.TestAffinityRequired)
+	ginkgo.It(testID, ginkgo.Label(tags...), func() {
```
ramperher, line 153: I think this workflow is not working as expected, I'll show you an example in the comment including these minor comments.

cnf-certification-test/lifecycle/suite.go
```diff
@@ -456,3 +464,51 @@ func testCPUIsolation(env *provider.TestEnvironment) {
...
+
+	testhelper.AddTestResultLog("Non-compliant", podsDesiringAffinityRequiredMissingLabel, tnf.ClaimFilePrintf, ginkgo.Fail)
+}
+
+func testAffinityRequiredDeployments(env *provider.TestEnvironment) {
+	testhelper.SkipIfEmptyAny(ginkgo.Skip, env.AffinityRequiredDeployments)
+
+	var deploymentsDesiringAffinityRequiredMissingLabel []*provider.Deployment
+	for _, dep := range env.AffinityRequiredDeployments {
+		// Check if the statefulset is Affinity compliant.
```
ramperher, line 489: typo: deployment

CATALOG.md
```diff
@@ -338,6 +338,19 @@ Exception Process|There is no documented exception process for this.
...
+
+Property|Description
+---|---
+Test Case Name|affinity-required
+Test Case Label|lifecycle-affinity-required
+Unique ID|http://test-network-function.com/testcases/lifecycle/affinity-required
+Version|v1.0.0
+Description|http://test-network-function.com/testcases/lifecycle/affinity-required Checks that affinity rules are in place if AffinityRequired: true labels are set on Pods, Deployments, and StatefulSets.
+Result Type|informative
+Suggested Remediation|If a pod/statefulset/deployment is required to use affinity rules, please add 'AffinityRequired: true' as a label
```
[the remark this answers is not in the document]
sebrandon1: Done.

README.md
```diff
@@ -75,7 +75,8 @@ For Network Interfaces:
 The label test-network-function.com/skip_connectivity_tests excludes pods from all connectivity tests. The label value is not important, only its presence.
 The label test-network-function.com/skip_multus_connectivity_tests excludes pods from multus connectivity tests. Tests on default interface are still done. The label value is not important, only its presence.
 
-
+### AffinityRequired
+For CNF workloads that require pods to use Pod or Node Affinity rules, the label `AffinityRequired: true` must be included on either the Deployment, StatefulSet, or Pod YAML.  This will prevent any tests for anti-affinity to fail as well as test your workloads for affinity rules that support your CNF's use-case.
```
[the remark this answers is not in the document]
sebrandon1: Done.

cnf-certification-test/identifiers/remediation.go
sebrandon1: Done.

cnf-certification-test/identifiers/identifiers.go
sebrandon1: Done.

cnf-certification-test/lifecycle/suite.go
sebrandon1: Ah yes, it's because of the multiple `It` Ginkgo blocks.  I'll fix.

cnf-certification-test/lifecycle/suite.go
sebrandon1: Done.

ramperher (approved):
Everything working fine right now, thanks!

pkg/provider/deployments.go
```diff
@@ -65,6 +66,30 @@ func (d *Deployment) ToString() string {
...
+			return false
+		}
+		return result
+	}
+	return false
+}
+
+func (d *Deployment) IsAffinityCompliant() error {
+	var errMsg error
+	//nolint:gocritic
```
greyerof, line 82: Gocritic can be avoided easily changing this if/else if/else if for a swich, or just three ifs with a fast return:
```
func (d *Deployment) IsAffinityCompliant() error {
	if d.Spec.Template.Spec.Affinity == nil {
		return fmt.Errorf("%s has been found with an AffinityRequired flag but is missing corresponding affinity rules", d.String())
	}

	if d.Spec.Template.Spec.Affinity.PodAntiAffinity != nil {
		return fmt.Errorf("%s has been found with an AffinityRequired flag but has anti-affinity rules", d.String())
	}

	if d.Spec.Template.Spec.Affinity.NodeAffinity == nil {
		return fmt.Errorf("%s has been found with an AffinityRequired flag but is missing corresponding pod/node affinity rules", d.String())
	}

	return nil
}
```
Nit 1: The existence of this method in the Deployment struct implicitly says that all deployments should have an Affinity field that satisfies this checks. But that's only the case for deployments with the "AffinityRequired" label. Why not checking that label here and do the appropriate affinity checks based on it? Same for stateulfSets and pods.
 
Nit 2: With  IsX() functions I usually expect a `bool` as a return param. In this case, I would have done this:
`IsAffinityCompliant() (isCompliant bool, nonCompliantReason string)`

greyerof (approved):
Some (nit) remarks that can be done in a separate PR if agreed.

pkg/provider/deployments.go
```diff
@@ -65,6 +66,30 @@ func (d *Deployment) ToString() string {
 	)
 }
 
+func (d *Deployment) AffinityRequired() bool {
+	if val, ok := d.Labels[AffinityRequiredKey]; ok {
+		result, err := strconv.ParseBool(val)
+		if err != nil {
+			return false
```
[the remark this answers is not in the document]
sebrandon1: Done.

pkg/provider/deployments.go
sebrandon1: Done.

PR #529, opened 2022-09-20
Fix the reclaim policy test to look through PVCs

sebrandon1:
Related to #390 

@greyerof pointed out that the prior implementation was trying to match arbitrary pod.Volume names to PersistentVolume names which may/may not be the same name.  The pod volume is tied to a PVC, which is tied to a PV.  We just needed to take the extra steps to take a deeper look into the PVCs and corresponding PVs.


sebrandon1:
@ramperher can you check into why the above DCI run failed?  Not sure but it doesn't look like it attempted to run the suite?

ramperher:
> @ramperher can you check into why the above DCI run failed? Not sure but it doesn't look like it attempted to run the suite?

It failed in a previous step, while deploying the resources under test, I'll recheck.

cnf-certification-test/lifecycle/volumes/volumes.go
```diff
@@ -0,0 +1,48 @@
...
+		if pvcs[i].Name == pvcName {
+			return &pvcs[i]
+		}
+	}
+	return nil
+}
+
+func IsPodVolumeReclaimPolicyDelete(vol *corev1.Volume, pvs []corev1.PersistentVolume, pvcs []corev1.PersistentVolumeClaim) bool {
+	// Safeguard against nil
+	if vol.PersistentVolumeClaim != nil {
```
greyerof, line 34: If this check fails, this function returns false, which implies that all volumes in a pod should be persistent volumes... but I don't think all pods in a CNF must to use persistent storage. I believe we should check pv's reclaim policy only for pod's volumes that were configured to use a pvc.

pkg/autodiscover/autodiscover_pv.go
```diff
@@ -31,3 +31,11 @@ func getPersistentVolumes(oc corev1client.CoreV1Interface) ([]corev1.PersistentV
 	}
 	return pvs.Items, nil
 }
+
+func getPersistentVolumeClaims(oc corev1client.CoreV1Interface) ([]corev1.PersistentVolumeClaim, error) {
+	pvcs, err := oc.PersistentVolumeClaims("").List(context.TODO(), metav1.ListOptions{})
```
greyerof, line 36: Suggestion: As pvcs are namespaced objects, shouldn't we List() the ones in the configured CNF's namespaces?

ramperher:
> * SUCCESS https://www.distributed-ci.io/jobs/7084e7ee-9e6a-46ac-baac-e198cde8b375/jobStates
>   from [Fix the reclaim policy test to look through PVCs #529](https://github.com/test-network-function/cnf-certification-test/pull/529)

The test is fine from my side, the previous one might have failed because of a transient issue

pkg/autodiscover/autodiscover_pv.go
sebrandon1: I kept this namespace field empty because I figured this was our "general" PVC collecting function and we would filter later like we do with the other resources.

cnf-certification-test/lifecycle/volumes/volumes.go
sebrandon1: Good call, I'll adjust to account for that.

sebrandon1: In the latest [commit](https://github.com/test-network-function/cnf-certification-test/pull/529/commits/74c95f21a0c24db8bb38124d6cb4d3fbf88c1278) I added a check to skip volumes that don't contain PVCs.

greyerof (approved):

cnf-certification-test/lifecycle/volumes/volumes.go
```diff
@@ -0,0 +1,45 @@
...
+	for i := range pvcs {
+		if pvcs[i].Name == pvcName {
+			return &pvcs[i]
+		}
+	}
+	return nil
+}
+
+func IsPodVolumeReclaimPolicyDelete(vol *corev1.Volume, pvs []corev1.PersistentVolume, pvcs []corev1.PersistentVolumeClaim) bool {
+	// Check if the Volume has a PVC that actually exists.
```
edcdavid, line 33: // Check if the Volume has a PVC that actually exists. -> // Check if the Volume is bound to a PVC.

cnf-certification-test/lifecycle/suite.go
```diff
@@ -424,18 +425,24 @@ func testPodPersistentVolumeReclaimPolicy(env *provider.TestEnvironment) {
...
+			if !volumes.IsPodVolumeReclaimPolicyDelete(&put.Spec.Volumes[pvIndex], env.PersistentVolumes, env.PersistentVolumeClaims) {
+				persistentVolumesBadReclaim = append(persistentVolumesBadReclaim, put.String())
+				tnf.ClaimFilePrintf("%s contains volume: %s has been found without a reclaim policy of DELETE.", put.String(), &put.Spec.Volumes[pvIndex].Name)
+				break
 			}
 		}
 	}
 
 	if n := len(persistentVolumesBadReclaim); n > 0 {
-		errMsg := fmt.Sprintf("Persistent Volumes found that are missing a reclaim policy of DELETE: %d. See logs for more detail.", n)
```
edcdavid, line 438: Optional: use AddTestResultLog

edcdavid (approved):

cnf-certification-test/lifecycle/volumes/volumes.go
sebrandon1: Fixed.

cnf-certification-test/lifecycle/suite.go
sebrandon1: Fixed.

PR #530, opened 2022-09-20
Add tags to differentiate tests by document

sebrandon1:
Just a proof of concept that shows how we can have some different flexibility options when it comes to versioning based on the requirements documents that tests get introduced in.


sebrandon1:
Closing this PoC.

PR #553, opened 2022-09-27
Turn groupings into env funcs

sebrandon1:
I created `pkg/provider/filters.go` to house all of these environment funcs to filter Pods, Deployments, StatefulSets on the fly versus once during the autodiscover.  I can see where this might be problematic as far as running the filtering multiple times however I think it will help clean up the code in the long run.

For example, see `lifecycle/suite.go`.  Instead of using our slice of NonGuaranteedPods and then filtering, I'm combining all of the logic into `GetNonGuaranteedPodsWithAntiAffinity()`.  This can help get all of the _filtering_ out of the way prior to running the test.

image

### DIFF
--- a/cnf-certification-test/lifecycle/suite.go
+++ b/cnf-certification-test/lifecycle/suite.go
@@ -93,7 +93,7 @@ var _ = ginkgo.Describe(common.LifecycleTestKey, func() {
 		if env.GetWorkerCount() < minWorkerNodesForLifecycle {
 			ginkgo.Skip("Skipping pod high availability test because invalid number of available workers.")
 		}
-		testhelper.SkipIfEmptyAll(ginkgo.Skip, env.AntiAffinityRequiredDeployments, env.AntiAffinityRequiredStatefulSets)
+		testhelper.SkipIfEmptyAll(ginkgo.Skip, env.GetAntiAffinityRequiredDeployments(), env.GetAntiAffinityRequiredStatefulSets())
 		testHighAvailability(&env)
 	})
 
@@ -102,7 +102,7 @@ var _ = ginkgo.Describe(common.LifecycleTestKey, func() {
 		if env.GetWorkerCount() < minWorkerNodesForLifecycle {
 			ginkgo.Skip("Skipping pod scheduling test because invalid number of available workers.")
 		}
-		testhelper.SkipIfEmptyAny(ginkgo.Skip, env.NonGuaranteedPods)
+		testhelper.SkipIfEmptyAny(ginkgo.Skip, env.GetNonGuaranteedPodsWithAntiAffinity())
 		testPodNodeSelectorAndAffinityBestPractices(&env)
 	})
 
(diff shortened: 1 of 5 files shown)


cnf-certification-test/lifecycle/suite.go
```diff
@@ -102,13 +102,13 @@ var _ = ginkgo.Describe(common.LifecycleTestKey, func() {
 		if env.GetWorkerCount() < minWorkerNodesForLifecycle {
 			ginkgo.Skip("Skipping pod scheduling test because invalid number of available workers.")
 		}
-		testhelper.SkipIfEmptyAny(ginkgo.Skip, env.NonGuaranteedPods)
+		testhelper.SkipIfEmptyAny(ginkgo.Skip, env.GetNonGuaranteedPodsWithAntiAffinity())
```
ramperher, line 105: Hi Brandon, just comparing the job executed for [this change](https://www.distributed-ci.io/jobs/1baa8941-bab7-4ad7-b0d6-d76a49a73caf/tests) and another from current [main branch](https://www.distributed-ci.io/jobs/0e57e2ec-fa49-4c26-b68f-90d6e4643ee4/tests), I saw `lifecycle-pod-scheduling` test is now being skipped. Is it because of this change, that now considers that the test is only executed for non guaranteed pods (in terms of QoS), but also having anti affinity rules?

sebrandon1: So I think the `main` branch run is only passing because there are no pods to test against.  The change here tells the code to skip if there are no pods that meet those criteria.  I think the main branch should probably "skip" as well but we only skip based on these TestEnvironment groupings being empty prior to executing the test.

ramperher: Okey, got it! Just to know that it was expected to do that change. I'll test it with a new change I have included in our example tested by tnf: https://softwarefactory-project.io/r/c/dci-openshift-app-agent/+/26228, just to check that, effectively, a non guaranteed pod with affinity rule enabled is tested. You'll see the results in future DCI jobs launched under this PR.

ramperher:
> * SUCCESS https://www.distributed-ci.io/jobs/b0079cf3-1b8d-4972-9984-0e00be01c9c3/jobStates
>   from [Turn groupings into env funcs #553](https://github.com/test-network-function/cnf-certification-test/pull/553)

Test looks fine from my side, confirming everything is working fine.

pkg/provider/provider.go
```diff
@@ -68,24 +68,16 @@ type TestEnvironment struct { // rename this with testTarget
...
-	DebugPods                map[string]*corev1.Pod // map from nodename to debugPod
-	GuaranteedPods           []*Pod
-	NonGuaranteedPods        []*Pod
-	AffinityRequiredPods     []*Pod
-	AntiAffinityRequiredPods []*Pod
-	HugepagesPods            []*Pod
-	AllPods                  []*Pod `json:"AllPods"`
+	Pods          []*Pod                 `json:"testPods"`
+	DebugPods     map[string]*corev1.Pod // map from nodename to debugPod
+	HugepagesPods []*Pod
```
greyerof, line 73: Can't we do another getter in filters_test.go for this HugepagesPods list?

greyerof (approved):
I also like the idea of having a clean env struct. Good job!

pkg/provider/provider.go
sebrandon1: Yes I can add that.

edcdavid (approved):

Issue #505, opened 2022-09-13
When using service-mesh in the pods under test, several tests failed because of istio-proxy container


ramperher:
Hi all,

With [this change](https://softwarefactory-project.io/r/c/dci-openshift-app-agent/+/26082) in dci-openshift-app-agent, I am testing the inclusion of service-mesh (using Aspenmesh) in the pods under test, in order to execute and validate `platform-alteration-service-mesh-usage test`.

I can confirm the test is working fine and passing. However, there are several tests that now are failing because of the `istio-proxy` container being deployed in the pods under test. These are the following:

- access-control-one-process-per-container
- observability-termination-policy
- networking-undeclared-container-ports-usage
- lifecycle-container-shutdown
- lifecycle-liveness
- lifecycle-pod-scheduling
- lifecycle-statefulset-scaling
- platform-alteration-isredhat-release

[DCI job](https://www.distributed-ci.io/jobs/d17c5dc9-2d8e-445d-b596-596771b725e4/tests) where you can see an example of execution.

Having said that, do you think it is feasible to omit `istio-proxy` container from the tests where containers are check? (because, for example, we can't do anything with lifecycle-statefulset-scaling, I suppose). I mean, while executing these tests, if the container name is `istio-proxy`, just do not execute the tests on that container. What's your opinion on this?

sebrandon1:
@edcdavid Do we just want to filter out any containers that are named `istio-proxy` from the env.Containers in general?

ramperher:
Is there any news regarding this? @sebrandon1 @edcdavid

ramperher:
I'll close this for the moment, if we work again with istio and we continue experiencing issues, I'll let you know

PR #554, opened 2022-09-27
Skip known container names

sebrandon1:
build-depends: 26082

In some recent tests where a partner is actually using Istio, there are a number container related tests that are failing/flagging istio-proxy containers.  The tests are:
```
networking-undeclared-container-ports-usage
observability-termination-policy
lifecycle-liveness
lifecycle-image-pull-policy
lifecycle-container-shutdown
```

So what I did was add a check function on the `Container` object that will allow us to ignore any containers that have a known container name that causes failures.

### DIFF
--- a/pkg/provider/provider.go
+++ b/pkg/provider/provider.go
@@ -169,7 +169,7 @@ func buildTestEnvironment() { //nolint:funlen
 	for i := 0; i < len(pods); i++ {
 		aNewPod := NewPod(&pods[i])
 		env.Pods = append(env.Pods, &aNewPod)
-		env.Containers = append(env.Containers, getPodContainers(&pods[i])...)
+		env.Containers = append(env.Containers, getPodContainers(&pods[i], true)...)
 	}
 	pods = data.AllPods
 	for i := 0; i < len(pods); i++ {
@@ -220,7 +220,7 @@ func buildTestEnvironment() { //nolint:funlen
 	logrus.Infof("Operators found: %d", len(env.Operators))
 }
 
-func getPodContainers(aPod *corev1.Pod) (containerList []*Container) {
+func getPodContainers(aPod *corev1.Pod, ignoreContainers bool) (containerList []*Container) {
 	for j := 0; j < len(aPod.Spec.Containers); j++ {
 		cut := &(aPod.Spec.Containers[j])
 		var state corev1.ContainerStatus
@@ -233,7 +233,13 @@ func getPodContainers(aPod *corev1.Pod) (containerList []*Container) {
 		container := Container{Podname: aPod.Name, Namespace: aPod.Namespace,
 			NodeName: aPod.Spec.NodeName, Container: cut, Status: state, Runtime: aRuntime, UID: uid,
 			ContainerImageIdentifier: buildContainerImageSource(aPod.Spec.Containers[j].Image)}
-		containerList = append(containerList, &container)
+
+		// Build slices of containers based on whether or not we are "ignoring" them or not.
+		if ignoreContainers && container.HasIgnoredContainerName() {
+			continue
+		} else {
+			containerList = append(containerList, &container)
+		}
 	}
 	return containerList
 }
--- a/pkg/provider/provider.go
+++ b/pkg/provider/provider.go
@@ -78,6 +78,7 @@ type TestEnvironment struct { // rename this with testTarget
 	// StatefulSet Groupings
 	StatefulSets []*StatefulSet `json:"testStatefulSets"`
 
+	// Note: Containers is a filtered list of objects based on a block list of disallowed container names.
 	Containers             []*Container `json:"testContainers"`
 	Operators              []Operator   `json:"testOperators"`
 	PersistentVolumes      []corev1.PersistentVolume
@@ -169,6 +170,7 @@ func buildTestEnvironment() { //nolint:funlen
 	for i := 0; i < len(pods); i++ {
 		aNewPod := NewPod(&pods[i])
 		env.Pods = append(env.Pods, &aNewPod)
+		// Note: 'getPodContainers' is returning a filtered list of Container objects.
 		env.Containers = append(env.Containers, getPodContainers(&pods[i], true)...)
 	}
 	pods = data.AllPods
@@ -220,7 +222,7 @@ func buildTestEnvironment() { //nolint:funlen
 	logrus.Infof("Operators found: %d", len(env.Operators))
 }
 
-func getPodContainers(aPod *corev1.Pod, ignoreContainers bool) (containerList []*Container) {
+func getPodContainers(aPod *corev1.Pod, useIgnoreList bool) (containerList []*Container) {
 	for j := 0; j < len(aPod.Spec.Containers); j++ {
 		cut := &(aPod.Spec.Containers[j])
 		var state corev1.ContainerStatus
@@ -235,7 +237,7 @@ func getPodContainers(aPod *corev1.Pod, ignoreContainers bool) (containerList []
 			ContainerImageIdentifier: buildContainerImageSource(aPod.Spec.Containers[j].Image)}
 
 		// Build slices of containers based on whether or not we are "ignoring" them or not.
-		if ignoreContainers && container.HasIgnoredContainerName() {
+		if useIgnoreList && container.HasIgnoredContainerName() {
 			continue
 		} else {
 			containerList = append(containerList, &container)
--- a/pkg/provider/containers.go
+++ b/pkg/provider/containers.go
@@ -26,6 +26,11 @@ import (
 	corev1 "k8s.io/api/core/v1"
 )
 
+var (
+	// Certain tests that have been known to fail because of injected containers (such as Istio) that fail certain tests.
+	ignoredContainerNames = []string{"istio-proxy"}
+)
+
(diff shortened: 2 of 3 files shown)


sebrandon1:
Marked WIP.  I'm going to switch up the way I do the filtering I think.

ramperher:
Hi @sebrandon1, I've been testing this with an OCP cluster using Aspenmesh, and I've seen that we are still having issues with some tests, continuing with what I reported on https://github.com/test-network-function/cnf-certification-test/issues/505 issue.

Comparing the results from, for example, [this job](https://www.distributed-ci.io/jobs/5a8cf677-5ccf-43d8-9780-5c998758e5b2/tests) and the job reported in the https://github.com/test-network-function/cnf-certification-test/issues/505 issue, we have the following.

Right now, there are some tests that are now passing, and are:

- observability-termination-policy 
- lifecycle-container-shutdown
- lifecycle-liveness-probe

We are still having problems, particularly, in the following tests:

- access-control-one-process-per-container
- networking-undeclared-container-ports-usage
- platform-alteration-isredhat-release
- lifecycle-pod-scheduling
- lifecycle-statefulset-scaling - stuck here around 60 seconds

And not, we have also these tests failing (because we are executing all tests available in this PR's code, which are right now in the latest commit from main branch):

- access-control-ssh-daemons
- networking-reserved-partner-ports
- networking-nftables
- lifecycle-startup-probe
- lifecycle-deployment-scaling - stuck here around 60 seconds
- lifecycle-pod-recreation - stuck here around 400-1200 seconds

So, the new list of tests that are having issues is:

- access-control-one-process-per-container
- access-control-ssh-daemons
- networking-reserved-partner-ports
- networking-nftables
- networking-undeclared-container-ports-usage
- platform-alteration-isredhat-release
- lifecycle-pod-scheduling
- lifecycle-statefulset-scaling - stuck here around 60 seconds
- lifecycle-startup-probe
- lifecycle-deployment-scaling - stuck here around 60 seconds
- lifecycle-pod-recreation - stuck here around 400-1200 seconds

For more information about the particular failures for each test, you can see the DCI job aforementioned (it is the last one reported by dci-bot)

Hope it helps.

sebrandon1:
I'll take a look at your findings!

sebrandon1:
@ramperher Looking at the output from [here](https://www.distributed-ci.io/jobs/bd5494e9-fc96-42eb-8edd-487a7c99f53b/jobStates) it seems like its healthier.

pkg/provider/provider.go
```diff
@@ -195,7 +195,9 @@ func buildTestEnvironment() { //nolint:funlen,gocyclo
 		} else {
 			env.AntiAffinityRequiredPods = append(env.AntiAffinityRequiredPods, &aNewPod)
 		}
-		env.Containers = append(env.Containers, getPodContainers(&pods[i])...)
+
+		// Build slices of containers and containers specifically not ignored globally.
+		env.Containers = append(env.Containers, getPodContainers(&pods[i], true)...)
```
edcdavid, line 200: Seems like the container list in the Pod object is not filtered while the container list itself is filtered. Seems confusing to me. Shouldn't we have a dedicated filtered object?

ramperher:
> @ramperher Looking at the output from [here](https://www.distributed-ci.io/jobs/bd5494e9-fc96-42eb-8edd-487a7c99f53b/jobStates) it seems like its healthier.

I have to check in a cluster with istio installed, the default DCI jobs that are launched are not deployed in a cluster with istio :sweat_smile: I'll try a run later on under these conditions and I'll tell you!

sebrandon1:
Sounds good @ramperher looking forward to the results.

pkg/provider/provider.go
sebrandon1: Yes I did this on purpose.  The Service Mesh test uses `containers` derived from the pod itself, but the `env.Containers` is used everywhere else.  So I added a boolean to `getPodContainers()` to be able to skip the filtering for containers stored in the `Pod` object, but apply the filter for the overall list.

ramperher:
> * SUCCESS https://www.distributed-ci.io/jobs/f4e1cb57-af64-4d0a-af29-a925b277654f/jobStates
>   from [[wip] Enable platform-alteration-service-mesh-usage tnf test dci-labs/dallas-pipelines#480](https://github.com/dci-labs/dallas-pipelines/pull/480)

@sebrandon1 we are improving, now three tests have been fixed:

- lifecycle-startup-probe
- access-control-ssh-daemons
- platform-alteration-isredhat-release

But these others are still failing:

- access-control-one-process-per-container
- networking-reserved-partner-ports
- networking-nftables
- networking-undeclared-container-ports-usage
- lifecycle-pod-scheduling
- lifecycle-statefulset-scaling - stuck here around 60 seconds
- lifecycle-deployment-scaling - stuck here around 60 seconds
- lifecycle-pod-recreation - stuck here around 400-1200 seconds

And also, this one is failing too (I didn't report it yesterday because it failed in some jobs but passed in others, but it seems it has to be checked too):

- observability-pod-disruption-budget

Finally, I've seen that lifecycle-cpu-isolation is being skipped in the previous executions and this last one, whereas it is executed currently in stable jobs.

so, gathering all the info here, and trying to check case by case:

- failed:
  - access-control-one-process-per-container - all the pods are hitting this issue. However, in normal DCI jobs, this is never failing, so my impression is that istio-proxy is affecting here, maybe also creating processes in the other container in the pod that causes that the container has more than one process running
  - networking-reserved-partner-ports - same containers that failed in the previous test are appearing here as having reserved ports, mainly 15xxx
  - networking-nftables - by checking the logs, it is clear that istio includes nftables rules that make this test to fail
  - networking-undeclared-container-ports-usage - port 15090 appears here
  - lifecycle-pod-scheduling - istio implies the presence of node selector clauses
  - lifecycle-statefulset-scaling (stuck here around 60 seconds)
  - lifecycle-deployment-scaling (stuck here around 60 seconds)
  - lifecycle-pod-recreation (stuck here around 400-1200 seconds) - in these three tests, my impression is that the presence of istio-proxy container difficults the work of doing scaling or pod recreation. In fact, in a Deployment under test, the second pod was not even deployed
  - observability-pod-disruption-budget - I think this test failed when the Deployment is not able to deploy all the pods again when doing the pod recreation.
- skipped:
  - lifecycle-cpu-isolation - I'm under the impression that the presence of istio makes some transformation in the pods, so that some of the conditions to be met to pass this test on guaranteed pods are removed, and due to that, the test is skipped, as there are no pods that fulfils the conditions.

My conclusion: I'm under the impression we cannot do much more changes on cnf-certification-test side, because these failures are really tied to istio. So, here, I am thinking on two possible ways of handling this: 1) admit these tests can fail in case we have a cluster with istio and service-mesh configured, and not affecting to the certification results (and document consequently this), 2) continue checking if there's a way of making these test to pass under these circumstances, which from my point of view, will be difficult in some cases.

pkg/provider/provider.go
edcdavid: right, I am jut saying that it might be confusing to someone going through the container list and expect it to match the pod's container list since there is no indication that the container list is filtered (other than looking at the code).

sebrandon1: This sounds like a good compromise.  I'll add a comment on the struct.

sebrandon1:
I wonder if this might be a good use of tags.  We could potentially tag all of the tests that we know fail with Istio installed and skip them.

edcdavid (approved):

sebrandon1:
I think as-is this PR covers its case but there is definitely more to do for detecting clusters that are using istio and either handling potential false flags or skipping.

ramperher (approved):
I agree with you, @sebrandon1. Thanks for all this effort! But yes, after this PR, it should be considered what I commented before; probably to add new tags in the tests that are expected to fail when istio is being used and to warn about it in the README. Also, to clarify what's the impact in terms of certification process when istio is present, as there are several tests that fail because of that. But as I said, it can be checked in a different PR.

pkg/provider/provider.go
```diff
@@ -220,7 +222,7 @@ func buildTestEnvironment() { //nolint:funlen
 	logrus.Infof("Operators found: %d", len(env.Operators))
 }
 
-func getPodContainers(aPod *corev1.Pod) (containerList []*Container) {
+func getPodContainers(aPod *corev1.Pod, ignoreContainers bool) (containerList []*Container) {
```
greyerof, line 225: Suggestion: `ignoreContainers` seems a weird name to me for a param of a function called `getPodContainers`. What about just passing a map of black-listed names to filter out?
```
func getPodContainers(aPod *corev1.Pod, blackListedNames map[string]bool) (containerList []*Container) {
    	for j := 0; j < len(aPod.Spec.Containers); j++ {
		    cut := &(aPod.Spec.Containers[j])
		    if blackListedNames[cut.Name] {
		        continue
		    }
		    
		    container := Container{Podname: aPod.Name, Namespace: aPod.Namespace,
		    ...
		    ...
```

sebrandon1: Good idea, I wonder if I can pass that list through.  I'll try it.

sebrandon1: I think changing the name to something like `useIgnoreList` might be more useful.

PR #560, opened 2022-09-29
Clean up reporting

sebrandon1:
Changes: 
- Removed extraneous `Container: %s` strings when all is needed is `%s`.
- Added claim file prints for the `testDeploymentScaling` func.
- Fixed `FindRogueContainersDeclaringPorts` and `testUndeclaredContainerPortsUsage` to only report a port once to avoid output like:

```
pod: test-0 ns: production-cnf is listening on port 15090 protocol %!d(string=TCP), but that port was not declared in any container spec.
pod: test-0 ns: production-cnf is listening on port 15090 protocol %!d(string=TCP), but that port was not declared in any container spec.
pod: test-0 ns: production-cnf is listening on port 15090 protocol %!d(string=TCP), but that port was not declared in any container spec.
pod: test-1 ns: production-cnf is listening on port 15090 protocol %!d(string=TCP), but that port was not declared in any container spec.
pod: test-1 ns: production-cnf is listening on port 15090 protocol %!d(string=TCP), but that port was not declared in any container spec.
pod: test-1 ns: production-cnf is listening on port 15090 protocol %!d(string=TCP), but that port was not declared in any container spec.
pod: test-1 ns: production-cnf is listening on port 15090 protocol %!d(string=TCP), but that port was not declared in any container spec.
pod: test-1 ns: production-cnf is listening on port 15090 protocol %!d(string=TCP), but that port was not declared in any container spec.
pod: test-1 ns: production-cnf is listening on port 15090 protocol %!d(string=TCP), but that port was not declared in any container spec.
```

### DIFF
--- a/cnf-certification-test/networking/netcommons/netcommons.go
+++ b/cnf-certification-test/networking/netcommons/netcommons.go
@@ -134,6 +134,15 @@ func GetIPVersion(aIP string) (IPVersion, error) {
 	return IPv6, nil
 }
 
+func ExistsInPortInfoSlice(lp netutil.PortInfo, lps []netutil.PortInfo) bool {
+	for _, i := range lps {
+		if lp.PortNumber == i.PortNumber && lp.Protocol == i.Protocol {
+			return true
+		}
+	}
+	return false
+}
+
 // FilterIPListByIPVersion filters a list of ip strings by the provided version
 // e.g. a list of mixed ipv4 and ipv6 when filtered with ipv6 version will return a list with
 // the ipv6 addresses
@@ -170,8 +179,10 @@ func FindRoguePodsListeningToPorts(pods []*provider.Pod, portsToTest map[int32]b
 			failedContainers++
 			continue
 		}
+		var portsAlreadyReportedRogue []netutil.PortInfo
 		for port := range listeningPorts {
-			if portsToTest[int32(port.PortNumber)] {
+			if portsToTest[int32(port.PortNumber)] && !ExistsInPortInfoSlice(port, portsAlreadyReportedRogue) {
+				portsAlreadyReportedRogue = append(portsAlreadyReportedRogue, port)
 				tnf.ClaimFilePrintf("%s has one container listening on port %d that has been reserved", put, port.PortNumber)
 				roguePods = append(roguePods, put.String())
 			}
--- a/cnf-certification-test/networking/suite.go
+++ b/cnf-certification-test/networking/suite.go
@@ -153,9 +153,11 @@ func testUndeclaredContainerPortsUsage(env *provider.TestEnvironment) {
 
 		// Verify that all the listening ports have been declared in the container spec
 		failedPod := false
+		var listeningPortsAlreadyReported []netutil.PortInfo // short circuit to not report the same port more than once
 		for listeningPort := range listeningPorts {
-			if !declaredPorts[listeningPort] {
-				tnf.ClaimFilePrintf("%s is listening on port %d protocol %d, but that port was not declared in any container spec.", put, portInfo.PortNumber, portInfo.Protocol)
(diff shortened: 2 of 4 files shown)


cnf-certification-test/networking/suite.go
```diff
@@ -153,9 +153,11 @@ func testUndeclaredContainerPortsUsage(env *provider.TestEnvironment) {
 
 		// Verify that all the listening ports have been declared in the container spec
 		failedPod := false
+		var listeningPortsAlreadyReported []netutil.PortInfo // short circuit to not report the same port more than once
 		for listeningPort := range listeningPorts {
-			if !declaredPorts[listeningPort] {
-				tnf.ClaimFilePrintf("%s is listening on port %d protocol %d, but that port was not declared in any container spec.", put, portInfo.PortNumber, portInfo.Protocol)
+			if !declaredPorts[listeningPort] && !netcommons.ExistsInPortInfoSlice(listeningPort, listeningPortsAlreadyReported) {
```
greyerof, line 158: I think this filter you add here is hiding a possible bug. I've reviewed the code of this TC (testUndeclaredContainerPortsUsage) but I still don't see why it should print the same port several times... It doesn't make sense. This filter shouldn't be necessary. We're missing something here.

This tc gets the list of declared ports in the containers manifest of each pod and then gets the list of all ports that are actually opened (and listening) by the containers' apps on that pod only. As all containers in a pod share the same networking namespace, it's not possible for the same port to be opened (and listening) more than 1 time. That's why we run the `ss` command through `nsenter` on the first container of that pod only: we should also see the ports opened by the other containers in the same pod.

cnf-certification-test/networking/netcommons/netcommons.go
```diff
@@ -150,8 +168,10 @@ func FilterIPListByIPVersion(ipList []string, aIPVersion IPVersion) []string {
 func FindRogueContainersDeclaringPorts(containers []*provider.Container, portsToTest map[int32]bool) []string {
 	var rogueContainers []string
 	for _, cut := range containers {
+		var portsAlreadyReportedRogue []int32
 		for _, port := range cut.Ports {
-			if portsToTest[port.ContainerPort] {
+			if portsToTest[port.ContainerPort] && !ExistsInInt32Slice(port.ContainerPort, portsAlreadyReportedRogue) {
```
greyerof, line 173: Do we really want this? Containers from different pods are allowed to open the same port number. I believe we should print a trace for each container so partners can fix several pods before running the TNF cert tool again.

sebrandon1: Let me look at this again, you may be right.

sebrandon1: Okay I removed this one.

cnf-certification-test/networking/suite.go
sebrandon1: @greyerof I was just playing around with the `ss` command on my kind cluster and I see:
```
[root@kind-worker3 /]# ss -tulwn | grep 9100
tcp   LISTEN 0      4096       127.0.0.1:9100       0.0.0.0:*          
tcp   LISTEN 0      4096      172.18.0.4:9100       0.0.0.0:*    
```

I wonder if that's what's happening with the multiple printouts where the localhost and the IP of the host are showing up in the list twice (or more).

In my case above `172.18.0.4` is the IP of my worker3.

greyerof: Yep. I don't think that worker's port appears when using nsenter, so there should be something else.

sebrandon1: I will update my PR to remove this part of the cleanup so we can focus on a potential fix to the `ss` command or however we interpret the output.

sebrandon1:
Closing this in favor of: https://github.com/test-network-function/cnf-certification-test/pull/564

#564 contains everything minus the changes to the networking suite.

Issue #551, opened 2022-09-25
Catalog.md should indicate which test falls under "extended test cases"


mwlinca:
Extended test set is a new suite of test cases introduced in v4.1 release, in which the option of whether to include these extended test cases in the execution of CNF Certification Test Suite can be configured in tnf_config.yml

Correspondingly, the catalog.md file should reflect which test cases fall under "extended test set", the execution of which is configurable via tnf_config.yml.

PR #562, opened 2022-09-30
Add tags to the catalog entries

sebrandon1:
Resolves: #551 

Any catalog entry generated by the `AddCatalogEntry` helper function will now include its tags in the catalog.  This will be the initial commit and we will have to have a subsequent commit that translates all of the existing catalog entries to use the helper function.


greyerof (commented):
Shouldn't we also print the "common" tag?

edcdavid (approved):

greyerof (approved):

PR #564, opened 2022-10-03
Reporting cleanup

sebrandon1:
Follow up to: https://github.com/test-network-function/cnf-certification-test/pull/560

This is everything from #560 minus the changes for the networking suite that might have covered up a potential bug with how we were using `ss`.

### DIFF
--- a/cnf-certification-test/networking/suite.go
+++ b/cnf-certification-test/networking/suite.go
@@ -155,7 +155,7 @@ func testUndeclaredContainerPortsUsage(env *provider.TestEnvironment) {
 		failedPod := false
 		for listeningPort := range listeningPorts {
 			if !declaredPorts[listeningPort] {
-				tnf.ClaimFilePrintf("%s is listening on port %d protocol %s, but that port was not declared in any container spec.", put, portInfo.PortNumber, portInfo.Protocol)
+				tnf.ClaimFilePrintf("%s is listening on port %d protocol %s, but that port was not declared in any container spec.", put, listeningPort.PortNumber, listeningPort.Protocol)
 				failedPod = true
 			}
 		}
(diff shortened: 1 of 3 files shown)


greyerof (approved):

cnf-certification-test/networking/suite.go
```diff
@@ -155,7 +155,7 @@ func testUndeclaredContainerPortsUsage(env *provider.TestEnvironment) {
 		failedPod := false
 		for listeningPort := range listeningPorts {
 			if !declaredPorts[listeningPort] {
-				tnf.ClaimFilePrintf("%s is listening on port %d protocol %d, but that port was not declared in any container spec.", put, portInfo.PortNumber, portInfo.Protocol)
+				tnf.ClaimFilePrintf("%s is listening on port %d protocol %s, but that port was not declared in any container spec.", put, portInfo.PortNumber, portInfo.Protocol)
```
greyerof, line 158: So I finally found the reason for the same trace appearing [several times for the same port](https://github.com/test-network-function/cnf-certification-test/pull/560#issue-1391331519): this trace is wrongly printing `portInfo` instead of `listeningPort`. Can you fix that in this PR, please?
```
tnf.ClaimFilePrintf("%s is listening on port %d protocol %s, but that port was not declared in any container spec.", put, listeningPort.PortNumber, listeningPort.Protocol)
```
This won't make the TC to pass when istio is used, but at least the port will appear once, as it should be.

sebrandon1: Oh great!  Good catch.  I'll fix it.

sebrandon1: Fixed.

edcdavid (approved):

PR #572, opened 2022-10-06
Adjust online check to use 'correct' hash

sebrandon1:
This is (what we believe) is a database problem workaround.  @edcdavid and I tracked the function back to the API and it seems there's a mismatch in the upstream database.  See: https://catalog.redhat.com/api/containers/v1/repositories/registry/registry.connect.redhat.com/repository/rocketchat/rocketchat/tag/0.56.0-1


sebrandon1:
Closing this, as it looks like the API is now fixed.

PR #579, opened 2022-10-12
Adding postStart Test case

edcdavid:
Similar to preStop check, adding a check for postStart. Also harmonized description strings for the preStrop test


ramperher:
> 

I've seen the command is not working in your case because you don't have permissions to interact with our tools, I have to check that. I'll test the command with my user to see if it sends the message fine.

cnf-certification-test/lifecycle/suite.go
```diff
@@ -183,6 +189,22 @@ func testContainersPreStop(env *provider.TestEnvironment) {
...
+		logrus.Debugln("check container ", cut.String(), " post start lifecycle ")
+
+		if cut.Lifecycle == nil || (cut.Lifecycle != nil && cut.Lifecycle.PostStart == nil) {
+			badContainers = append(badContainers, cut.Name)
+			tnf.ClaimFilePrintf("%s does not have postStart defined", cut)
+		}
+	}
+	if len(badContainers) > 0 {
+		tnf.ClaimFilePrintf("Containers have been found missing lifecycle postStart definitions: %v", badContainers)
+		ginkgo.Fail("Containers have been found missing lifecycle postStart definitions.")
```
sebrandon1, line 204: For this TC, and the `preStop` TC, should we switch them to use the `testhelper.AddTestResultLog` helper func?

sebrandon1 (approved):

cnf-certification-test/lifecycle/suite.go
edcdavid: fixed

greyerof (approved):

PR #592, opened 2022-10-14
Add preflight-lib reference

sebrandon1:
This is my WIP branch adding support for the tests written in the `openshift-preflight` repo and automatically importing them into our ginkgo output.


ramperher:
In our team, we'll be quite interested in this, because if we can execute tnf and preflight from the same point and in one single run, it would be great. Will you define instructions to explain how to run preflight?

sebrandon1:
@ramperher I'll probably have a "demo" or a short video or something to describe what I'm doing in more detail.  But the idea behind this is that I created what I'm calling the preflight `lib` folder.  See my branch [here](https://github.com/sebrandon1/openshift-preflight/tree/preflight_lib_test/lib).  What I did was split out all of the logic that the current `cmd` package contains for executing their checks inside of openshift-preflight and moved it to `lib`.  This effectively splits out the cobra/viper code from the execution code.

By making these routines public, I was able to build up my own (container-only as of the time of this comment) runners and execute the preflight code straight from the TNF code base.  I created a new "suite" on our side called "preflight" and you just include that the way you would include access-control, networking, etc.  

On our side, the Ginkgo It blocks are created dynamically so lets say if the upstream openshift-preflight repo adds more tests, all we have to do is bump the release version of preflight and pull it in and it will automatically run it in TNF.

I am still working on the "operator side" of the preflight code this week but it should be very similar to what I did for the container side.

ramperher:
> @ramperher I'll probably have a "demo" or a short video or something to describe what I'm doing in more detail. But the idea behind this is that I created what I'm calling the preflight `lib` folder. See my branch [here](https://github.com/sebrandon1/openshift-preflight/tree/preflight_lib_test/lib). What I did was split out all of the logic that the current `cmd` package contains for executing their checks inside of openshift-preflight and moved it to `lib`. This effectively splits out the cobra/viper code from the execution code.
> 
> By making these routines public, I was able to build up my own (container-only as of the time of this comment) runners and execute the preflight code straight from the TNF code base. I created a new "suite" on our side called "preflight" and you just include that the way you would include access-control, networking, etc.
> 
> On our side, the Ginkgo It blocks are created dynamically so lets say if the upstream openshift-preflight repo adds more tests, all we have to do is bump the release version of preflight and pull it in and it will automatically run it in TNF.
> 
> I am still working on the "operator side" of the preflight code this week but it should be very similar to what I did for the container side.

It sounds nice! if you need to test it with DCI when you have something ready, don't hesitate to tell us, we can give it a try!

sebrandon1:
https://github.com/redhat-openshift-ecosystem/openshift-preflight/pull/809 has merged and I will be using this preflight branch as my testing ground: https://github.com/komish/openshift-preflight/tree/add-lib-ha

PR #596, opened 2022-10-18
DPDK cpu exec check

bnshr:


cnf-certification-test/networking/suite.go
```diff
@@ -121,8 +121,35 @@ var _ = ginkgo.Describe(common.NetworkingTestKey, func() {
...
+	testhelper.SkipIfEmptyAny(ginkgo.Skip, dpdkPods)
+
+	ginkgo.By("Check if cpu pinning is enabled")
+
+	cpuPinningPods := provider.GetCPUPinningPods(dpdkPods)
+	testhelper.SkipIfEmptyAny(ginkgo.Skip, cpuPinningPods)
+
+	ginkgo.By("Check exec probe is not happening")
+
+	for _, cpuPinnedPod := range cpuPinningPods {
```
edcdavid, line 144: Add a cpuPinningPodsWithDpdk() filter in provider/pods.go to directly get the pods here

cnf-certification-test/networking/suite.go
```diff
@@ -121,8 +121,35 @@ var _ = ginkgo.Describe(common.NetworkingTestKey, func() {
 		testhelper.SkipIfEmptyAny(ginkgo.Skip, env.Pods)
 		testPartnerSpecificTCPPorts(&env)
 	})
+	testID, tags = identifiers.GetGinkgoTestIDAndLabels(identifiers.TestDpdkCPUPinningExecProbe)
+	ginkgo.It(testID, ginkgo.Label(tags...), func() {
+		testhelper.SkipIfEmptyAny(ginkgo.Skip, env.Pods)
```
edcdavid, line 126: Add a cpuPinningPodsWithDpdk() filter in provider/pods.go to directly skip this test if such pods are not detected since it will be not applicable

cnf-certification-test/networking/suite.go
```diff
@@ -121,8 +121,30 @@ var _ = ginkgo.Describe(common.NetworkingTestKey, func() {
...
+		testExecProbDenyAtCPUPinning(&env)
+	})
 })
 
+func testExecProbDenyAtCPUPinning(env *provider.TestEnvironment) {
```
edcdavid, line 131: Change:
func testExecProbDenyAtCPUPinning(env *provider.TestEnvironment) {
	ginkgo.By("Check if DPDK pod with cpu pinning is running")

	dpdkPods := env.GetCPUPinningPodsWithDpdk()
	testhelper.SkipIfEmptyAny(ginkgo.Skip, dpdkPods)
to:
func testExecProbDenyAtCPUPinning(env *provider.TestEnvironment, dpdkPods []*provider.Pod) {

cnf-certification-test/networking/suite.go
```diff
@@ -121,8 +121,30 @@ var _ = ginkgo.Describe(common.NetworkingTestKey, func() {
 		testhelper.SkipIfEmptyAny(ginkgo.Skip, env.Pods)
 		testPartnerSpecificTCPPorts(&env)
 	})
+	testID, tags = identifiers.GetGinkgoTestIDAndLabels(identifiers.TestDpdkCPUPinningExecProbe)
+	ginkgo.It(testID, ginkgo.Label(tags...), func() {
+		testhelper.SkipIfEmptyAny(ginkgo.Skip, env.Pods)
```
edcdavid, line 126: change:
testhelper.SkipIfEmptyAny(ginkgo.Skip, env.Pods)
to:	
dpdkPods := env.GetCPUPinningPodsWithDpdk()
testhelper.SkipIfEmptyAny(ginkgo.Skip, dpdkPods)
testExecProbDenyAtCPUPinning(&env, dpdkPods)

cnf-certification-test/networking/suite.go
```diff
@@ -121,8 +121,26 @@ var _ = ginkgo.Describe(common.NetworkingTestKey, func() {
...
+		testExecProbDenyAtCPUPinning(dpdkPods)
+	})
 })
 
+func testExecProbDenyAtCPUPinning(dpdkPods []*provider.Pod) {
+	ginkgo.By("Check if exec probe is happening")
+
+	for _, cpuPinnedPod := range dpdkPods {
+		for _, x := range cpuPinnedPod.Containers {
+			if len(x.Command) != 0 || len(x.Args) != 0 {
```
edcdavid, line 137: is this the right command to check here? This seems to be the main entry point for the container which is ok to block. Shouldn't the test be for liveliness and readiness probes that are happening in parallel?

bnshr: I am not sure why liveliness and readiness probe in parallel should be blocked. @mwlinca Is that what is required?

ramperher:
DCI job is not working because of an issue with Pyxis API that is affecting to affiliated-certification-container-is-certified. Apart from this, this change is not breaking DCI, so it's fine (in our case, we are not using DPDK in our test workload, so this test is omitted in our setup)

pkg/provider/filters.go
```diff
@@ -115,3 +123,78 @@ func (env *TestEnvironment) GetHugepagesPods() []*Pod {
...
+				filteredPods = append(filteredPods, pod)
+			}
+		}
+	}
+	return filteredPods
+}
+
+func filterCPUPinningPods(pods []*Pod) []*Pod {
+	const (
+		verifyCPUAffinityCommand = "for pid in $(ps -A -o pid=); do taskset -pc $pid 2>/dev/null; done"
```
greyerof, line 159: Here you rely on the "taskset" program to be available in the CNF's container image. It might be installed on UBI-based images, but what if partners are using they're own images with file systems that do not have the taskset program installed? We can leave this implementation, but, at least, I recommend investigating if it's possible to use `nsenter` for this purpose.

pkg/provider/filters.go
```diff
@@ -115,3 +123,78 @@ func (env *TestEnvironment) GetHugepagesPods() []*Pod {
...
+	if len(pods) == 0 {
+		return filteredPods
+	}
+
+	o := clientsholder.GetClientsHolder()
+
+	for _, pod := range pods {
+		ctx := clientsholder.NewContext(pod.Namespace, pod.Name, pod.Spec.Containers[0].Name)
+		outStr, errStr, err := o.ExecCommandContainer(ctx, verifyCPUAffinityCommand)
+		if outStr == "" {
```
greyerof, line 175: Why not just `if outStr == "" && err != nil && errStr != "" {` ?

Other than that, although it might not be the case here, just note that stdout and stderr are not mutually exclusive. Some programs will print you output in both streams. I recommend this pattern here:
```
outStr, errStr, err := o.ExecCommandContainer(ctx, verifyCPUAffinityCommand)
if err != nil || errStr != "" {
  logrus.Errorf("failed to execute command...", pod, command, errStr, outStr, err) // see I also added outStr here as extra debugging info.
  continue
}

lines := strings.Split(outStr, newLineCharacter)
```

bnshr: Thanks for the recommendation, but it is not applicable here. The command executed in the container would give error each time and that's why the error output is sent to null device. So, the output of the command has to be checked at first.

bnshr: Why not `if outStr == "" && err != nil && errStr != "" {` -> I did not choose this for the better readability of the flow. The focus is more into the outputStr.

pkg/provider/filters.go
bnshr: taskset is available in UBI image. Could you share me some partner image where I can verify again?

pkg/provider/filters.go
```diff
@@ -115,3 +123,78 @@ func (env *TestEnvironment) GetHugepagesPods() []*Pod {
...
+func filterCPUPinningPods(pods []*Pod) []*Pod {
+	const (
+		verifyCPUAffinityCommand = "for pid in $(ps -A -o pid=); do taskset -pc $pid 2>/dev/null; done"
+		newLineCharacter         = "\n"
+		spaceCharacter           = " "
+	)
+
+	var filteredPods []*Pod
+
+	if len(pods) == 0 {
```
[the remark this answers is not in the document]
bnshr: I suppose, I did not want to instantiate client holder, for nothing in that case `o := clientsholder.GetClientsHolder()` - Takes memory space!

cnf-certification-test/networking/suite.go
```diff
@@ -121,8 +121,26 @@ var _ = ginkgo.Describe(common.NetworkingTestKey, func() {
...
+	})
 })
 
+func testExecProbDenyAtCPUPinning(dpdkPods []*provider.Pod) {
+	ginkgo.By("Check if exec probe is happening")
+
+	for _, cpuPinnedPod := range dpdkPods {
+		for _, x := range cpuPinnedPod.Containers {
+			if x.LivenessProbe != nil || x.StartupProbe != nil || x.ReadinessProbe != nil {
+				ginkgo.Fail("Exec prob is not allowed")
```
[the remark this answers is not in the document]
sebrandon1: That was @edcdavid's func 😄 but yes that's the way to do it now.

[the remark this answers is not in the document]
sebrandon1: Agreed with the above comment about collecting all of the failing containers into a slice.

pkg/provider/filters.go
edcdavid: After discussion between @greyerof @bnshr and myself, we agree to the following assumption for now: all guaranteed pods are assumed to be using cpu pinning. So to get the list of pods using cpu pinning is equivalent to the list of guaranteed pods ( already in the testenvironment struct).

cnf-certification-test/networking/suite.go
edcdavid: exec probes confirmed this morning

pkg/provider/filters.go
sebrandon1: Oh great!  We get to re-use that slice.

pkg/provider/filters.go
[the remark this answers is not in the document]
greyerof: Ok, but note that the clientsholder is a singleton, so it's instantiated the first time the GetClientsHolder() function is called, which happens right [before any TC has run yet](https://github.com/test-network-function/cnf-certification-test/blob/06a1a514635ffd893f6f30c7420383b25d3cffaa/cnf-certification-test/suite_test.go#L151). Further calls to GetClientsHolder() will just retrieve the pointer to the previously initialized struct.

pkg/provider/filters.go
greyerof: @bnshr , I don't understand your request. Partners may use whatever base image they want, but it's not possible to tell/predict which one in advance. They could also delete a lot of (not-used) binaries/programs from the base image just for the sake of reducing the size of the final container image.

pkg/provider/filters.go
greyerof: Ok, I just tried to convey that, in go, after invoking any function that can return any error (along with other values), it's idiomatic to check the error variable first. As a general rule, if err is not nil, the rest of the returned values are usually expected to be zeroed/empty/nil. See go's [builtin os.ReadDir](https://github.com/golang/go/blob/939f9fd64a740ef770a0ca3588f8c193dceebe83/src/os/dir.go#L118 ) function.

The `ExecCommandContainer()` function could also fail because of [errors occurred even before](https://github.com/test-network-function/cnf-certification-test/blob/a55d59cf11149b3895b3ff0a8996ea704d4fb707/internal/clientsholder/command.go#L60) the command can actually run inside the container: networking issues, the container has just stopped or is in a CrashLoopBack phase... In those cases, I don't see quite useful to check stdOut first.

pkg/provider/filters.go
bnshr: @edcdavid Is it true that these dpdk pods would be running with QoS as Guaranteed always? Or we need to verify that also from our end?

pkg/provider/filters.go
bnshr: Thanks, it seems to me that you did not run the command that we are executing it. It's simply not applicable to check err at first, I know about the best practices but we have to check result at first to track the valid error. I would suggest to run the `verifyCPUAffinityCommand` inside a container to test it. Btw, I do not think that we would need this piece of code anymore as we would probably not check affinity with `taskset`.

pkg/provider/filters.go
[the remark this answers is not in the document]
bnshr: Ah, cool!

pkg/provider/filters.go
edcdavid: I am not sure if dpdk pods need guaranteed qos to work at all, they should not if there are enough cores in the cpuset (as application is expecting) . However so far all DPDK apps in RAN space that I have seen needed the guaranteed QOS so that they can achieve their performance target. In my understanding the testcase itself is only applying to these guaranteed pods with extra modifications (remove interupts, etc) so as to really use 100% of the core with no noise. checking that we have DPDK in a guaranteed pod should be enough to identify pods to run this test in.

bnshr: Thanks!

pkg/provider/filters.go
```diff
@@ -115,3 +123,32 @@ func (env *TestEnvironment) GetHugepagesPods() []*Pod {
...
+
+func (env *TestEnvironment) GetCPUPinningPodsWithDpdk() []*Pod {
+	return filterDPDKRunningPods(env.GetGuaranteedPods())
+}
+
+func filterDPDKRunningPods(pods []*Pod) []*Pod {
+	var filteredPods []*Pod
+	const (
+		dpdkDriver           = "vfio-pci"
+		findDeviceSubCommand = "find . -name"
```
edcdavid, line 135: "find . -name" -> "find /sys -name"

edcdavid (approved):

sebrandon1 (approved):

edcdavid (commented):
/re-run

edcdavid (commented):
/dci-rerun

ramperher:
> @ramperher the DCI stays in "pending" status. Any idea why this is happening?

Maybe it got stuck somehow, I'll schedule a new one, anyway I saw in the past that it was passing fine.

ramperher:
> > @ramperher the DCI stays in "pending" status. Any idea why this is happening?
> 
> Maybe it got stuck somehow, I'll schedule a new one, anyway I saw in the past that it was passing fine.

@edcdavid I have the answer, DCI job is in pending status because the job was not launched, our logic was not able to rebase this PR code from main branch. If you need DCI job green check to merge, the change must be rebased and synchronized with main branch before.
Anyway, from my side it's fine to merge, I've already seen this was passing in previous changes, it does not affect to us.

edcdavid (commented):
/dci-rerun

edcdavid (commented):
/dci-rerun

PR #607, opened 2022-10-24
add new test for the sec context

aabughosh:
add new filed for the checking

fix the istioProxyContainerUID

rebase the PR

update the PR with fixing mnake lint

### DIFF
--- a/cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go
+++ b/cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go
@@ -31,7 +31,7 @@ var (
 	category3AddCapabilities = []string{"NET_ADMIN, NET_RAW"}
 	category4AddCapabilities = []string{"NET_ADMIN, NET_RAW, IPC_LOCK"}
 	Allowvolumes             = []string{"configMap, downwardAPI, emptyDir,persistentVolumeClaim,projected,secret"}
-	Catagory1                = ContainerSCC{false,
+	Category1                = ContainerSCC{false,
 		false,
 		false,
 		false,
@@ -47,7 +47,7 @@ var (
 		true,
 		true}
 
-	Catagory2 = ContainerSCC{false,
+	Category2 = ContainerSCC{false,
 		false,
 		false,
 		false,
@@ -63,7 +63,7 @@ var (
 		true,
 		true}
 
-	Catagory3 = ContainerSCC{false,
+	Category3 = ContainerSCC{false,
 		false,
 		false,
 		false,
@@ -78,7 +78,7 @@ var (
 		"catagory3",
 		true,
 		true}
-	Catagory4 = ContainerSCC{false,
+	Category4 = ContainerSCC{false,
 		false,
 		false,
 		false,
@@ -206,13 +206,13 @@ func Checkcategory(containers []v1.Container, containerSCC ContainerSCC) []strin
 		percontainerSCC := GetContainerSCC(cut, containerSCC)
 		// after building the containerSCC need to check to which category it is
 		switch percontainerSCC {
-		case Catagory1:
+		case Category1:
 			logrus.Info("is ok")
-		case Catagory2:
+		case Category2:
 			logrus.Info("is ok")
-		case Catagory3:
+		case Category3:
 			logrus.Info("is ok")
-		case Catagory4:
+		case Category4:
 			logrus.Info("is ok")
 		default:
 			badCcontainer = append(badCcontainer, cut.Name)
--- a/cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go
+++ b/cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go
@@ -228,6 +228,7 @@ const (
 	CategoryID4String       = "OtherTypes"
 )
 
+// print the strings
 func (category CategoryID) String() string {
 	switch category {
 	case CategoryID1:
--- a/cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go
+++ b/cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go
@@ -36,6 +36,7 @@ type ContainerSCC struct {
 
 var (
 	requiredDropCapabilities = []string{"MKNOD", "SETUID", "SETGID", "KILL"}
+	dropAll                  = []string{"ALL"}
 	category2AddCapabilities = []string{"NET_ADMIN, NET_RAW"}
 	category3AddCapabilities = []string{"NET_ADMIN, NET_RAW, IPC_LOCK"}
 	Category1                = ContainerSCC{false,
@@ -151,7 +152,7 @@ func updateCapabilities(cut *v1.Container, containerSCC ContainerSCC) ContainerS
 		logrus.Info("sliceDropCapabilities", sliceDropCapabilities)
 
 		sort.Strings(requiredDropCapabilities)
-		if reflect.DeepEqual(sliceDropCapabilities, requiredDropCapabilities) {
+		if reflect.DeepEqual(sliceDropCapabilities, requiredDropCapabilities) || reflect.DeepEqual(sliceDropCapabilities, dropAll) {
 			containerSCC.HaveDropCapabilities = haveRequiredDrop
 		}
 
--- a/cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go
+++ b/cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go
@@ -8,7 +8,7 @@ import (
 	"github.com/test-network-function/cnf-certification-test/pkg/provider"
 	"github.com/test-network-function/cnf-certification-test/pkg/stringhelper"
 	"github.com/test-network-function/cnf-certification-test/pkg/tnf"
-	v1 "k8s.io/api/core/v1"
+	corev1 "k8s.io/api/core/v1"
 )
 
 const (
@@ -30,7 +30,7 @@ type ContainerSCC struct {
 	FsGroup                bool
 	SeLinuxContext         bool
 	Capabilities           string
-	HaveDropCapabilities   string
+	HaveDropCapabilities   bool
 	AllVolumeAllowed       bool
 }
 
@@ -52,10 +52,10 @@ var (
 		true,
 		true,
 		category1,
-		haveRequiredDrop,
+		true,
 		true}
 
-	Category2 = ContainerSCC{false,
+	Category1NoUID0 = ContainerSCC{false,
 		false,
 		false,
 		false,
@@ -68,7 +68,23 @@ var (
 		true,
 		true,
 		category1,
-		haveRequiredDrop,
+		true,
+		true}
+
+	Category2 = ContainerSCC{false,
+		false,
+		false,
+		false,
+		false,
+		true,
+		false,
+		true,
+		false,
+		true,
+		true,
+		true,
+		"category2",
+		true,
 		true}
 
 	Category3 = ContainerSCC{false,
@@ -83,28 +99,12 @@ var (
 		true,
 		true,
 		true,
-		"category2",
-		haveRequiredDrop,
-		true}
-
-	Category4 = ContainerSCC{false,
-		false,
-		false,
-		false,
-		false,
-		true,
-		false,
-		true,
-		false,
-		true,
-		true,
-		true,
 		"category3",
-		haveRequiredDrop,
+		true,
 		true}
 )
 
-func GetContainerSCC(cut *v1.Container, containerSCC ContainerSCC) ContainerSCC {
+func GetContainerSCC(cut *corev1.Container, containerSCC ContainerSCC) ContainerSCC {
 	containerSCC.HostPorts = false
 	for _, aPort := range cut.Ports {
 		if aPort.HostPort != 0 {
@@ -140,8 +140,8 @@ func GetContainerSCC(cut *v1.Container, containerSCC ContainerSCC) ContainerSCC 
 	return containerSCC
 }
 
-func updateCapabilities(cut *v1.Container, containerSCC ContainerSCC) ContainerSCC {
-	containerSCC.HaveDropCapabilities = "notHaveRequiredDrop"
+func updateCapabilities(cut *corev1.Container, containerSCC ContainerSCC) ContainerSCC {
+	containerSCC.HaveDropCapabilities = false
 	if cut.SecurityContext != nil && cut.SecurityContext.Capabilities != nil {
 		var sliceDropCapabilities []string
 		for _, ncc := range cut.SecurityContext.Capabilities.Drop {
@@ -153,14 +153,14 @@ func updateCapabilities(cut *v1.Container, containerSCC ContainerSCC) ContainerS
 
 		sort.Strings(requiredDropCapabilities)
 		if reflect.DeepEqual(sliceDropCapabilities, requiredDropCapabilities) || reflect.DeepEqual(sliceDropCapabilities, dropAll) {
-			containerSCC.HaveDropCapabilities = haveRequiredDrop
+			containerSCC.HaveDropCapabilities = true
 		}
 
-		if checkContainCateegory(cut.SecurityContext.Capabilities.Add, category2AddCapabilities) {
+		if checkContainCategory(cut.SecurityContext.Capabilities.Add, category2AddCapabilities) {
 			logrus.Info("category is category2")
 			containerSCC.Capabilities = "category2"
 		} else {
-			if checkContainCateegory(cut.SecurityContext.Capabilities.Add, category3AddCapabilities) {
+			if checkContainCategory(cut.SecurityContext.Capabilities.Add, category3AddCapabilities) {
 				containerSCC.Capabilities = "category3"
 			} else {
 				if len(cut.SecurityContext.Capabilities.Add) > 0 {
@@ -177,7 +177,7 @@ func updateCapabilities(cut *v1.Container, containerSCC ContainerSCC) ContainerS
 	return containerSCC
 }
 
-func AllVolumeAllowed(volumes []v1.Volume) bool {
+func AllVolumeAllowed(volumes []corev1.Volume) bool {
 	countVolume := 0
 	for j := 0; j < len(volumes); j++ {
 		if volumes[j].ConfigMap != nil {
@@ -202,40 +202,78 @@ func AllVolumeAllowed(volumes []v1.Volume) bool {
 	return countVolume == len(volumes)
 }
 
-func CheckCategory(containers []v1.Container, containerSCC ContainerSCC) []string {
-	var badCcontainer []string
+type PodListcategory struct {
+	Containername string
+	Podname       string
+	NameSpace     string
+	Category      string
+}
+
+//nolint:funlen
+func CheckCategory(containers []corev1.Container, containerSCC ContainerSCC, podName, nameSpace string) []PodListcategory {
+	var ContainerList []PodListcategory
+	var categoryinfo PodListcategory
 	for j := 0; j < len(containers); j++ {
 		cut := &(containers[j])
 		percontainerSCC := GetContainerSCC(cut, containerSCC)
 		tnf.ClaimFilePrintf("percontainerSCC is ", percontainerSCC)
-
 		// after building the containerSCC need to check to which category it is
 		switch percontainerSCC {
 		case Category1:
-			logrus.Info("is ok")
+			categoryinfo = PodListcategory{
+				Containername: cut.Name,
+				Podname:       podName,
+				NameSpace:     nameSpace,
+				Category:      "Category1",
+			}
+			logrus.Info("Category1")
+		case Category1NoUID0:
+			categoryinfo = PodListcategory{
+				Containername: cut.Name,
+				Podname:       podName,
+				NameSpace:     nameSpace,
+				Category:      "Category1-no-uid0",
+			}
+			logrus.Info("its Category1-no-uid0")
 		case Category2:
-			logrus.Info("its Category 1-no-uid0")
-		case Category3:
-			badCcontainer = append(badCcontainer, cut.Name)
+			categoryinfo = PodListcategory{
+				Containername: cut.Name,
+				Podname:       podName,
+				NameSpace:     nameSpace,
+				Category:      "Category2",
+			}
 			logrus.Info("its Category2")
-		case Category4:
-			badCcontainer = append(badCcontainer, cut.Name)
+		case Category3:
+			categoryinfo = PodListcategory{
+				Containername: cut.Name,
+				Podname:       podName,
+				NameSpace:     nameSpace,
+				Category:      "Category3",
+			}
 			logrus.Info("its Category3")
 		default:
+			categoryinfo = PodListcategory{
+				Containername: cut.Name,
+				Podname:       podName,
+				NameSpace:     nameSpace,
+				Category:      "OtherType",
+			}
+
 			logrus.Info("no one from the categories")
-			badCcontainer = append(badCcontainer, cut.Name)
 		}
+		ContainerList = append(ContainerList, categoryinfo)
 	}
-	return badCcontainer
+	return ContainerList
 }
-func checkContainCateegory(addCapability []v1.Capability, categoryAddCapabilities []string) bool {
+
+func checkContainCategory(addCapability []corev1.Capability, categoryAddCapabilities []string) bool {
 	for _, ncc := range addCapability {
 		return stringhelper.StringInSlice(categoryAddCapabilities, string(ncc), true)
 	}
 	return len(addCapability) > 0
 }
 
-func CheckPod(pod *provider.Pod) []string {
+func CheckPod(pod *provider.Pod) []PodListcategory {
 	var containerSCC ContainerSCC
 	containerSCC.HostIPC = pod.Spec.HostIPC
 	containerSCC.HostNetwork = pod.Spec.HostNetwork
@@ -263,5 +301,5 @@ func CheckPod(pod *provider.Pod) []string {
 		logrus.Info("FsGroupis false")
 		containerSCC.FsGroup = false
 	}
-	return CheckCategory(pod.Spec.Containers, containerSCC)
+	return CheckCategory(pod.Spec.Containers, containerSCC, pod.Name, pod.Namespace)
 }
--- a/cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go
+++ b/cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go
@@ -21,8 +21,8 @@ type OkNok int
 
 const (
 	OKNOK = iota
-	OK
-	NOK
+	NOK   // 0
+	OK    // 1
 )
 
 const (
@@ -186,7 +186,7 @@ func updateCapabilities(cut *provider.Container, containerSCC *ContainerSCC) {
 				containerSCC.Capabilities = "category3"
 			} else {
 				if len(cut.SecurityContext.Capabilities.Add) > 0 {
-					containerSCC.Capabilities = "category5"
+					containerSCC.Capabilities = "category4"
 				} else {
 					logrus.Info("category is category1")
 					containerSCC.Capabilities = category1
@@ -361,83 +361,95 @@ func CheckPod(pod *provider.Pod) []PodListcategory {
 func compareCategory(refCategory, containerSCC ContainerSCC, id CategoryID) bool {
 	result := true
 	tnf.ClaimFilePrintf("Testing if pod belongs to category %s", &id)
-	if refCategory.AllVolumeAllowed < containerSCC.AllVolumeAllowed {
-		tnf.ClaimFilePrintf("AllVolumeAllowed = %s but expected <= %s -  NOK", containerSCC.AllVolumeAllowed, refCategory.AllVolumeAllowed)
-		result = false
-	} else {
+	if refCategory.AllVolumeAllowed == containerSCC.AllVolumeAllowed {
 		tnf.ClaimFilePrintf("AllVolumeAllowed = %s - OK", containerSCC.AllVolumeAllowed)
-	}
-	if refCategory.FsGroup < containerSCC.FsGroup {
-		tnf.ClaimFilePrintf("FsGroup = %s but expected <= %s - NOK", containerSCC.FsGroup, refCategory.FsGroup)
-		result = false
 	} else {
-		tnf.ClaimFilePrintf("FsGroup = %s - OK", containerSCC.FsGroup)
+		result = false
+		tnf.ClaimFilePrintf("AllVolumeAllowed = %s but expected >=<=%s -  NOK", containerSCC.AllVolumeAllowed, refCategory.AllVolumeAllowed)
 	}
-	if refCategory.HaveDropCapabilities < containerSCC.HaveDropCapabilities {
-		tnf.ClaimFilePrintf("HaveDropCapabilities = %s but expected <= %s - NOK", containerSCC.HaveDropCapabilities, refCategory.HaveDropCapabilities)
+	if refCategory.RunAsUser == containerSCC.RunAsUser {
+		tnf.ClaimFilePrintf("RunAsUser = %s - OK", containerSCC.RunAsUser)
+	} else {
+		tnf.ClaimFilePrintf("RunAsUser = %s but expected  %s - NOK", containerSCC.RunAsUser, refCategory.RunAsUser)
+		result = false
+	}
+	if refCategory.RunAsNonRoot >= containerSCC.RunAsNonRoot {
+		tnf.ClaimFilePrintf("RunAsNonRoot = %s - OK", containerSCC.RunAsNonRoot)
+	} else {
+		tnf.ClaimFilePrintf("RunAsNonRoot = %s but expected  %s - NOK", containerSCC.RunAsNonRoot, refCategory.RunAsNonRoot)
+		result = false
+	}
+	if refCategory.FsGroup == containerSCC.FsGroup {
+		tnf.ClaimFilePrintf("FsGroup = %s - OK", containerSCC.FsGroup)
+	} else {
+		tnf.ClaimFilePrintf("FsGroup = %s but expected  %s - NOK", containerSCC.FsGroup, refCategory.FsGroup)
+		result = false
+	}
+	if refCategory.HaveDropCapabilities == containerSCC.HaveDropCapabilities {
+		tnf.ClaimFilePrintf("DropCapabilities list - OK")
+	} else {
+		tnf.ClaimFilePrintf("HaveDropCapabilities = %s but expected  %s - NOK", containerSCC.HaveDropCapabilities, refCategory.HaveDropCapabilities)
 		tnf.ClaimFilePrintf("its didnt have all the required (MKNOD, SETUID, SETGID, KILL)/(ALL) drop value ")
 		result = false
-	} else {
-		tnf.ClaimFilePrintf("DropCapabilities list - OK")
 	}
-	if refCategory.HostDirVolumePlugin < containerSCC.HostDirVolumePlugin {
-		tnf.ClaimFilePrintf("HostDirVolumePlugin = %s but expected <= %s - NOK", containerSCC.HostDirVolumePlugin, refCategory.HostDirVolumePlugin)
-		result = false
-	} else {
+	if refCategory.HostDirVolumePlugin == containerSCC.HostDirVolumePlugin {
 		tnf.ClaimFilePrintf("HostDirVolumePlugin = %s - OK", containerSCC.HostDirVolumePlugin)
+	} else {
+		tnf.ClaimFilePrintf("HostDirVolumePlugin = %s but expected  %s - NOK", containerSCC.HostDirVolumePlugin, refCategory.HostDirVolumePlugin)
+		result = false
 	}
-	if refCategory.HostIPC < containerSCC.HostIPC {
+	if refCategory.HostIPC >= containerSCC.HostIPC {
+		tnf.ClaimFilePrintf("HostIPC = %s - OK", containerSCC.HostIPC)
+	} else {
 		result = false
 		tnf.ClaimFilePrintf("HostIPC = %s but expected <= %s - NOK", containerSCC.HostIPC, refCategory.HostIPC)
-	} else {
-		tnf.ClaimFilePrintf("HostIPC = %s - OK", containerSCC.HostIPC)
 	}
-	if refCategory.HostNetwork < containerSCC.HostNetwork {
+	if refCategory.HostNetwork >= containerSCC.HostNetwork {
+		tnf.ClaimFilePrintf("HostNetwork = %s - OK", containerSCC.HostNetwork)
+	} else {
 		result = false
 		tnf.ClaimFilePrintf("HostNetwork = %s but expected <= %s - NOK", containerSCC.HostNetwork, refCategory.HostNetwork)
-	} else {
-		tnf.ClaimFilePrintf("HostNetwork = %s - OK", containerSCC.HostNetwork)
 	}
-	if refCategory.HostPID < containerSCC.HostPID {
+	if refCategory.HostPID >= containerSCC.HostPID {
+		tnf.ClaimFilePrintf("HostPID = %s - OK", containerSCC.HostPID)
+	} else {
 		result = false
 		tnf.ClaimFilePrintf("HostPID = %s but expected <= %s - NOK", containerSCC.HostPID, refCategory.HostPID)
-	} else {
-		tnf.ClaimFilePrintf("HostPID = %s - OK", containerSCC.HostPID)
 	}
-	if refCategory.HostPorts < containerSCC.HostPorts {
+	if refCategory.HostPorts >= containerSCC.HostPorts {
+		tnf.ClaimFilePrintf("HostPorts = %s - OK", containerSCC.HostPorts)
+	} else {
 		result = false
 		tnf.ClaimFilePrintf("HostPorts = %s but expected <= %s - NOK", containerSCC.HostPorts, refCategory.HostPorts)
-	} else {
-		tnf.ClaimFilePrintf("HostPorts = %s - OK", containerSCC.HostPorts)
 	}
-	if refCategory.PrivilegeEscalation < containerSCC.PrivilegeEscalation {
-		//result = false
-		tnf.ClaimFilePrintf("PrivilegeEscalation = %s but expected <= %s - NOK", containerSCC.PrivilegeEscalation, refCategory.PrivilegeEscalation)
-	} else {
+	if refCategory.PrivilegeEscalation >= containerSCC.PrivilegeEscalation {
 		tnf.ClaimFilePrintf("HostNetwork = %s - OK", containerSCC.HostNetwork)
+	} else {
+		result = false
+		tnf.ClaimFilePrintf("PrivilegeEscalation = %s but expected <= %s - NOK", containerSCC.PrivilegeEscalation, refCategory.PrivilegeEscalation)
 	}
-	if refCategory.PrivilegedContainer < containerSCC.PrivilegedContainer {
+	if refCategory.PrivilegedContainer >= containerSCC.PrivilegedContainer {
+		tnf.ClaimFilePrintf("PrivilegedContainer = %s - OK", containerSCC.PrivilegedContainer)
+	} else {
 		result = false
 		tnf.ClaimFilePrintf("PrivilegedContainer = %s but expected <= %s - NOK", containerSCC.PrivilegedContainer, refCategory.PrivilegedContainer)
-	} else {
-		tnf.ClaimFilePrintf("PrivilegedContainer = %s - OK", containerSCC.PrivilegedContainer)
 	}
-	if refCategory.ReadOnlyRootFilesystem < containerSCC.ReadOnlyRootFilesystem {
+	if refCategory.ReadOnlyRootFilesystem >= containerSCC.ReadOnlyRootFilesystem {
+		tnf.ClaimFilePrintf("ReadOnlyRootFilesystem = %s - OK", containerSCC.ReadOnlyRootFilesystem)
+	} else {
 		result = false
 		tnf.ClaimFilePrintf("ReadOnlyRootFilesystem = %s but expected <= %s - NOK", containerSCC.ReadOnlyRootFilesystem, refCategory.ReadOnlyRootFilesystem)
-	} else {
-		tnf.ClaimFilePrintf("ReadOnlyRootFilesystem = %s - OK", containerSCC.ReadOnlyRootFilesystem)
 	}
-	if refCategory.SeLinuxContext < containerSCC.SeLinuxContext {
-		result = false
-		tnf.ClaimFilePrintf("SeLinuxContext = %s but expected <= %s - NOK", containerSCC.SeLinuxContext, refCategory.SeLinuxContext)
-		tnf.ClaimFilePrintf("SeLinuxContext expected to be non nil")
-	} else {
+	if refCategory.SeLinuxContext == containerSCC.SeLinuxContext {
 		tnf.ClaimFilePrintf("SeLinuxContext is not nil - OK")
-	}
-	if refCategory.Capabilities < containerSCC.Capabilities {
+	} else {
 		result = false
-		tnf.ClaimFilePrintf("Capabilities = %s but expected <= %s - NOK", containerSCC.Capabilities, refCategory.Capabilities)
+		tnf.ClaimFilePrintf("SeLinuxContext = %s but expected  %s - NOK", containerSCC.SeLinuxContext, refCategory.SeLinuxContext)
+		tnf.ClaimFilePrintf("SeLinuxContext expected to be non nil")
+	}
+	if refCategory.Capabilities != containerSCC.Capabilities {
+		result = false
+		tnf.ClaimFilePrintf("Capabilities = %s but expected  %s - NOK", containerSCC.Capabilities, refCategory.Capabilities)
 	}
 	return result
 }
--- a/cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go
+++ b/cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go
@@ -337,7 +337,9 @@ func checkContainerCategory(containers []corev1.Container, containerSCC Containe
 
 func checkContainCategory(addCapability []corev1.Capability, categoryAddCapabilities []string) bool {
 	for _, ncc := range addCapability {
-		return stringhelper.StringInSlice(categoryAddCapabilities, string(ncc), false)
+		if !stringhelper.StringInSlice(categoryAddCapabilities, string(ncc), true) {
+			return false
+		}
 	}
 	return len(addCapability) > 0
 }
--- a/cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go
+++ b/cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go
@@ -45,7 +45,7 @@ var (
 		false,
 		true,
 		true,
-		"",
+		"category1,2",
 		true,
 		true}
 
@@ -61,7 +61,7 @@ var (
 		true,
 		true,
 		true,
-		"",
+		"category1,2",
 		true,
 		true}
 
@@ -139,6 +139,7 @@ func updateCapabilities(cut *v1.Container, containerSCC ContainerSCC) ContainerS
 		for _, ncc := range cut.SecurityContext.Capabilities.Drop {
 			sliceDropCapabilities = append(sliceDropCapabilities, string(ncc))
 		}
+		logrus.Info("cut.SecurityContext.Capabilities.Drop", cut.SecurityContext.Capabilities.Drop)
 		sort.Strings(sliceDropCapabilities)
 		sort.Strings(requiredDropCapabilities)
 		containerSCC.HaveDropCapabilities = reflect.DeepEqual(sliceDropCapabilities, requiredDropCapabilities)
@@ -152,13 +153,13 @@ func updateCapabilities(cut *v1.Container, containerSCC ContainerSCC) ContainerS
 				if len(cut.SecurityContext.Capabilities.Add) > 0 {
 					containerSCC.Capabilities = "category5"
 				} else {
-					logrus.Info("category is category")
-					containerSCC.Capabilities = ""
+					logrus.Info("category is category1,2")
+					containerSCC.Capabilities = "category1,2"
 				}
 			}
 		}
 	} else {
-		containerSCC.Capabilities = ""
+		containerSCC.Capabilities = "category1,2"
 	}
 	return containerSCC
 }
--- a/cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go
+++ b/cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go
@@ -11,6 +11,11 @@ import (
 	v1 "k8s.io/api/core/v1"
 )
 
+const (
+	category1 = "category1"
+	other     = "other"
+)
+
 type ContainerSCC struct {
 	HostDirVolumePlugin    bool
 	HostIPC                bool
@@ -25,15 +30,16 @@ type ContainerSCC struct {
 	FsGroup                bool
 	SeLinuxContext         bool
 	Capabilities           string
-	HaveDropCapabilities   bool
+	HaveDropCapabilities   string
 	AllVolumeAllowed       bool
 }
 
 var (
-	requiredDropCapabilities = []string{"KILL", "MKNOD", "SETUID", "SETGID"}
-	category3AddCapabilities = []string{"NET_ADMIN, NET_RAW"}
-	category4AddCapabilities = []string{"NET_ADMIN, NET_RAW, IPC_LOCK"}
-	Category1                = ContainerSCC{false,
+	requiredDropCapabilities        = []string{"KILL", "MKNOD", "SETUID", "SETGID"}
+	requiredDropCapabilitiesForCat1 = []string{"KILL", "MKNOD", "SETUID", "SETGID", "NET_RAW"}
+	category2AddCapabilities        = []string{"NET_ADMIN, NET_RAW"}
+	category3AddCapabilities        = []string{"NET_ADMIN, NET_RAW, IPC_LOCK"}
+	Category1                       = ContainerSCC{false,
 		false,
 		false,
 		false,
@@ -45,8 +51,8 @@ var (
 		true,
 		true,
 		true,
-		"category1,2",
-		true,
+		category1,
+		category1,
 		true}
 
 	Category2 = ContainerSCC{false,
@@ -58,11 +64,11 @@ var (
 		false,
 		false,
 		false,
+		false,
 		true,
 		true,
-		true,
-		"category1,2",
-		true,
+		category1,
+		other,
 		true}
 
 	Category3 = ContainerSCC{false,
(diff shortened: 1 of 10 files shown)


cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go
```diff
@@ -0,0 +1,103 @@
...
+	const istioProxyContainerUID = 1337
+	containerSCC.HostPorts = false
+	for _, aPort := range cut.Ports {
+		if aPort.HostPort != 0 {
+			containerSCC.HostPorts = true
+			break
+		}
+	}
+	if cut.SecurityContext != nil && cut.SecurityContext.AllowPrivilegeEscalation != nil {
+		if *(cut.SecurityContext.AllowPrivilegeEscalation) {
```
rdavid, line 80: No need for condition, no need for `()`, `.` has higher precedence than `*`.
```
containerSCC.PrivilegedContainer = *cut.SecurityContext.AllowPrivilegeEscalation
```

cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go
```diff
@@ -0,0 +1,103 @@
...
+			containerSCC.PrivilegedContainer = true
+		} else {
+			containerSCC.PrivilegedContainer = false
+		}
+	}
+	if cut.SecurityContext != nil && cut.SecurityContext.Capabilities != nil {
+		containerSCC.Capabilities = cut.SecurityContext.Capabilities.String()
+	} else {
+		containerSCC.Capabilities = ""
```
rdavid, line 89: In some cases you set explicit default values, in some you don't. Set it in all cases. Or consider not to use same variable for data store for all runs, you could pass it as constant and create copy, then set only other than defaults and return the new data store. Empty strings, false booleans, zero integers are defaults in Go.

aabughosh:
> 

hi @ramperher  https://docs.google.com/document/d/1MbzJsTMJwTOcYdSeI1ulmrjjYAgKGBKh/edit?copiedFromTrash# 
all the information is written here on page 51
there is 4 categories that the container  will passed if it from one of them

cnf-certification-test/accesscontrol/suite.go
```diff
@@ -611,3 +618,37 @@ func Test1337UIDs(env *provider.TestEnvironment) {
 
 	testhelper.AddTestResultLog("Non-compliant", badPods, tnf.ClaimFilePrintf, ginkgo.Fail)
 }
+
+func testContainerSCC(env *provider.TestEnvironment) {
+	var containerSCC securitycontextcontainer.ContainerSCC
+	const istioProxyContainerUID = 1337
+	var badCut []string
```
sebrandon1, line 625:
```suggestion
	var badContainers []string
```

cnf-certification-test/identifiers/identifiers.go
```diff
@@ -272,6 +272,12 @@ var (
 		Url:     formTestURL(common.AccessControlTestKey, "security-context-non-root-user-check"),
 		Version: VersionOne,
 	}
+	// TestSecContextIdentifier tests that pods or containers are not running with root permissions
+	TestSecContextIdentifier = claim.Identifier{
```
sebrandon1, line 276:
```suggestion
	TestSccContextIdentifier = claim.Identifier{
```

cnf-certification-test/accesscontrol/suite.go
```diff
@@ -611,3 +618,37 @@ func Test1337UIDs(env *provider.TestEnvironment) {
...
+				logrus.Info("is ok")
+			case securitycontextcontainer.Catagory3:
+				logrus.Info("is ok")
+			case securitycontextcontainer.Catagory4:
+				logrus.Info("is ok")
+			default:
+				badCut = append(badCut, cut.Name)
+			}
+		}
+	}
```
sebrandon1, line 652: I think this whole section should be in its own func and have unit tests added.

cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go
```diff
@@ -0,0 +1,103 @@
...
+	HostIPC                bool
+	HostNetwork            bool
+	HostPID                bool
+	HostPorts              bool
+	PrivilegeEscalation    bool
+	PrivilegedContainer    bool
+	RunAsUser              bool
+	ReadOnlyRootFilesystem bool
+	RunAsNonRoot           bool
+	Capabilities           string
```
edcdavid, line 18: Add API references in the code documentation:
https://docs.openshift.com/container-platform/4.11/authentication/managing-security-context-constraints.html
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.25/
https://andreaskaris.github.io/blog/openshift/scc/
After reading some more about the runasgroup, I think that we should check that it is set to something. It is not specified in the example SCC probably because it is not known in advance, but the mustRunAs seems to suggest that the intent is to have it set (even if the range is not set)

cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go
```diff
@@ -0,0 +1,103 @@
...
+	containerSCC.HostPorts = false
+	for _, aPort := range cut.Ports {
+		if aPort.HostPort != 0 {
+			containerSCC.HostPorts = true
+			break
+		}
+	}
+	if cut.SecurityContext != nil && cut.SecurityContext.AllowPrivilegeEscalation != nil {
+		if *(cut.SecurityContext.AllowPrivilegeEscalation) {
+			containerSCC.PrivilegedContainer = true
```
edcdavid, line 81: Shoudn't we have two test here? one for PrivilegeEscalation and one for PrivilegedContainer. PrivilegedContainer right now is unused.

cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go
```diff
@@ -0,0 +1,103 @@
...
+		} else {
+			containerSCC.PrivilegedContainer = false
+		}
+	}
+	if cut.SecurityContext != nil && cut.SecurityContext.Capabilities != nil {
+		containerSCC.Capabilities = cut.SecurityContext.Capabilities.String()
+	} else {
+		containerSCC.Capabilities = ""
+	}
+	if cut.SecurityContext != nil && cut.SecurityContext.RunAsUser != nil && *cut.SecurityContext.RunAsUser == int64(istioProxyContainerUID) {
```
edcdavid, line 91: uid 1337 is valid only for service mesh (istio). In the regular case we should check that it is not null. We can't check for a specific value since we don't know it in advance

cnf-certification-test/accesscontrol/suite.go
```diff
@@ -611,3 +618,37 @@ func Test1337UIDs(env *provider.TestEnvironment) {
 
 	testhelper.AddTestResultLog("Non-compliant", badPods, tnf.ClaimFilePrintf, ginkgo.Fail)
 }
+
+func testContainerSCC(env *provider.TestEnvironment) {
+	var containerSCC securitycontextcontainer.ContainerSCC
+	const istioProxyContainerUID = 1337
+	var badCut []string
+	for _, pod := range env.Pods {
+		containerSCC.HostIPC = pod.Spec.HostIPC
```
edcdavid, line 627: The container level parameters should take precedence over the pod level ones, only if they are defined. looks like the GetContainerSCC is always overwriting the values collected at the pod level. To track whether a parameter is set of not, maybe we could use pointers instead of values. So that it it is null, the param is not se, for example:
```
	HostDirVolumePlugin    *bool
	HostIPC                *bool
	HostNetwork            *bool
	HostPID                *bool
	HostPorts              *bool
	PrivilegeEscalation    *bool
	PrivilegedContainer    *bool
	RunAsUser              *bool
	ReadOnlyRootFilesystem *bool
	RunAsNonRoot           *bool 
	fsGroup               *bool 
	seLinuxContext:       *bool 
	volumes               string[]
```

Other values missing from pod level:
- runAsGroup
- runAsNonRoot
- selinuxOptions

ramperher:
> > 
> 
> hi @ramperher https://docs.google.com/document/d/1MbzJsTMJwTOcYdSeI1ulmrjjYAgKGBKh/edit?copiedFromTrash# all the information is written here on page 51 there is 4 categories that the container will passed if it from one of them

I don't have access to it, unfortunately, I don't have permissions :( can you give me access to it?

Anyway, this only answers my first question. I have still doubts with these other ones:

> * If no security context is set in the pods under test, is this test expected to fail?
>   
>   * I'm saying this because, in [our case](https://github.com/redhat-cip/dci-openshift-app-agent/blob/master/samples/tnf_test_example/hooks/templates/test_pods.yml.j2), which mimics your [local pod under test](https://github.com/test-network-function/cnf-certification-test-partner/blob/main/test-target/local-pod-under-test.yaml), is not using any security context and it is failing.
> * CATALOG.md said that it is suggested to "Change the pod and containers "runAsUser" uid to something other than root(0)". However, according to this, if this test fails, it is because we are running as root. And in our case, by checking the DCI job, you can see that "access-control-security-context-non-root-user-check" test is passing (so we are not using root user). Can you clarify this?
> 
> Finally, after saying this and taking into account this, can you provide an example of a pod passing this test suite, to replicate it on our lab? (in case we need to modify something, I mean).

Can we clarify this before continuing?

ramperher:
> > > 
> > 
> > 
> > hi @ramperher https://docs.google.com/document/d/1MbzJsTMJwTOcYdSeI1ulmrjjYAgKGBKh/edit?copiedFromTrash# all the information is written here on page 51 there is 4 categories that the container will passed if it from one of them
> 
> I don't have access to it, unfortunately, I don't have permissions :( can you give me access to it?
> 
> Anyway, this only answers my first question. I have still doubts with these other ones:
> 
> > * If no security context is set in the pods under test, is this test expected to fail?
> >   
> >   * I'm saying this because, in [our case](https://github.com/redhat-cip/dci-openshift-app-agent/blob/master/samples/tnf_test_example/hooks/templates/test_pods.yml.j2), which mimics your [local pod under test](https://github.com/test-network-function/cnf-certification-test-partner/blob/main/test-target/local-pod-under-test.yaml), is not using any security context and it is failing.
> > * CATALOG.md said that it is suggested to "Change the pod and containers "runAsUser" uid to something other than root(0)". However, according to this, if this test fails, it is because we are running as root. And in our case, by checking the DCI job, you can see that "access-control-security-context-non-root-user-check" test is passing (so we are not using root user). Can you clarify this?
> > 
> > Finally, after saying this and taking into account this, can you provide an example of a pod passing this test suite, to replicate it on our lab? (in case we need to modify something, I mean).
> 
> Can we clarify this before continuing?

I'll try by using anyuid SCC by default in all our testing pods, to see if this makes this new test to pass: https://softwarefactory-project.io/r/c/dci-openshift-app-agent/+/26502. I'll do some rechecks in the DCI jobs.

cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go
```diff
@@ -0,0 +1,222 @@
...
+	Capabilities           string
+	HaveDropCapabilities   bool
+	AllVolumeAllowed       bool
+}
+
+var (
+	requiredDropCapabilities = []string{"KILL", "MKNOD", "SETUID", "SETGID"}
+	category3AddCapabilities = []string{"NET_ADMIN, NET_RAW"}
+	category4AddCapabilities = []string{"NET_ADMIN, NET_RAW, IPC_LOCK"}
+	Allowvolumes             = []string{"configMap, downwardAPI, emptyDir,persistentVolumeClaim,projected,secret"}
```
edcdavid, line 33: This seems to be unused

cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go
```diff
@@ -0,0 +1,222 @@
...
+		var sliceDropCapabilities []string
+		for _, ncc := range cut.SecurityContext.Capabilities.Drop {
+			sliceDropCapabilities = append(sliceDropCapabilities, string(ncc))
+		}
+		sort.Strings(sliceDropCapabilities)
+		sort.Strings(requiredDropCapabilities)
+		containerSCC.HaveDropCapabilities = reflect.DeepEqual(sliceDropCapabilities, requiredDropCapabilities)
+		contain := true
+
+		for _, ncc := range cut.SecurityContext.Capabilities.Add {
```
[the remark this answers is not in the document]
aabughosh: this failed on the make lint so i did the switch case

cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go
```diff
@@ -0,0 +1,221 @@
...
+			if !contains(category3AddCapabilities, string(ncc)) {
+				contain = false
+			}
+		}
+		if contain {
+			containerSCC.Capabilities = "catagory3"
```
edcdavid, line 148: catagory -> category

cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go
```diff
@@ -0,0 +1,221 @@
...
+			contain = true
+			for _, ncc := range cut.SecurityContext.Capabilities.Add {
+				if !contains(category4AddCapabilities, string(ncc)) {
+					contain = false
+				}
+			}
+			if contain {
+				containerSCC.Capabilities = "catagory4"
+			} else {
+				containerSCC.Capabilities = "catagory5"
```
edcdavid, line 159: catagory -> category

cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go
```diff
@@ -0,0 +1,221 @@
...
+}
+
+func Checkcategory(containers []v1.Container, containerSCC ContainerSCC) []string {
+	var badCcontainer []string
+	for j := 0; j < len(containers); j++ {
+		cut := &(containers[j])
+		percontainerSCC := GetContainerSCC(cut, containerSCC)
+		// after building the containerSCC need to check to which category it is
+		switch percontainerSCC {
+		case Category1:
```
edcdavid, line 208: When doing the comparisons, we should not do a simple "exact" comparison. Maybe we can have a special equal function that would match the scc categories. We should match categories from lowest to highest. If there is not exact match but the container under test parameter needs less privileges than the SCC, it should match. For instance a category3 SCC's compliant container does not need to have privilegedEscalation to be true. If it is false it is also ok.It might need to be SCC level 3 because of other properties, such as capabilities.

cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go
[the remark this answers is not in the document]
edcdavid: not sure about what you mean here. This is just a pseudo code for a proposal. You would still need to write a new "checkCapabilities" function

ramperher:
Hi @aabughosh , if you check the last DCI job (https://www.distributed-ci.io/jobs/bf0b4f58-332b-4aa2-83ab-92c66776fe8e/jobStates?sort=date), you will see that it is not passing because this test is still failing.

Right now, what I am doing is to use the Category 1 SCC for all the pods under test. I've achieved to reduce the number of non-compliant pods from 4 (all the pods) to 1, but the test is still failing. Do you know what may be happening? I am applying exactly the same SCC proposed by Verizon in their docs, so it should not fail.

cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go
```diff
@@ -0,0 +1,240 @@
...
+					containerSCC.Capabilities = ""
+				}
+			}
+		}
+	} else {
+		containerSCC.Capabilities = ""
+	}
+	return containerSCC
+}
+func contains(elems []string, v string) bool {
```
sebrandon1, line 159: Use `stringhelper.StringInSlice` instead.

cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go
```diff
@@ -0,0 +1,240 @@
...
+			countVolume++
+		}
+		if volumes[j].Secret != nil {
+			countVolume++
+		}
+	}
+	return countVolume == len(volumes)
+}
+
+func Checkcategory(containers []v1.Container, containerSCC ContainerSCC) []string {
```
sebrandon1, line 193: Adjust the func name for camel case.

```suggestion
func CheckCategory(containers []v1.Container, containerSCC ContainerSCC) []string {
```

cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go
```diff
@@ -0,0 +1,240 @@
...
+			badCcontainer = append(badCcontainer, cut.Name)
+		}
+	}
+	return badCcontainer
+}
+func checkContainCateegory(addCapability []v1.Capability, categoryAddCapabilities []string) bool {
+	for _, ncc := range addCapability {
+		if !contains(categoryAddCapabilities, string(ncc)) {
+			return false
+		}
```
sebrandon1, line 218:
```suggestion
                return stringhelper.StringInSlice(categoryAddCapabilities, string(ncc))
```

cnf-certification-test/accesscontrol/suite.go
```diff
@@ -611,3 +618,14 @@ func Test1337UIDs(env *provider.TestEnvironment) {
...
 	testhelper.AddTestResultLog("Non-compliant", badPods, tnf.ClaimFilePrintf, ginkgo.Fail)
 }
+
+// a test for security context that are allowed from the documentation of the cnf
+// an allowed one will pass the test
+func testContainerSCC(env *provider.TestEnvironment) {
+	var badContainer []string
+	for _, pod := range env.Pods {
+		containersBad := securitycontextcontainer.CheckPod(pod)
+		badContainer = append(badContainer, containersBad...)
```
sebrandon1, line 628:
```suggestion
		badContainer = append(badContainer, securitycontextcontainer.CheckPod(pod)...)
```

cnf-certification-test/identifiers/remediation.go
```diff
@@ -26,6 +26,8 @@ const (
 
 	SecConNonRootUserRemediation = `Change the pod and containers "runAsUser" uid to something other than root(0)`
 
+	SecConRemediation = `Change the secutiry contesxt to be one of the 4 that are allowd on the documetation section 4.5`
```
sebrandon1, line 29:
```suggestion
	SecConRemediation = `Change the security context to be one of the 4 that are allowed on the documentation section 4.5`
```

sebrandon1 (changes requested):
Some more comments.  Thanks for the unit test!

cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go
```diff
@@ -0,0 +1,240 @@
...
+	}
+	containerSCC = updateCapabilities(cut, containerSCC)
+	if cut.SecurityContext != nil && cut.SecurityContext.AllowPrivilegeEscalation != nil {
+		containerSCC.PrivilegeEscalation = true
+	}
+	if cut.SecurityContext != nil && cut.SecurityContext.Privileged != nil {
+		containerSCC.PrivilegedContainer = *(cut.SecurityContext.Privileged)
+	}
+
```
edcdavid, line 113: extra space

cnf-certification-test/accesscontrol/suite.go
```diff
@@ -611,3 +618,14 @@ func Test1337UIDs(env *provider.TestEnvironment) {
 
 	testhelper.AddTestResultLog("Non-compliant", badPods, tnf.ClaimFilePrintf, ginkgo.Fail)
 }
+
+// a test for security context that are allowed from the documentation of the cnf
+// an allowed one will pass the test
+func testContainerSCC(env *provider.TestEnvironment) {
+	var badContainer []string
+	for _, pod := range env.Pods {
+		containersBad := securitycontextcontainer.CheckPod(pod)
```
edcdavid, line 627: securitycontextcontainer.CheckPod(pod) should return a full container name including namespace and pod name, not just the container name as it might be the duplicated.

edcdavid: This is returning the pods that do not fit in a category, Instead we need the list of pods that do not fit SCC category1. In addition to the list of pods not in category 1, calculate the category for the whole CNF and save it in the claim.json file. Would also be nice to store the detailed SCC category for each container (including a dump of the SCC objsct for each)

ramperher:
Hi again @aabughosh, I've tried with what we were discussing yesterday offline, and it seems that it is still not working. Here you have the last DCI job I've launched: https://www.distributed-ci.io/jobs/3b7f1851-6435-4874-ba08-10a0382754bd/tests/328f927e-e29f-4473-8dec-2bb8a91c22b6. All the pods launched are marked as non-compliant.

In my case, I am using exactly the same SCC reported by VZ in the CNF Requirements Doc (CNFs that do not require advanced networking features (Category 1)):

```
      kind: SecurityContextConstraints
      apiVersion: security.openshift.io/v1
      metadata:
        name: restricted-cat-1
      users: []
      groups: []
      priority: null
      allowHostDirVolumePlugin: false
      allowHostIPC: false
      allowHostNetwork: false
      allowHostPID: false
      allowHostPorts: false
      allowPrivilegeEscalation: true
      allowPrivilegedContainer: false
      allowedCapabilities: null
      defaultAddCapabilities: null
      requiredDropCapabilities:
        - KILL
        - MKNOD
        - SETUID
        - SETGID
        - NET_RAW
      fsGroup:
        type: MustRunAs
      readOnlyRootFilesystem: false
      runAsUser:
        type: MustRunAsRange
      seLinuxContext:
        type: MustRunAs
      supplementalGroups:
        type: RunAsAny
      volumes:
        - configMap
        - downwardAPI
        - emptyDir
        - persistentVolumeClaim
        - projected
        - secret
```

I remember you told me that dropCapabilities you were using were the following: `["KILL","MKNOD","SETUID","SETGID"]`. However, VZ doc also considers `NET_RAW`. Do we have to include it or not?

Anyway, if we include it, `access-control-security-context-capabilities-check` test is failing too, complaining about `Non compliant NET_RAW capability detected in container`. So, it is correct or not to use that capability? Because it would require changes in that tnf test in case NET_RAW is compliant, else VZ document should have to be reviewed.

Having said that, I've launched a new job to see if, by removing `NET_RAW`, the result changes. And it doesn't: https://www.distributed-ci.io/jobs/1f1129f5-5139-48a2-a3f5-96a037c4a16b/tests/b1766919-5c59-409d-aa7c-2104c8f97cfd

My opinion is that, if this test is trying to validate a SCC defined by VZ and that specific SCC is not passing the test, something wrong should be happening.

Also, some more comments to say:

- I agree with @edcdavid with the improvement in the reporting. It is really difficult to troubleshoot the errors with the current info.
- If I remember correctly, you commented me that this test is only validating category 1, isn't it? Because according to the code, current change mentions that the 4 categories are checked, and if the pods are not fitting in any of the 4 categories. Can you clarify this?

cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go
```diff
@@ -0,0 +1,240 @@
...
+	RunAsNonRoot           bool
+	FsGroup                bool
+	SeLinuxContext         bool
+	Capabilities           string
+	HaveDropCapabilities   bool
+	AllVolumeAllowed       bool
+}
+
+var (
+	requiredDropCapabilities = []string{"KILL", "MKNOD", "SETUID", "SETGID"}
```
edcdavid, line 31: also reported by @ramperher , missing NET_RAW

aabughosh: yah i just realize its just on category 1

cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go
```diff
@@ -0,0 +1,252 @@
...
+		}
+	}
+	return countVolume == len(volumes)
+}
+
+func CheckCategory(containers []v1.Container, containerSCC ContainerSCC) []string {
+	var badCcontainer []string
+	for j := 0; j < len(containers); j++ {
+		cut := &(containers[j])
+		percontainerSCC := GetContainerSCC(cut, containerSCC)
```
edcdavid, line 199: maybe we can save this percontainerSCC struc in the claim file

cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go
```diff
@@ -0,0 +1,267 @@
...
+		true,
+		false,
+		true,
+		true,
+		true,
+		"category2",
+		haveRequiredDrop,
+		true}
+
+	Category4 = ContainerSCC{false,
```
edcdavid, line 90: rename the categories to match req document

cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go
```diff
@@ -0,0 +1,267 @@
...
+
+import (
+	"reflect"
+	"sort"
+
+	"github.com/sirupsen/logrus"
+	"github.com/test-network-function/cnf-certification-test/pkg/provider"
+	"github.com/test-network-function/cnf-certification-test/pkg/stringhelper"
+	"github.com/test-network-function/cnf-certification-test/pkg/tnf"
+	v1 "k8s.io/api/core/v1"
```
sebrandon1, line 11:
```suggestion
	corev1 "k8s.io/api/core/v1"
```

To match #618

cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go
```diff
@@ -0,0 +1,267 @@
...
+				}
+			}
+		}
+	} else {
+		containerSCC.Capabilities = category1
+	}
+	return containerSCC
+}
+
+func AllVolumeAllowed(volumes []v1.Volume) bool {
```
sebrandon1, line 180: We should probably have a unit test for this function.  It can be added after this merges.

cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go
```diff
@@ -0,0 +1,267 @@
...
+		case Category4:
+			badCcontainer = append(badCcontainer, cut.Name)
+			logrus.Info("its Category3")
+		default:
+			logrus.Info("no one from the categories")
+			badCcontainer = append(badCcontainer, cut.Name)
+		}
+	}
+	return badCcontainer
+}
```
sebrandon1, line 230: Add space between funcs.

cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go
```diff
@@ -0,0 +1,267 @@
...
+func checkContainCateegory(addCapability []v1.Capability, categoryAddCapabilities []string) bool {
```
sebrandon1, line 231:
```suggestion
func checkContainCategory(addCapability []v1.Capability, categoryAddCapabilities []string) bool {
```

cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go
```diff
@@ -0,0 +1,267 @@
...
+		logrus.Info("RunAsNonRoot is ", *(cut.SecurityContext.RunAsNonRoot))
+		containerSCC.RunAsNonRoot = *cut.SecurityContext.RunAsNonRoot
+	}
+	if cut.SecurityContext != nil && cut.SecurityContext.SELinuxOptions != nil {
+		containerSCC.SeLinuxContext = true
+	}
+	return containerSCC
+}
+
+func updateCapabilities(cut *v1.Container, containerSCC ContainerSCC) ContainerSCC {
```
sebrandon1, line 143: Since this func is an "update", we should probably just modify a pointer and remove the return value:

```suggestion
func updateCapabilities(cut *v1.Container, containerSCC *ContainerSCC) {
```

sebrandon1 (changes requested):
Some more comments!

cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go
```diff
@@ -0,0 +1,267 @@
...
+		}
+		if volumes[j].Secret != nil {
+			countVolume++
+		}
+	}
+	return countVolume == len(volumes)
+}
+
+func CheckCategory(containers []v1.Container, containerSCC ContainerSCC) []string {
+	var badCcontainer []string
```
sebrandon1, line 206: Is this a typo?

```suggestion
	var badContainer []string
```

edcdavid (commented):
needs rebase, not actually running SCC tests

cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go
```diff
@@ -0,0 +1,343 @@
...
+}
+
+//nolint:funlen
+func CheckCategory(containers []corev1.Container, containerSCC ContainerSCC, podName, nameSpace string) []PodListcategory {
+	var ContainerList []PodListcategory
+	var categoryinfo PodListcategory
+	for j := 0; j < len(containers); j++ {
+		cut := &(containers[j])
+		percontainerSCC := GetContainerSCC(cut, containerSCC)
+		tnf.ClaimFilePrintf("percontainerSCC is ", percontainerSCC)
```
edcdavid, line 257: log is unreadable, also print the cut to add context

cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go
```diff
@@ -0,0 +1,343 @@
...
+const (
+	Undefined CategoryID = iota
+	CategoryID1
+	CategoryID1NoUID0
+	CategoryID2
+	CategoryID3
+	CategoryID4
+)
+
+type PodListcategory struct {
```
edcdavid, line 216: logs are not readible, we should add a stringer for this as well

cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go
```diff
@@ -0,0 +1,343 @@
...
+func CheckCategory(containers []corev1.Container, containerSCC ContainerSCC, podName, nameSpace string) []PodListcategory {
+	var ContainerList []PodListcategory
+	var categoryinfo PodListcategory
+	for j := 0; j < len(containers); j++ {
+		cut := &(containers[j])
+		percontainerSCC := GetContainerSCC(cut, containerSCC)
+		tnf.ClaimFilePrintf("percontainerSCC is ", percontainerSCC)
+		// after building the containerSCC need to check to which category it is
+		switch percontainerSCC {
+		case Category1:
```
edcdavid, line 260: This is hard to debug without a log of why it was not equal. It would be nice to know what fields are mismatched

aabughosh: i added that function that is printing to the claim file for each container the mismatched

cnf-certification-test/identifiers/remediation.go
```diff
@@ -26,6 +26,8 @@ const (
 
 	SecConNonRootUserRemediation = `Change the pod and containers "runAsUser" uid to something other than root(0)`
 
+	SecConRemediation = `Change the secutiry context to be one of the 4 that are allowed on the documentation section 4.5`
```
sebrandon1, line 29:
```suggestion
	SecConRemediation = `Change the security context to be one of the 4 that are allowed on the documentation section 4.5`
```

cnf-certification-test/identifiers/exceptions.go
```diff
@@ -34,6 +34,9 @@ const (
 
 	SecConNonRootUserExceptionProcess = `If your application needs root user access, please document why your application cannot be ran as
 											non-root and supply the reasoning for exception.`
+	SecConExceptionProcess = `If the contaier had the right configuration of the allowed category from the 4 list so the test will pass the 
+	list is on page 51 on the CNF Security Context Constraints (SCC) section 4.5, 
+	Applications MUST use one of the approved Security Context Constraints profiles unless an exception (SEAT) is approved by Verizon.`
```
sebrandon1, line 39: Do we want to explicitly use `Verizon` here?  @mwlinca ?

cnf-certification-test/accesscontrol/suite.go
```diff
@@ -611,3 +618,32 @@ func Test1337UIDs(env *provider.TestEnvironment) {
...
+			}
+			if cat.Category > highLevelCat {
+				highLevelCat = cat.Category
+			}
+		}
+	}
+	tnf.ClaimFilePrintf("list of goodContainer", goodContainer)
+	tnf.ClaimFilePrintf("list of badContainers", badContainer)
+	tnf.ClaimFilePrintf("high level of category", highLevelCat)
+	tnf.ClaimFilePrintf("list of all ", listCategoryContainer)
```
sebrandon1, line 647: I think the claim file prints should contain more context.

"Containers that do not fall within SCC Category 1: ... "
"Containers that fall within SCC Category 1: ...." 

Something that is more verbose for the partner to read in the logs.

cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go
```diff
@@ -0,0 +1,398 @@
...
+		logrus.Info("RunAsNonRoot is ", *(cut.SecurityContext.RunAsNonRoot))
+		containerSCC.RunAsNonRoot = *cut.SecurityContext.RunAsNonRoot
+	}
+	if cut.SecurityContext != nil && cut.SecurityContext.SELinuxOptions != nil {
+		containerSCC.SeLinuxContext = true
+	}
+	return containerSCC
+}
+
+func updateCapabilities(cut *corev1.Container, containerSCC ContainerSCC) ContainerSCC {
```
sebrandon1, line 144: I still think that this should just adjust the pointer versus make a copy of a struct.

```suggestion
func updateCapabilities(cut *corev1.Container, containerSCC *ContainerSCC) {
```

sebrandon1 (changes requested):
Some more comments

cnf-certification-test/identifiers/exceptions.go
aabughosh: i removed that

cnf-certification-test/accesscontrol/suite.go
```diff
@@ -611,3 +618,28 @@ func Test1337UIDs(env *provider.TestEnvironment) {
...
+			} else {
+				goodContainer = append(goodContainer, cat)
+			}
+			if cat.Category > highLevelCat {
+				highLevelCat = cat.Category
+			}
+		}
+	}
+	tnf.ClaimFilePrintf("CNF category (highest container category across all containers):  %s \n", highLevelCat)
+	tnf.ClaimFilePrintf("list of goodContainer that are Category1 or CategoryNoUID0 %+v \n", goodContainer)
```
sebrandon1, line 643:
```suggestion
	tnf.ClaimFilePrintf("List of containers that are Category1 or CategoryNoUID0 %+v \n", goodContainer)
```

cnf-certification-test/accesscontrol/suite.go
```diff
@@ -611,3 +618,28 @@ func Test1337UIDs(env *provider.TestEnvironment) {
...
+	testhelper.AddTestResultLog("list of badContainers that are not from Category1 or CategoryNoUID0 - ", badContainer, tnf.ClaimFilePrintf, ginkgo.Fail)
```
sebrandon1, line 644:
```suggestion
	testhelper.AddTestResultLog("List of non-compliant containers that are not from Category1 or CategoryNoUID0 - ", badContainer, tnf.ClaimFilePrintf, ginkgo.Fail)
```

sebrandon1 (changes requested):
Two improvements to strings

cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go
```diff
@@ -0,0 +1,454 @@
...
+	HostPID                    OkNok
+	HostPorts                  OkNok
+	PrivilegeEscalation        OkNok // this can be true or false
+	PrivilegedContainer        OkNok
+	RunAsUserPresent           OkNok // thes filed that checking if the value is present
+	ReadOnlyRootFilesystem     OkNok
+	RunAsNonRoot               OkNok
+	FsGroupPresent             OkNok
+	SeLinuxContextPresent      OkNok
+	Capabilities               string
```
edcdavid, line 60: capabilities -> CapabilitiesCategory

cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go
```diff
@@ -0,0 +1,454 @@
...
+	HaveDropCapabilities       OkNok
```
edcdavid, line 61: HaveDropCapabilities -> RequiredDropCapabilitiesPresent

cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go
```diff
@@ -0,0 +1,454 @@
...
+	if cut.SecurityContext != nil && cut.SecurityContext.RunAsNonRoot != nil && *cut.SecurityContext.RunAsNonRoot {
+		containerSCC.RunAsNonRoot = OK
+	}
+	if cut.SecurityContext != nil && cut.SecurityContext.SELinuxOptions != nil {
+		containerSCC.SeLinuxContextPresent = OK
+	}
+	return containerSCC
+}
+
+func updateCapabilities(cut *provider.Container, containerSCC *ContainerSCC) {
```
edcdavid, line 170: updateCapabilities -> updateCapabilitiesFromContainer

cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go
```diff
@@ -0,0 +1,454 @@
...
+		containerSCC.FsGroupPresent = NOK
+	}
+	return CheckCategory(pod.Spec.Containers, containerSCC, pod.Name, pod.Namespace)
+}
+
+//nolint:gocyclo,funlen
+func compareCategory(refCategory, containerSCC *ContainerSCC, id CategoryID) bool {
+	result := true
+	tnf.ClaimFilePrintf("Testing if pod belongs to category %s", &id)
+	if refCategory.AllVolumeAllowed == containerSCC.AllVolumeAllowed {
```
edcdavid, line 364: Add comment:
// AllVolumeAllowed reports whether the volumes in the container are compliant to the SCC. True means that all volumes declared in the pod are allowed in the SCC. False means that at least one volume is disallowed

cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go
```diff
@@ -0,0 +1,454 @@
...
+		tnf.ClaimFilePrintf("AllVolumeAllowed = %s - OK", containerSCC.AllVolumeAllowed)
+	} else {
+		result = false
+		tnf.ClaimFilePrintf("AllVolumeAllowed = %s but expected >=<=%s -  NOK", containerSCC.AllVolumeAllowed, refCategory.AllVolumeAllowed)
+	}
+	if refCategory.RunAsUserPresent == containerSCC.RunAsUserPresent {
```
edcdavid, line 370: Add comment:
// RunAsUserPresent reports whether the RunAsUser Field is set to something other than nil as requested by All SCC categories. True means that the RunAsUser Field is set. False means that it is not set (nil)

cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go
```diff
@@ -0,0 +1,454 @@
...
+		tnf.ClaimFilePrintf("RunAsUserPresent = %s - OK", containerSCC.RunAsUserPresent)
+	} else {
+		tnf.ClaimFilePrintf("RunAsUserPresent = %s but expected  %s - NOK", containerSCC.RunAsUserPresent, refCategory.RunAsUserPresent)
+		result = false
+	}
+	if refCategory.RunAsNonRoot >= containerSCC.RunAsNonRoot {
```
edcdavid, line 376: Add Comment:
// RunAsNonRoot is true if the RunAsNonRoot field is set to true, false otherwise.

cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go
```diff
@@ -0,0 +1,454 @@
...
+		tnf.ClaimFilePrintf("RunAsNonRoot = %s - OK", containerSCC.RunAsNonRoot)
+	} else {
+		tnf.ClaimFilePrintf("RunAsNonRoot = %s but expected  %s - NOK", containerSCC.RunAsNonRoot, refCategory.RunAsNonRoot)
+		result = false
+	}
+	if refCategory.FsGroupPresent == containerSCC.FsGroupPresent {
```
edcdavid, line 382: // FsGroupPresent reports whether the FsGroup Field is set to something other than nil as requested by All SCC categories. True means that the FsGroup Field is set. False means that it is not set (nil)

cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go
```diff
@@ -0,0 +1,454 @@
...
+		tnf.ClaimFilePrintf("FsGroupPresent  = %s - OK", containerSCC.FsGroupPresent)
+	} else {
+		tnf.ClaimFilePrintf("FsGroupPresent  = %s but expected  %s - NOK", containerSCC.FsGroupPresent, refCategory.FsGroupPresent)
+		result = false
+	}
+	if refCategory.HaveDropCapabilities == containerSCC.HaveDropCapabilities {
```
edcdavid, line 388: // HaveDropCapabilities is true if the drop DropCapabilities field has at least the set of required drop capabilities ( same required set for all categories ). False means that some required DropCapabilities are missing.

cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go
```diff
@@ -0,0 +1,454 @@
...
+	containerSCC.HaveDropCapabilities = NOK
+	if cut.SecurityContext != nil && cut.SecurityContext.Capabilities != nil {
+		var sliceDropCapabilities []string
+		for _, ncc := range cut.SecurityContext.Capabilities.Drop {
+			sliceDropCapabilities = append(sliceDropCapabilities, string(ncc))
+		}
+		sort.Strings(sliceDropCapabilities)
+
+		sort.Strings(requiredDropCapabilities)
+		if reflect.DeepEqual(sliceDropCapabilities, requiredDropCapabilities) || reflect.DeepEqual(sliceDropCapabilities, dropAll) {
```
edcdavid, line 180: Thinking more about it, here equal might be too restrictive. If a container has more DropCapabilities than needed for all categories it should not fail this test. So instead of equal we should test that the container DropCapabilities list contains at least the required ones

cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go
```diff
@@ -0,0 +1,454 @@
...
+		result = false
+	}
+	if refCategory.HaveDropCapabilities == containerSCC.HaveDropCapabilities {
+		tnf.ClaimFilePrintf("DropCapabilities list - OK")
+	} else {
+		tnf.ClaimFilePrintf("HaveDropCapabilities = %s but expected  %s - NOK", containerSCC.HaveDropCapabilities, refCategory.HaveDropCapabilities)
+		tnf.ClaimFilePrintf("its didnt have all the required (MKNOD, SETUID, SETGID, KILL)/(ALL) drop value ")
+		result = false
+	}
+	if refCategory.HostDirVolumePluginPresent == containerSCC.HostDirVolumePluginPresent {
```
edcdavid, line 395: Add Comment:
// HostDirVolumePluginPresent is true if a hostpath volume is configured, false otherwise. It is a deprecated field and is derived from the volume list currently configured in the container. see https://docs.openshift.com/container-platform/3.11/admin_guide/manage_scc.html#use-the-hostpath-volume-plugin

cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go
```diff
@@ -0,0 +1,454 @@
...
+		tnf.ClaimFilePrintf("HostDirVolumePluginPresent = %s - OK", containerSCC.HostDirVolumePluginPresent)
+	} else {
+		tnf.ClaimFilePrintf("HostDirVolumePluginPresent = %s but expected  %s - NOK", containerSCC.HostDirVolumePluginPresent, refCategory.HostDirVolumePluginPresent)
+		result = false
+	}
+	if refCategory.HostIPC >= containerSCC.HostIPC {
```
edcdavid, line 401: Add comment:
// HostIPC  is true if the HostIPC field is set to true, false otherwise.

cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go
```diff
@@ -0,0 +1,454 @@
...
+		tnf.ClaimFilePrintf("HostIPC = %s - OK", containerSCC.HostIPC)
+	} else {
+		result = false
+		tnf.ClaimFilePrintf("HostIPC = %s but expected <= %s - NOK", containerSCC.HostIPC, refCategory.HostIPC)
+	}
+	if refCategory.HostNetwork >= containerSCC.HostNetwork {
```
edcdavid, line 407: Add comment:
// HostNetwork  is true if the HostNetwork field is set to true, false otherwise.

cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go
```diff
@@ -0,0 +1,454 @@
...
+		tnf.ClaimFilePrintf("HostNetwork = %s - OK", containerSCC.HostNetwork)
+	} else {
+		result = false
+		tnf.ClaimFilePrintf("HostNetwork = %s but expected <= %s - NOK", containerSCC.HostNetwork, refCategory.HostNetwork)
+	}
+	if refCategory.HostPID >= containerSCC.HostPID {
```
edcdavid, line 413: Add comment:
// HostPID is true if the HostPID field is set to true, false otherwise.

cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go
```diff
@@ -0,0 +1,454 @@
...
+		tnf.ClaimFilePrintf("HostPID = %s - OK", containerSCC.HostPID)
+	} else {
+		result = false
+		tnf.ClaimFilePrintf("HostPID = %s but expected <= %s - NOK", containerSCC.HostPID, refCategory.HostPID)
+	}
+	if refCategory.HostPorts >= containerSCC.HostPorts {
```
edcdavid, line 419: Add comment:
// HostPorts is true if the HostPorts field is set to true, false otherwise.

cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go
```diff
@@ -0,0 +1,454 @@
...
+		result = false
+		tnf.ClaimFilePrintf("HostPorts = %s but expected <= %s - NOK", containerSCC.HostPorts, refCategory.HostPorts)
+	}
+	if refCategory.PrivilegeEscalation >= containerSCC.PrivilegeEscalation {
+		tnf.ClaimFilePrintf("HostNetwork = %s - OK", containerSCC.HostNetwork)
+	} else {
+		result = false
+		tnf.ClaimFilePrintf("PrivilegeEscalation = %s but expected <= %s - NOK", containerSCC.PrivilegeEscalation, refCategory.PrivilegeEscalation)
+	}
+	if refCategory.PrivilegedContainer >= containerSCC.PrivilegedContainer {
```
edcdavid, line 431: Add comment:
// PrivilegedContainer is true if the PrivilegedContainer field is set to true, false otherwise.

cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go
```diff
@@ -0,0 +1,454 @@
...
+		tnf.ClaimFilePrintf("PrivilegedContainer = %s - OK", containerSCC.PrivilegedContainer)
+	} else {
+		result = false
+		tnf.ClaimFilePrintf("PrivilegedContainer = %s but expected <= %s - NOK", containerSCC.PrivilegedContainer, refCategory.PrivilegedContainer)
+	}
+	if refCategory.ReadOnlyRootFilesystem >= containerSCC.ReadOnlyRootFilesystem {
```
edcdavid, line 437: Add comment:
// ReadOnlyRootFilesystem is true if the ReadOnlyRootFilesystem field is set to true, false otherwise.

cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go
```diff
@@ -0,0 +1,454 @@
...
+		tnf.ClaimFilePrintf("ReadOnlyRootFilesystem = %s - OK", containerSCC.ReadOnlyRootFilesystem)
+	} else {
+		result = false
+		tnf.ClaimFilePrintf("ReadOnlyRootFilesystem = %s but expected <= %s - NOK", containerSCC.ReadOnlyRootFilesystem, refCategory.ReadOnlyRootFilesystem)
+	}
+	if refCategory.SeLinuxContextPresent == containerSCC.SeLinuxContextPresent {
```
edcdavid, line 443: Add comment:
// SeLinuxContextPresent is true if the SeLinuxContext field is present and set to a value (e.g. not nil), false otherwise.

cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go
```diff
@@ -0,0 +1,454 @@
...
+		tnf.ClaimFilePrintf("SeLinuxContextPresent  is not nil - OK")
+	} else {
+		result = false
+		tnf.ClaimFilePrintf("SeLinuxContextPresent  = %s but expected  %s expected to be non nil - NOK", containerSCC.SeLinuxContextPresent, refCategory.SeLinuxContextPresent)
+	}
+	if refCategory.Capabilities != containerSCC.Capabilities {
```
edcdavid, line 449: Add comment:
// CapabilitiesCategory indicates the lowest SCC category to which the list of capabilities.add in the container can be mapped to.

cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go
```diff
@@ -0,0 +1,454 @@
...
+	case CategoryID4:
+		return CategoryID4String
+	case Undefined:
+		return CategoryID4String
+	}
+	return CategoryID4String
+}
+
+//nolint:gocritic
+func CheckCategory(containers []corev1.Container, containerSCC ContainerSCC, podName, nameSpace string) []PodListcategory {
```
edcdavid, line 289: checkCategory -> checkContainerCategory

cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go
```diff
@@ -0,0 +1,454 @@
...
+		OK,
+		NOK,
+		OK,
+		NOK,
+		OK,
+		OK,
+		OK,
+		category1,
+		OK,
+		OK}
```
edcdavid, line 84: Add comments to reference categories:
	Category1                = ContainerSCC{
		NOK, //HostDirVolumePluginPresent
		NOK, //HostIPC
		NOK, //HostNetwork
		NOK, //HostPID
		NOK, //HostPorts
		OK, //PrivilegeEscalation
		NOK, //PrivilegedContainer
		OK, //RunAsUserPresent
		NOK, //ReadOnlyRootFilesystem
		OK, //RunAsNonRoot
		OK, //FsGroupPresent
		OK, //SeLinuxContextPresent
		category1,//Capabilities
		OK, //HaveDropCapabilities
		OK} //AllVolumeAllowed

cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go
```diff
@@ -0,0 +1,458 @@
...
+		return OK, value
+	}
+	return NOK, value
+}
+
+type CategoryID int
+
+const (
+	Undefined CategoryID = iota
+	CategoryID1
```
edcdavid, line 244: Use these categories everywhere, remove all the string categories for instance "category1"

cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go
```diff
@@ -0,0 +1,458 @@
...
+		sort.Strings(requiredDropCapabilities)
+		if reflect.DeepEqual(sliceDropCapabilities, requiredDropCapabilities) || reflect.DeepEqual(sliceDropCapabilities, dropAll) {
+			containerSCC.HaveDropCapabilities = OK
+		}
+
+		if checkContainCategory(cut.SecurityContext.Capabilities.Add, category2AddCapabilities) {
+			containerSCC.Capabilities = "category2"
+		} else {
+			if checkContainCategory(cut.SecurityContext.Capabilities.Add, category3AddCapabilities) {
+				containerSCC.Capabilities = "category3"
```
edcdavid, line 192: remove all the string categories for instance "category3" here

cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer_test.go
```diff
@@ -0,0 +1,108 @@
...
+
+func TestCheckPod(t *testing.T) {
+	runAs := int64(20000)
+
+	testCases := []struct {
+		testSlice     *provider.Pod
+		expectedSlice []PodListcategory
+	}{
+		{
+			testSlice: createPod(runAs, false, true, []corev1.Capability{"KILL", "MKNOD", "SETUID", "SETGID"},[]corev1.Capability{}),
```
edcdavid, line 21: I added unit test for all base category, add more as needed

cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go
```diff
@@ -0,0 +1,495 @@
...
+	// AllVolumeAllowed reports whether the volumes in the container are compliant to the SCC (same volume list for all SCCs).
+	// True means that all volumes declared in the pod are allowed in the SCC.
+	// False means that at least one volume is disallowed
+	if refCategory.AllVolumeAllowed == containerSCC.AllVolumeAllowed {
+		tnf.ClaimFilePrintf("AllVolumeAllowed = %s - OK", containerSCC.AllVolumeAllowed)
+	} else {
+		result = false
+		tnf.ClaimFilePrintf("AllVolumeAllowed = %s but expected >=<=%s -  NOK", containerSCC.AllVolumeAllowed, refCategory.AllVolumeAllowed)
+	}
+	// RunAsUserPresent reports whether the RunAsUser Field is set to something other than nil as requested by All SCC categories.
```
edcdavid, line 390: Add additional comments:
// The runAsUser range can be specified in the SCC itself. If not, it is specified in the namespace, see
// https://docs.openshift.com/container-platform/4.11/authentication/managing-security-context-constraints.html#security-context-constraints-pre-allocated-values_configuring-internal-oauth
// runAsUser:
//  type: MustRunAsRange
//  uidRangeMin: 1000
//  uidRangeMax: 2000

cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go
```diff
@@ -0,0 +1,495 @@
...
+	// True means that the RunAsUser Field is set.
+	// False means that it is not set (nil)
+	if refCategory.RunAsUserPresent == containerSCC.RunAsUserPresent {
+		tnf.ClaimFilePrintf("RunAsUserPresent = %s - OK", containerSCC.RunAsUserPresent)
+	} else {
+		tnf.ClaimFilePrintf("RunAsUserPresent = %s but expected  %s - NOK", containerSCC.RunAsUserPresent, refCategory.RunAsUserPresent)
+		result = false
+	}
+	// RunAsNonRoot is true if the RunAsNonRoot field is set to true, false otherwise.
```
edcdavid, line 399: // Add comment
// if setting a range including the roor UID 0 ( for instance 0-2000), then this option can disallow it.

cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go
```diff
@@ -0,0 +1,495 @@
...
+	if refCategory.RunAsNonRoot >= containerSCC.RunAsNonRoot {
+		tnf.ClaimFilePrintf("RunAsNonRoot = %s - OK", containerSCC.RunAsNonRoot)
+	} else {
+		tnf.ClaimFilePrintf("RunAsNonRoot = %s but expected  %s - NOK", containerSCC.RunAsNonRoot, refCategory.RunAsNonRoot)
+		result = false
+	}
+	// FsGroupPresent reports whether the FsGroup Field is set to something other than nil as requested by All SCC categories.
+	// True means that the FsGroup Field is set.
```
edcdavid, line 407: Add additional comments:
// The FSGroup range can be specified in the SCC itself. If not, it is specified in the namespace, see
// https://docs.openshift.com/container-platform/4.11/authentication/managing-security-context-constraints.html#security-context-constraints-pre-allocated-values_configuring-internal-oauth
// fsGroup:
//  type: MustRunAs
//  ranges:
//  - min: 1000900000
//    max: 1000900010

cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go
```diff
@@ -0,0 +1,495 @@
...
+		tnf.ClaimFilePrintf("PrivilegedContainer = %s but expected <= %s - NOK", containerSCC.PrivilegedContainer, refCategory.PrivilegedContainer)
+	}
+	// ReadOnlyRootFilesystem is true if the ReadOnlyRootFilesystem field is set to true, false otherwise.
+	if refCategory.ReadOnlyRootFilesystem >= containerSCC.ReadOnlyRootFilesystem {
+		tnf.ClaimFilePrintf("ReadOnlyRootFilesystem = %s - OK", containerSCC.ReadOnlyRootFilesystem)
+	} else {
+		result = false
+		tnf.ClaimFilePrintf("ReadOnlyRootFilesystem = %s but expected <= %s - NOK", containerSCC.ReadOnlyRootFilesystem, refCategory.ReadOnlyRootFilesystem)
+	}
+	// SeLinuxContextPresent is true if the SeLinuxContext field is present and set to a value (e.g. not nil), false otherwise.
```
edcdavid, line 482: // An SELinuxContext strategy of MustRunAs with no level set. Admission looks for the openshift.io/sa.scc.mcs annotation to populate the level.

sebrandon1 (approved):
Looks good!

sebrandon1:
Looks like there is a linter failure:

```
ifElseChain: rewrite if-else to switch statement (gocritic)
```

ramperher (approved):
Fine from our side!

PR #608, opened 2022-10-24
Remove redundant affinity/anti-affinity tests

sebrandon1:
Thanks to @shaior for pointing out these redundant tests.

Essentially what we were doing was duplicating tests when all we needed to look at was the Pod YAMLs.

For example, `st.Spec.Template.Spec.Affinity.PodAntiAffinity` was what we were looking for when it came to the statefulset affinitycompliant check.  This was going into the "Template" section which essentially became the Pod YAML.

Also, I added a skip to the `highAvailability` test to skip whenever we encounter a deployment/statefulset that contains a pod spec like:

```
// Skip any AffinityRequired pods
//nolint:goconst
if dp.Spec.Template.Labels["AffinityRequired"] == "true" {
	continue
}
```

### DIFF
--- a/cnf-certification-test/identifiers/identifiers.go
+++ b/cnf-certification-test/identifiers/identifiers.go
@@ -104,8 +104,6 @@ var (
 	TestPodHugePages2M                       claim.Identifier
 	TestReservedExtendedPartnerPorts         claim.Identifier
 	TestAffinityRequiredPods                 claim.Identifier
-	TestAffinityRequiredDeployments          claim.Identifier
-	TestAffinityRequiredStatefulSets         claim.Identifier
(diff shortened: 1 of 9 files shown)


cnf-certification-test/identifiers/identifiers.go
```diff
@@ -104,8 +104,6 @@ var (
 	TestPodHugePages2M                       claim.Identifier
 	TestReservedExtendedPartnerPorts         claim.Identifier
 	TestAffinityRequiredPods                 claim.Identifier
-	TestAffinityRequiredDeployments          claim.Identifier
```
edcdavid, line 107: regenerate the catalog.md

rdavid (approved):
Removing code is the best part :-)

PR #631, opened 2022-11-02
Add preflight-lib reference

sebrandon1:
Follow up to #592 

Building upon work done by @komish in branch: https://github.com/komish/openshift-preflight/tree/add-lib-ha

- Adds the `preflight` test suite.
- Brings in external dependency from openshift-preflight repo.

PR #644, opened 2022-11-07
Created merged kube config file in the clientsholder.

greyerof:
build-depends: 26624
build-depends: https://github.com/dci-labs/dallas-pipelines/pull/570

Serialize the merged kube config into a temporary file and save the file path so it can be read on demand.

This could be used by Brandon's wip #631 for the preflight lib integration.

### DIFF
--- a/cnf-certification-test/suite_test.go
+++ b/cnf-certification-test/suite_test.go
@@ -122,6 +122,14 @@ func getGitVersion() string {
 	return gitDisplayRelease + " ( " + GitCommit + " )"
 }
 
+func deleteMergedKubeConfigfile(file string) {
+	log.Infof("Deleting merged kube config file %s", file)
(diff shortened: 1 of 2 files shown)


sebrandon1 (approved):
This is great!  Exactly what's needed.

sebrandon1:
`PANIC  [Nov  7 16:49:12.053][clientsholder.go: 136] Failed to create k8s clients holder: failed to create merged kube config file: failed to create temporary file: open /tmp/tnf-merged-kubeconfig-2690352657: no such file or directory `

greyerof:
Yep, I already saw that error... not sure why yet, though. Seems like the /tmp folder in this github ubuntu runners is not writable (by default?) or it does not exist... I'm investigating it.

PR #647, opened 2022-11-07
4.1.x Maintenance

sebrandon1:
build-depends: 26624
build-depends: https://github.com/dci-labs/dallas-pipelines/pull/570

Testing out how to cherry-pick in the commits that have gone in since we released 4.1.0.  I intend to make this like a weekly cadence thing where I pull in all of the fixes but leave any potential future new tests in the next 4.2 (or whatever minor release comes next).

As of right now I don't think I'll have it build an image automatically.

Maybe we don't even need this if we release on a more frequent basis.  Our last release was from July to November so that's too long for fixes to get in.


sebrandon1:
Merging these backports.  I'm going to see if I can do this weekly on Fridays.

PR #657, opened 2022-11-10
Change StringInSlice to work with any string based type

edcdavid:
Modify StringInSlice to work with all string types

### DIFF
--- a/pkg/stringhelper/stringhelper.go
+++ b/pkg/stringhelper/stringhelper.go
@@ -21,14 +21,14 @@ import (
 )
 
 // StringInSlice checks a slice for a given string.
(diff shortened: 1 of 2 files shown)


greyerof:
Why not just fixing the panic in this commit (just the nil checkings) and then create a separate PR for those StringSlice improvement?

sebrandon1:
Was `cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go` added to the commit by accident?

edcdavid:
> Was `cnf-certification-test/accesscontrol/securitycontextcontainer/securitycontextcontainer.go` added to the commit by accident?

yes, good catch !

sebrandon1 (approved):

PR #660, opened 2022-11-10
Removed duplication of deployment scaling

shaior:
The TC removed is a duplication of lifecycle-deployment-scaling. 
There are only 2 scaling TCS: 
1. statefulset.
2. deployment.

### DIFF
--- a/CATALOG.md
+++ b/CATALOG.md
@@ -545,20 +545,7 @@ Suggested Remediation|Add a readiness probe to deployed containers
 Best Practice Reference|https://connect.redhat.com/sites/default/files/2022-05/Cloud%20Native%20Network%20Function%20Requirements%201-3.pdf Section 5.2.16, 12.1 and 12.5
 Exception Process|There is no documented exception process for this.
 Tags|common
-#### scaling
 
-Property|Description
----|---
-Test Case Name|scaling
-Test Case Label|lifecycle-scaling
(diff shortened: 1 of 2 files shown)


CATALOG.md
```diff
@@ -545,20 +545,7 @@ Suggested Remediation|Add a readiness probe to deployed containers
 Best Practice Reference|https://connect.redhat.com/sites/default/files/2022-05/Cloud%20Native%20Network%20Function%20Requirements%201-3.pdf Section 5.2.16, 12.1 and 12.5
 Exception Process|There is no documented exception process for this.
 Tags|common
-#### scaling
```
greyerof, line 548: The CATALOG.md file is auto-generated by a `make` target. It reads the identifiers.go and creates the markdown text accordingly. In short, if you want to remove any TC:
1. Remove the TC identifier from identifiers.go. In your case, [this one](https://github.com/test-network-function/cnf-certification-test/blob/a41b88a1027f616cf1ef95013f9c39de97be9d48/cnf-certification-test/identifiers/identifiers.go#L1154).
2. Run `make build`: the CATALOG.md file will be automatically updated. Run `git diff` to make sure that file has the expected changes.
3. Commit changes in both files and update the PR.

shaior: Thank you for the explanation.

sebrandon1 (approved):

PR #673, opened 2022-11-11
Fix networking tests that fail with istio-proxy related issues

sebrandon1:
Fix a couple of `networking` suite tests that still fail due to istio-proxy being injected into pods.

Created a method in the `Pod` struct called `ContainsIstioProxy` which just allows us whether or not a pod has an injected `istio-proxy` container.  The "[]Pod.Containers" are not filtered the way the "env.Containers" slices are filtered, see #554.

Notes:
- GetReservedIstioPorts() returns a hardcoded slice of ports and commented details about what they are used for.

### DIFF
--- a/cnf-certification-test/networking/netcommons/netcommons.go
+++ b/cnf-certification-test/networking/netcommons/netcommons.go
@@ -160,6 +160,20 @@ func FindRogueContainersDeclaringPorts(containers []*provider.Container, portsTo
 	return rogueContainers
 }
 
+var ReservedIstioPorts = map[int32]bool{
+	// https://istio.io/latest/docs/ops/deployment/requirements/#ports-used-by-istio
+	15090: true, // Envoy Prometheus telemetry
+	15053: true, // DNS port, if capture is enabled
+	15021: true, // Health checks
+	15020: true, // Merged Prometheus telemetry from Istio agent, Envoy, and application
+	15009: true, // HBONE port for secure networks
+	15008: true, // HBONE mTLS tunnel port
+	15006: true, // Envoy inbound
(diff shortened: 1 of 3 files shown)


sebrandon1:
An alternate suggestion to this, might be to also filter the `pod.Containers` objects as well to remove them from slices at the autodiscover level.  Let me know what you think.

cnf-certification-test/networking/netcommons/netcommons.go
```diff
@@ -160,7 +160,24 @@ func FindRogueContainersDeclaringPorts(containers []*provider.Container, portsTo
 	return rogueContainers
 }
 
+func GetReservedIstioPorts() map[int32]bool {
```
edcdavid, line 163: I would make it a variable:
```
var ReservedIstioPorts = map[int32]bool{
		15090: true, // Envoy Prometheus telemetry
		15053: true, // DNS port, if capture is enabled
		15021: true, // Health checks
		15020: true, // Merged Prometheus telemetry from Istio agent, Envoy, and application
		15009: true, // HBONE port for secure networks
		15008: true, // HBONE mTLS tunnel port
		15006: true, // Envoy inbound
		15004: true, // Debug port
		15001: true, // Envoy outbound
		15000: true, // Envoy admin port (commands/diagnostics)
	}
```

edcdavid (approved):

greyerof (approved):

cnf-certification-test/networking/netcommons/netcommons.go
sebrandon1: Done.

PR #677, opened 2022-11-14
Remove fetch tool to use OCT instead

jmontesi:
The fetch tool as well as the DB for containers, helm charts and operatos are removed from the CNF repo. The OCT tool should be used instead to build the DB that CNF consumes. Also, as of this change there is no need for a Github workflow to maintain this internal DB.


ramperher:
Discussing with @jmontesi about why DCI job is not being launched, seems to be a problem with rebasing the fork. I've seen yesterday's check and it passed: https://www.distributed-ci.io/jobs/28e68349-2e7b-4816-85fd-1f23445a3b32/jobStates?sort=date, so from my side, feel free to merge even though the check is red.

bnshr (approved):

Issue #676, opened 2022-11-14
networking-iptables test failing in OCP 4.12


ramperher:
Hi team.

Coming here to report an issue with networking-iptables test in OCP 4.12. My impression is that OCP 4.12 resources may be creating some iptables rules in the pods deployed in the cluster (maybe because the security policies that are updated starting in this OCP version), and it is hitting the failure on this test.

Here you have a DCI job example on OCP 4.12:

https://www.distributed-ci.io/jobs/476e4771-b682-4c36-8da7-58d15630a08e/tests/f6e71ac0-d4e8-4b2a-a083-9baf46ec4a3c

```
Non-compliant iptables config on: container: test pod: test-648dc964f9-blg7d ns: test-cnf log: # Generated by iptables-save v1.8.4 on Sat Nov 12 06:22:59 2022
*filter
:INPUT ACCEPT [122:9784]
:FORWARD ACCEPT [0:0]
:OUTPUT ACCEPT [119:9713]
-A FORWARD -p tcp -m tcp --dport 22623 --tcp-flags FIN,SYN,RST,ACK SYN -j REJECT --reject-with icmp-port-unreachable
-A FORWARD -p tcp -m tcp --dport 22624 --tcp-flags FIN,SYN,RST,ACK SYN -j REJECT --reject-with icmp-port-unreachable
-A FORWARD -d 169.254.169.254/32 -p tcp -m tcp ! --dport 53 -j REJECT --reject-with icmp-port-unreachable
-A FORWARD -d 169.254.169.254/32 -p udp -m udp ! --dport 53 -j REJECT --reject-with icmp-port-unreachable
-A OUTPUT -p tcp -m tcp --dport 22623 --tcp-flags FIN,SYN,RST,ACK SYN -j REJECT --reject-with icmp-port-unreachable
-A OUTPUT -p tcp -m tcp --dport 22624 --tcp-flags FIN,SYN,RST,ACK SYN -j REJECT --reject-with icmp-port-unreachable
-A OUTPUT -d 169.254.169.254/32 -p tcp -m tcp ! --dport 53 -j REJECT --reject-with icmp-port-unreachable
-A OUTPUT -d 169.254.169.254/32 -p udp -m udp ! --dport 53 -j REJECT --reject-with icmp-port-unreachable
COMMIT
# Completed on Sat Nov 12 06:22:59 2022

[...]

PR #684, opened 2022-11-15
Zeroing iptables chains counters before comparing

edcdavid:
According to https://bugzilla.redhat.com/show_bug.cgi?id=1512550 chain counters might be non-zero even without "-c" option for iptables-save. Zeroing iptables chain counters before making comparison for whitlisting.
This is to address https://github.com/test-network-function/cnf-certification-test/issues/676


sebrandon1 (approved):

ramperher (approved):
Now it's passing the iptables test in OCP 4.12, thanks!!

PR #699, opened 2022-11-28
Stop if oc doesn't exist

rdavid:


edcdavid (approved):

run-cnf-suites.sh
```diff
@@ -22,6 +22,24 @@ usage_error() {
 	usage
 	exit 1
 }
+
+# Checks whether all commands exits. Loops over the arguments, each one is a
+# command name.
+cmd_exists() {
+	local arg ret=0
+	for arg do
```
greyerof, line 30: Just for the record: My knowledge about bash is quite limited, so I spent "only" half an hour doing google searches about this way to iterate over arguments in bash, as only examples of "for elem in list" will appear. Finally, I found an example in this "advanced bash scripting" guide: https://tldp.org/LDP/abs/html/loops1.html#EX22 (example 11-6):
```
Omitting the in [list] part of a for loop causes the loop to operate on $@ -- the [positional parameters]
(https://tldp.org/LDP/abs/html/internalvariables.html#POSPARAMREF).
#!/bin/bash

#  Invoke this script both with and without arguments,
#+ and see what happens.

for a
do
 echo -n "$a "
done

#  The 'in list' missing, therefore the loop operates on '$@'
#+ (command-line argument list, including whitespace).

echo

exit 0
```
Today, I hate bash scripting a little bit more :)

greyerof (approved):

run-cnf-suites.sh
rdavid: @rdavid I love POSIX scripting, bash sucks :-)

PR #701, opened 2022-11-29
Get rid of 'oc' binary dependency.

greyerof:
We used the oc command to detect whether the suite was running against an OCP cluster or not (kind/minikube) in the cnf-run-suites.sh script. If no OCP was detected by the 'oc' command, the TNF_NON_OCP_CLUSTER env var was exported to be "true".

As we already used the go-client to get the OCP version, I just changed the code to use that information to avoid the need for the external TNF_NON_OCP_CLUSTER.

The getOpenshiftVersion() has been modified to return err only in case the client's Get() operation failed with error different than "ErrIsNotFound" or the field "versions.Name" is not the expected one in the ClusterOperator object returned. Both cases will lead to stop the execution as it doesn't make any sense to run any TC in that context.


pkg/autodiscover/autodiscover.go
```diff
@@ -190,22 +195,23 @@ func namespacesListToStringList(namespaceList []configuration.Namespace) (string
...
 	clusterOperator, err = oClient.ClusterOperators().Get(context.TODO(), "openshift-apiserver", metav1.GetOptions{})
-	// error here indicates logged in as non-admin, log and move on
 	if err != nil {
 		switch {
-		case kerrors.IsForbidden(err), kerrors.IsNotFound(err):
-			logrus.Infof("OpenShift Version not found (must be logged in to cluster as admin): %v", err)
-			err = nil
+		case kerrors.IsNotFound(err):
+			logrus.Warnf("Unable to get ClusterOperator CR from openshift-apiserver. Running in a non-OCP cluster.")
+			return "", nil
```
rdavid, line 202: Add const `none` or `0.0.0+nil` in case of not found.

pkg/autodiscover/autodiscover.go
```diff
@@ -190,22 +195,23 @@ func namespacesListToStringList(namespaceList []configuration.Namespace) (string
...
 	var clusterOperator *configv1.ClusterOperator
 	clusterOperator, err = oClient.ClusterOperators().Get(context.TODO(), "openshift-apiserver", metav1.GetOptions{})
-	// error here indicates logged in as non-admin, log and move on
 	if err != nil {
 		switch {
-		case kerrors.IsForbidden(err), kerrors.IsNotFound(err):
-			logrus.Infof("OpenShift Version not found (must be logged in to cluster as admin): %v", err)
-			err = nil
+		case kerrors.IsNotFound(err):
+			logrus.Warnf("Unable to get ClusterOperator CR from openshift-apiserver. Running in a non-OCP cluster.")
```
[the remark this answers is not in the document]
greyerof: I don't think the err value is needed here, as it's an expected outcome: no resource of type ClusterOperator found, which is basically what the Warnf string is also printing.

jmontesi:
You can also remove the `oc` installation from the `Dockerfile`, as well as the `OPENSHIFT_VERSION` build-param and the copy of the binary to the final TNF image.

pkg/autodiscover/autodiscover.go
greyerof: Done.

rdavid (approved):

pkg/autodiscover/autodiscover.go
```diff
@@ -190,22 +197,23 @@ func namespacesListToStringList(namespaceList []configuration.Namespace) (string
 func getOpenshiftVersion(oClient clientconfigv1.ConfigV1Interface) (ver string, err error) {
 	var clusterOperator *configv1.ClusterOperator
 	clusterOperator, err = oClient.ClusterOperators().Get(context.TODO(), "openshift-apiserver", metav1.GetOptions{})
-	// error here indicates logged in as non-admin, log and move on
 	if err != nil {
 		switch {
-		case kerrors.IsForbidden(err), kerrors.IsNotFound(err):
```
edcdavid, line 196: Why not keep reporting this error but this time as an error instead of just log?

pkg/autodiscover/autodiscover.go
```diff
@@ -190,22 +197,23 @@ func namespacesListToStringList(namespaceList []configuration.Namespace) (string
...
-				// openshift-apiserver does not report version,
-				// clusteroperator/openshift-apiserver does, and only version number
-				return ver.Version, nil
-			}
+
+	for _, ver := range clusterOperator.Status.Versions {
+		if ver.Name == tnfCsvTargetLabelName {
+			// openshift-apiserver does not report version,
+			// clusteroperator/openshift-apiserver does, and only version number
+			return ver.Version, nil
```
edcdavid, line 214: print the version in the logs

edcdavid (approved):

pkg/autodiscover/autodiscover.go
greyerof: Not sure what you mean exactly here, can you elaborate, please?

rdavid (approved):

pkg/autodiscover/autodiscover.go
edcdavid: I was just suggesting to return an error for IsForbidden cases

edcdavid: never mind, it is already returned in default : !

PR #704, opened 2022-11-30
internal/api: refactor the api with the certification DB

jmontesi:


cnf-certification-test/certification/suite.go
```diff
@@ -43,51 +43,55 @@ const (
...
 	ginkgo.BeforeEach(func() {
+		var err error
 		env = provider.GetTestEnvironment()
-		err := api.LoadCatalog(env.GetOfflineDBPath())
+		certificator, err = certdb.GetCertificator(env.GetOfflineDBPath())
 		if err != nil {
-			logrus.Infof("Offline DB not available, err: %v", err)
+			errMsg := fmt.Sprintf("Cannot access the certification DB, err: %v", err)
+			logrus.Errorf(errMsg)
+			ginkgo.Fail(errMsg)
```
greyerof, line 54: Maybe we already have this pattern somewhere else, but I don't think it's necessary to log the same msg with logrus and  with the ginkgo.Fail(). Both will appear in the stdout, so, in this cases, I think that only the ginkgo.Fail(errMsg) is needed.

cnf-certification-test/certification/certtool/certtool.go
```diff
@@ -30,7 +30,7 @@ const (
 )
 
 var (
-	CertAPIClient api.CertificationValidator
+	CertAPIClient certdb.ResourceCertificator
```
greyerof, line 33: I don't think this variable is needed anymore. Same with the functions that use it in this file: they seem not to be used:
GetContainerCertificationRequestFunction
GetOperatorCertificationRequestFunction
WaitForCertificationRequestToSuccess

internal/certdb/certdb.go
```diff
@@ -0,0 +1,32 @@
...
+import (
+	"fmt"
+
+	"github.com/test-network-function/cnf-certification-test/internal/certdb/offlinecheck"
+	"github.com/test-network-function/cnf-certification-test/internal/certdb/onlinecheck"
+	"helm.sh/helm/v3/pkg/release"
+)
+
+//go:generate moq -out api_moq.go . CertAPIClientFuncs
+type ResourceCertificator interface {
```
greyerof, line 12: This is a personal opinion, but "Certificator" suffix seems a bit misleading to me, as we're not certifiying any resources, just getting their certification status. I would stick to "StatusChecker" or "CertificationStatusValidator". I see you also changed some function receiver names from "checker" to "validator", so maybe the second option is better for consistency.

internal/certdb/certdb.go
```diff
@@ -0,0 +1,32 @@
...
+
+import (
+	"fmt"
+
+	"github.com/test-network-function/cnf-certification-test/internal/certdb/offlinecheck"
+	"github.com/test-network-function/cnf-certification-test/internal/certdb/onlinecheck"
+	"helm.sh/helm/v3/pkg/release"
+)
+
+//go:generate moq -out api_moq.go . CertAPIClientFuncs
```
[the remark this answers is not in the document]
jmontesi: Done.

internal/certdb/certdb.go
```diff
@@ -0,0 +1,32 @@
...
+}
+
+func GetCertificator(offlineDBPath string) (ResourceCertificator, error) {
+	// use the online certificator by default
+	onlineValidator := onlinecheck.NewOnlineValidator()
+	if onlineValidator.IsServiceReachable() {
+		return onlineValidator, nil
+	}
+
+	// use the offline DB for disconnected environments
```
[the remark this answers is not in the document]
jmontesi: Done.

cnf-certification-test/certification/suite.go
jmontesi: Done.

greyerof (approved):

Issue #672, opened 2022-11-11
DeploymentConfig must be considered in deployment-scaling and pod-owner-type tests.


tginer:
Using latest release v4.1.0, Replicas inside the **DeploymentConfig** resource are not being considered. 

**lifecycle-deployment-scaling** was skipped with log:
`Test skipped because there are no []*provider.Deployment to test`
yet the application had multiple replicas defined.

**lifecycle-pod-owner-type** was failed with log:
`Tests that CNF Pod(s) are deployed as part of a ReplicaSet(s)/StatefulSet(s)`
and the application has ReplicaSets defined from DeploymentConfig resource

sebrandon1:
It seems like `DeploymentConfig` objects are OCP only.  Right now we're only looking at the first-class native `Deployment` objects in Kubernetes.

In the OpenShift [document](https://docs.openshift.com/container-platform/4.11/applications/deployments/what-deployments-are.html#deployments-comparing-deploymentconfigs_what-deployments-are) about `DeploymentConfig` objects, it says:

```
Both Kubernetes Deployment objects and OpenShift Container Platform-provided DeploymentConfig objects are supported in OpenShift Container Platform; however, it is recommended to use Deployment objects unless you need a specific feature or behavior provided by DeploymentConfig objects.
```

Is there a certain feature/behavior the partner is using that they need to use the `DeploymentConfig` objects versus regular `Deployment` objects?  We can look into adding support for `DeploymentConfig`.

sebrandon1:
After doing some more discussion and digging into DeploymentConfigs, it seems to be more of a "legacy" support feature rather than something that OCP users should actually be using.  

It seems like the main feature is DeploymentConfig pushes is being able to do inplace updates but native Deployment objects already have the ability to do a rolling update, doc [here](https://kubernetes.io/docs/tutorials/kubernetes-basics/update/update-intro/).

Video with more context on DeploymentConfigs: https://www.youtube.com/watch?v=FFqy-H7slXE

sebrandon1:
Just for future reference, I started a branch here: https://github.com/sebrandon1/cnf-certification-test/commit/8dace9fd73a355e1dd8e1df3e56a192642818863 that started into adding DeploymentConfig support into our autodiscover process but I didn't get to the part where I added the objects into our tests.

tginer:
Hi @sebrandon1 , 

[...]

PR #709, opened 2022-12-01
Detect pods created by Openshift's DeploymentConfig resources

shimritproj:
Related to #672 

Check if the pods is created by Openshift's DeploymentConfig resources.

### DIFF
--- a/pkg/provider/pods.go
+++ b/pkg/provider/pods.go
@@ -185,21 +185,20 @@ func (p *Pod) ContainsIstioProxy() bool {
 	return false
 }
 
-func (p *Pod) CreatedByDeploymentConfig() bool {
-	for _, data := range p.ObjectMeta.GetOwnerReferences() {
-		if data.Kind == replicationController {
-			oc := clientsholder.GetClientsHolder()
-			replicationControllers, err := oc.K8sClient.CoreV1().ReplicationControllers(p.Namespace).Get(context.TODO(), data.Name, v1.GetOptions{})
+func (p *Pod) CreatedByDeploymentConfig() (bool, error) {
+	oc := clientsholder.GetClientsHolder()
+	for _, podOwner := range p.ObjectMeta.GetOwnerReferences() {
+		if podOwner.Kind == replicationController {
+			replicationControllers, err := oc.K8sClient.CoreV1().ReplicationControllers(p.Namespace).Get(context.TODO(), podOwner.Name, v1.GetOptions{})
 			if err != nil {
-				logrus.Errorln("Error when listing pods in ns=", p.Namespace, " err: ", err)
-				continue
+				return false, err
 			}
-			for _, x := range replicationControllers.GetOwnerReferences() {
-				if x.Name == data.Name && x.Kind == deploymentConfig {
-					return true
+			for _, rcOwner := range replicationControllers.GetOwnerReferences() {
+				if rcOwner.Name == podOwner.Name && rcOwner.Kind == deploymentConfig {
+					return true, err
 				}
 			}
 		}
 	}
-	return false
+	return false, nil
 }
--- a/pkg/provider/pods.go
+++ b/pkg/provider/pods.go
@@ -17,19 +17,24 @@
 package provider
 
 import (
+	"context"
 	"fmt"
 	"strconv"
 	"strings"
 
 	"github.com/sirupsen/logrus"
+	"github.com/test-network-function/cnf-certification-test/internal/clientsholder"
 	"github.com/test-network-function/cnf-certification-test/pkg/tnf"
 	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 const (
-	hugePages2Mi = "hugepages-2Mi"
-	hugePages1Gi = "hugepages-1Gi"
-	hugePages    = "hugepages"
+	hugePages2Mi          = "hugepages-2Mi"
+	hugePages1Gi          = "hugepages-1Gi"
+	hugePages             = "hugepages"
+	replicationController = "ReplicationController"
+	deploymentConfig      = "DeploymentConfig"
 )
 
 type Pod struct {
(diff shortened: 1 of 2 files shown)


pkg/provider/provider.go
```diff
@@ -220,6 +220,11 @@ func buildTestEnvironment() { //nolint:funlen
 	operators := createOperators(data.Csvs, data.Subscriptions, data.AllInstallPlans, data.AllCatalogSources, false, false, true)
 	env.Operators = operators
 	logrus.Infof("Operators found: %d", len(env.Operators))
+	for _, p := range env.Pods {
+		if !(p.CreatedByDeploymentConfig()) {
```
rdavid, line 224: There is no need for additional `()`, just:
```
if !p.CreatedByDeploymentConfig() {
```

pkg/provider/pods.go
```diff
@@ -179,3 +179,12 @@ func (p *Pod) ContainsIstioProxy() bool {
 	}
 	return false
 }
+
+func (p *Pod) CreatedByDeploymentConfig() bool {
+	for _, data := range p.ObjectMeta.GetOwnerReferences() {
+		if data.Kind != "ReplicationController" {
```
bnshr, line 185: Use a const here instead of quoted string.

bnshr (changes requested):

pkg/provider/pods.go
greyerof, line 185: A couple of issues here:
1. You return "false" as soon as the first "ownerReference" is not a ReplicationController. What if you have more than one ownerReference in the slice but the ReplicationController reference is the last one?
2. DeploymentConfig resources will create ReplicationControllers to deploy the pods. But you can also have ReplicationControllers in your cluster that have been deployed manually, without any DeploymentConfig. You need to return "true" only in case the owner of the pod is a ReplicationController whose owner is a DeploymentConfig. For example, if your pod's name is "X":
```
Pod X:
   ownerRerefences:
      - kind: ReplicationController
	  - name: rc1
	  
ReplicationController (mame: rc1)
   ownerRerefences:
      - kind: DeploymentConfig
```

pkg/provider/provider.go
greyerof, line 224: The idea of the check is the opposite: add a warning log trace when a pod has been deployed by a DeploymentConfig. Something like this:
```
for _, pod := range env.Pods {
   if pod.CreatedByDeploymentConfig() {
      logrus.Warnf("Pod %s has been deployed using a DeploymentConfig, please use Deployment or StatefulSet instead.", pod.String())
   }
}
```

pkg/provider/pods.go
shimritproj: I agree and fixed it.

bnshr (approved):

sebrandon1:
After some discussion with @shimritproj, we might be able to combine what we have here with the commit that I have in https://github.com/sebrandon1/cnf-certification-test/commit/8dace9fd73a355e1dd8e1df3e56a192642818863.

This PR basically just mentions to the user that their pods have an owner type of `DeploymentConfig` and they should probably use either a normal Deployment or StatefulSet/ReplicaSet per the recommendation of the [official document](https://docs.openshift.com/container-platform/4.11/applications/deployments/what-deployments-are.html#deployments-comparing-deploymentconfigs_what-deployments-are).

pkg/provider/pods.go
```diff
@@ -179,3 +184,22 @@ func (p *Pod) ContainsIstioProxy() bool {
...
+func (p *Pod) CreatedByDeploymentConfig() bool {
+	for _, data := range p.ObjectMeta.GetOwnerReferences() {
+		if data.Kind == replicationController {
+			oc := clientsholder.GetClientsHolder()
+			replicationControllers, err := oc.K8sClient.CoreV1().ReplicationControllers(p.Namespace).Get(context.TODO(), data.Name, v1.GetOptions{})
+			if err != nil {
+				logrus.Errorln("Error when listing pods in ns=", p.Namespace, " err: ", err)
+				continue
+			}
+			for _, x := range replicationControllers.GetOwnerReferences() {
```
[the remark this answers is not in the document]
shimritproj: I agree.

pkg/provider/pods.go
```diff
@@ -179,3 +184,22 @@ func (p *Pod) ContainsIstioProxy() bool {
...
 }
+
+func (p *Pod) CreatedByDeploymentConfig() bool {
+	for _, data := range p.ObjectMeta.GetOwnerReferences() {
+		if data.Kind == replicationController {
+			oc := clientsholder.GetClientsHolder()
+			replicationControllers, err := oc.K8sClient.CoreV1().ReplicationControllers(p.Namespace).Get(context.TODO(), data.Name, v1.GetOptions{})
+			if err != nil {
+				logrus.Errorln("Error when listing pods in ns=", p.Namespace, " err: ", err)
+				continue
```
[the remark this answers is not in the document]
shimritproj: I agree.

pkg/provider/provider.go
```diff
@@ -223,6 +223,15 @@ func buildTestEnvironment() { //nolint:funlen
 	operators := createOperators(data.Csvs, data.Subscriptions, data.AllInstallPlans, data.AllCatalogSources, false, false, true)
 	env.Operators = operators
 	logrus.Infof("Operators found: %d", len(env.Operators))
+	for _, pod := range env.Pods {
+		isCreatedByDeploymentConfig, err := pod.CreatedByDeploymentConfig()
+		if err != nil {
+			logrus.Warnf("Pod %s, err: %s", pod.String(), err)
+		}
+		if !isCreatedByDeploymentConfig {
```
greyerof, line 231: As I said in my previous comment, this value (isCreatedByDeploymentConfig) doesn't make any sense when err != nil, because the error means that you weren't able to determine whether it was created by a deploymentConfig or not, so just reporting the Warnf should be enough. You can add a "continue" in the `if err != nil`:
```
for _, pod := range env.Pods {
		isCreatedByDeploymentConfig, err := pod.CreatedByDeploymentConfig()
		if err != nil {
			logrus.Errorf("...")
			continue
		}
		if !isCreatedByDeploymentConfig {
		   logrus.Warnf("...")
		}
}
```

pkg/provider/provider.go
```diff
@@ -223,6 +223,15 @@ func buildTestEnvironment() { //nolint:funlen
 	operators := createOperators(data.Csvs, data.Subscriptions, data.AllInstallPlans, data.AllCatalogSources, false, false, true)
 	env.Operators = operators
 	logrus.Infof("Operators found: %d", len(env.Operators))
+	for _, pod := range env.Pods {
+		isCreatedByDeploymentConfig, err := pod.CreatedByDeploymentConfig()
+		if err != nil {
+			logrus.Warnf("Pod %s, err: %s", pod.String(), err)
```
greyerof, line 229: Please try to be more verbose with this kind of log traces. Also, maybe using `logrus.Errorf` is more appropriate as you're actually reporting an important error here. My suggestion:
```
logrus.Errorf("Failed to know whether pod %s was created by a DeploymentConfig: %v", pod.String(), err)
```

greyerof (approved):
Added a couple of minor suggestions.

sebrandon1 (approved):
Approved pending @greyerof's suggestions

PR #711, opened 2022-12-01
Fix shellcheck errors, no logic change

rdavid:


ramperher:
@rdavid , it looks like something has changed in the script and tnf is not being able to run correctly on the DCI job check, I tested it two times and in both cases the execution failed when trying to run the script. I think it needs to be reviewed.

rdavid:
@ramperher Fixed!

PR #714, opened 2022-12-05
Fix shellcheck errors, logic is not changed

rdavid:


rdavid:
@sebrandon1 DCI test successed.

sebrandon1 (approved):

PR #738, opened 2022-12-13
Add doc section about running all of the tests at once

sebrandon1:
Adds a note about running all of the tests, common and extended.

### DIFF
--- a/docs/test-standalone.md
+++ b/docs/test-standalone.md
@@ -89,6 +89,13 @@ To select the test to be executed when running `run-cnf-suites.sh` with the foll
 
     The test labels work the same as the suite labels, so you can select more than one test with the filtering mechanism shown before.
 
+### Run all of the tests
+You can run all of the tests (including the intrusive tests and the extended suite) with the following commands:
+
+```shell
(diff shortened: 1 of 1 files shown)


docs/test-standalone.md
```diff
@@ -89,6 +89,13 @@ To select the test to be executed when running `run-cnf-suites.sh` with the foll
 
     The test labels work the same as the suite labels, so you can select more than one test with the filtering mechanism shown before.
 
+### Run all of the tests
+You can run all of the tests (including the intrusive tests and the extended suite) with the following commands:
+
+```shell
+ TNF_NON_INTRUSIVE_ONLY=false ./run-cnf-suites.sh -l common,extended
```
rdavid, line 96: There is no need for setting the variable, it is false by default:
 https://github.com/test-network-function/cnf-certification-test/blob/main/script/run-container.sh#L26

rdavid (approved):

docs/test-standalone.md
sebrandon1: Good call, I will remove that.

greyerof (approved):

PR #744, opened 2022-12-15
Adjust Dockerfile to be able to built locally

sebrandon1:
Notable changes:
- Changed the github workflows to run `make build-image-local` instead of copy/pasting the docker build command information.
- Removed the logic from the Dockerfile that clones the repo(not sure why we ever did this) and builds the binary to just use the already built binary as part of `make build-cnf-tests` which is ran in `make build-image-local` as a pre-req.
- Added default values/strings in the Makefile.
- Removed `jq` and `git` dependencies from the first stage of the Dockerfile.
- BONUS: Added `operator-sdk` binary to the image PATH to be able to use in the preflight work in #631.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,6 @@ FROM registry.access.redhat.com/ubi8/ubi:latest AS build
 
 ARG OPENSHIFT_VERSION
 ENV OPENSHIFT_VERSION=${OPENSHIFT_VERSION}
-
 ENV TNF_DIR=/usr/tnf
 ENV TNF_SRC_DIR=${TNF_DIR}/tnf-src
 ENV TNF_BIN_DIR=${TNF_DIR}/cnf-certification-test
@@ -14,20 +13,11 @@ ENV TEMP_DIR=/tmp
 # Install dependencies
 RUN yum install -y gcc make wget
 
-# Download oc binary
-ENV OC_BIN_TAR="openshift-client-linux.tar.gz"
-ENV OC_DL_URL="https://mirror.openshift.com/pub/openshift-v4/clients/ocp"/${OPENSHIFT_VERSION}/${OC_BIN_TAR}
-ENV OC_BIN="/usr/local/oc/bin"
-RUN wget --directory-prefix=${TEMP_DIR} ${OC_DL_URL} && \
-    mkdir -p ${OC_BIN} && \
-    tar -C ${OC_BIN} -xzf ${TEMP_DIR}/${OC_BIN_TAR} && \
-    chmod a+x ${OC_BIN}/oc
-
 # Download operator-sdk binary
-ENV OPERATOR_SDK_DL_URL=https://github.com/operator-framework/operator-sdk/releases/download/v1.26.0
-ENV OSDK_BIN="/usr/local/osdk/bin"
+ENV OPERATOR_SDK_DL_URL=https://github.com/operator-framework/operator-sdk/releases/download/v1.26.0 \
+    OSDK_BIN="/usr/local/osdk/bin"
 RUN mkdir -p ${OSDK_BIN} && \
-    curl -LO ${OPERATOR_SDK_DL_URL}/operator-sdk_linux_amd64 && \
+    curl --location --remote-name ${OPERATOR_SDK_DL_URL}/operator-sdk_linux_amd64 && \
     mv operator-sdk_linux_amd64 ${OSDK_BIN}/operator-sdk && \
     chmod +x ${OSDK_BIN}/operator-sdk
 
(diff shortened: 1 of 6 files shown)


script/get-release-level.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
```
jmontesi, line 1: After [PR#701](https://github.com/test-network-function/cnf-certification-test/pull/701) I think we can get rid of the `oc` command installed in the `Dockerfile`, as well as the `OPENSHIFT_VERSION` variable and therefore this new script wouldn't be needed.

Makefile
```diff
@@ -133,4 +141,10 @@ get-db:
 	docker pull ${OCT_IMAGE}
 	docker run -v ${REPO_DIR}/cmd/tnf/fetch:/tmp/dump:Z --user $(shell id -u):$(shell id -g) --env OCT_DUMP_ONLY=true ${OCT_IMAGE}
 delete-db:
-	rm -rf ${REPO_DIR}/cmd/tnf/fetch
\ No newline at end of file
+	rm -rf ${REPO_DIR}/cmd/tnf/fetch
+
+build-image-local: install-tools build-cnf-tests
```
jmontesi, line 146: Isn't it more convenient to perform these steps inside the `Dockerfile`? Otherwise the validity of the image generated using this `Dockerfile` is dependent of being run through the `Makefile`.

Dockerfile
```diff
@@ -34,25 +23,18 @@ RUN wget --directory-prefix=${TEMP_DIR} ${OC_DL_URL} && \
...
+ENV OPERATOR_SDK_DL_URL=https://github.com/operator-framework/operator-sdk/releases/download/v1.26.0
+ENV OSDK_BIN="/usr/local/osdk/bin"
+RUN mkdir -p ${OSDK_BIN} && \
+    curl -LO ${OPERATOR_SDK_DL_URL}/operator-sdk_linux_amd64 && \
+    mv operator-sdk_linux_amd64 ${OSDK_BIN}/operator-sdk && \
+    chmod +x ${OSDK_BIN}/operator-sdk
 
-# Build TNF binary
+# Copy all of the files into the source directory and then switch contexts
+COPY . ${TNF_SRC_DIR}
```
jmontesi, line 35: Another optimization could be to remove all the source code from the final TNF image, as I don't it's needed. Maybe keeping this `TNF_SRC_DIR` separeted from `TNF_DIR` is a good way to avoid copying it.

script/get-release-level.sh
sebrandon1: Excellent!  I'll remove it.

Makefile
sebrandon1: I'm 50/50.  On one hand it keeps it all contained within the Dockerfile, but makes it more complex because the environment already has all of the deps needed.

Dockerfile
```diff
@@ -34,25 +23,18 @@ RUN wget --directory-prefix=${TEMP_DIR} ${OC_DL_URL} && \
...
-ARG TNF_VERSION
-ARG TNF_SRC_URL=$TNF_SRC_URL
-ARG GIT_CHECKOUT_TARGET=$TNF_VERSION
-
-# Clone the TNF source repository and checkout the target branch/tag/commit
-RUN git clone --no-single-branch --depth=1 ${TNF_SRC_URL} ${TNF_SRC_DIR}
-RUN git -C ${TNF_SRC_DIR} fetch origin ${GIT_CHECKOUT_TARGET}
-RUN git -C ${TNF_SRC_DIR} checkout ${GIT_CHECKOUT_TARGET}
+# Download operator-sdk binary
+ENV OPERATOR_SDK_DL_URL=https://github.com/operator-framework/operator-sdk/releases/download/v1.26.0
```
rdavid, line 27: You can do:
```
ENV \
	FOO=BAR \
	BAR=FOO
```

Dockerfile
```diff
@@ -34,25 +23,18 @@ RUN wget --directory-prefix=${TEMP_DIR} ${OC_DL_URL} && \
...
+ENV OSDK_BIN="/usr/local/osdk/bin"
+RUN mkdir -p ${OSDK_BIN} && \
+    curl -LO ${OPERATOR_SDK_DL_URL}/operator-sdk_linux_amd64 && \
```
rdavid, line 30: Should it be a tab instead of spaces as indentation?

rdavid, line 30: It is nicer to use descriptive flags in scripts, e.g.:
```
curl --location --remote-name <URL>
```

sebrandon1: Done.

sebrandon1: I am using tabs.  Is it not showing up as tabs?

Dockerfile
sebrandon1: Great suggestion.  I adjusted all of the ENV sections to reduce the number of layers.

Dockerfile
sebrandon1: Maybe this can be an optimization in a follow up PR?

Dockerfile
rdavid: Here, at the comments, GitHub shows spaces :-)

rdavid (approved):

Makefile
jmontesi: I prefer leaving the `Dockerfile` as independent as possible, containing all the dependencies the building environment requires but would like to hear a second opinion here.

Dockerfile
jmontesi: Sure, no problem.

Makefile
sebrandon1: I'll bring back the binary building within the Dockefile to encapsulate it.

sebrandon1: Done.

jmontesi (approved):
Update the description with the latest changes.

Dockerfile
```diff
@@ -2,17 +2,18 @@ FROM registry.access.redhat.com/ubi8/ubi:latest AS build
 
 ARG OPENSHIFT_VERSION
 ENV OPENSHIFT_VERSION=${OPENSHIFT_VERSION}
```
greyerof, line 4: So if we're not using this OPENSHIFT_VERSION arg/env any more, they can be removed, right?

Dockerfile
```diff
@@ -25,33 +26,22 @@ RUN if [[ "$(uname -m)" -eq "x86_64" ]] ; then \
...
-RUN git -C ${TNF_SRC_DIR} checkout ${GIT_CHECKOUT_TARGET}
-
-# Build TNF binary
+# Copy all of the files into the source directory and then switch contexts
+COPY . ${TNF_SRC_DIR}
 WORKDIR ${TNF_SRC_DIR}
 
 RUN make install-tools && \
-	make build-cnf-tests-debug
+	make build-cnf-tests
```
greyerof, line 44: Caution: this will build the release version of the tnf tool, which means sometimes we might not get too much information from panics, due to the compiler optimizations. I'm ok with it, but I believe we should also provide a a way for users to run the tnf in debug mode, just in case. Can be done in further PRs, if agreed.

greyerof (approved):

ramperher (approved):
I'm fine from DCI perspective, because there were some jobs that were failing but not it's working fine, and as we use the Dockerfile to compile the image in some cases, it's important to have it working.

Dockerfile
sebrandon1: Good call, will remove.

Dockerfile
sebrandon1: Agreed, we can add a debug image in a future PR.

PR #756, opened 2023-01-03
Add additional labels to existing tests

sebrandon1:
Added two new tags `telco` and `faredge`, and based off of the spreadsheet where we determined which tests are considered to be tagged specifically.

### DIFF
--- a/cnf-certification-test/identifiers/identifiers.go
+++ b/cnf-certification-test/identifiers/identifiers.go
@@ -36,6 +36,8 @@ const (
 const (
 	TagCommon   = "common"
 	TagExtended = "extended"
+	TagTelco    = "telco"
+	TagFarEdge  = "faredge"
 )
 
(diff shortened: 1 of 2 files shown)


rdavid (approved):

cnf-certification-test/identifiers/identifiers.go
```diff
@@ -36,6 +36,10 @@ const (
 const (
 	TagCommon   = "common"
 	TagExtended = "extended"
+	TagTelco    = "telco"
```
jmontesi, line 39: Should we then remove the "common" and "extended" tags? I think there is some overlap with "general" and "vz".

sebrandon1: I agree, I figured we would discuss that as well as part of this PR.

jmontesi (approved):

PR #759, opened 2023-01-04
Build KubeConfig byte array

sebrandon1:
Based on changes from #631 

Adds a `clientsHolder.KubeConfig` byte array variable.

I'm trying to make the preflight lib PR less daunting before the final merge.

### DIFF
--- a/internal/clientsholder/clientsholder.go
+++ b/internal/clientsholder/clientsholder.go
@@ -171,7 +171,7 @@ func createTempFile(prefix string) (file *os.File, err error) {
 func createByteArrayKubeConfig(kubeConfig *clientcmdapi.Config) ([]byte, error) {
 	yamlBytes, err := clientcmd.Write(*kubeConfig)
 	if err != nil {
-		return []byte{}, fmt.Errorf("failed to generate yaml bytes from kubeconfig: %w", err)
+		return nil, fmt.Errorf("failed to generate yaml bytes from kubeconfig: %w", err)
 	}
 	return yamlBytes, nil
 }
--- a/internal/clientsholder/clientsholder.go
+++ b/internal/clientsholder/clientsholder.go
@@ -55,6 +55,7 @@ type ClientsHolder struct {
 	K8sNetworkingClient  networkingv1.NetworkingV1Interface
 	MachineCfg           ocpMachine.Interface
 	MergedKubeConfigFile string
+	KubeConfig           []byte
 	ready                bool
 }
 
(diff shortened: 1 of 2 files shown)


internal/clientsholder/clientsholder.go
```diff
@@ -55,6 +55,7 @@ type ClientsHolder struct {
 	K8sNetworkingClient  networkingv1.NetworkingV1Interface
 	MachineCfg           ocpMachine.Interface
 	MergedKubeConfigFile string
+	KubeConfig           []byte
```
jmontesi, line 58: Do we need to keep the two possibilities for storing the `KubeConfig`?

jmontesi, line 58: Do we need to keep the two possibilities for storing the `KubeConfig`?

internal/clientsholder/clientsholder.go
```diff
@@ -167,12 +168,20 @@ func createTempFile(prefix string) (file *os.File, err error) {
 	return file, nil
 }
 
+func createByteArrayKubeConfig(kubeConfig *clientcmdapi.Config) ([]byte, error) {
+	yamlBytes, err := clientcmd.Write(*kubeConfig)
+	if err != nil {
+		return []byte{}, fmt.Errorf("failed to generate yaml bytes from kubeconfig: %w", err)
```
jmontesi, line 174: I guess you can just return `nil` instead of an empty `byte` slice.

sebrandon1: Done.

internal/clientsholder/clientsholder.go
jmontesi: But the temp file already contains the `KubeConfig`. Can't you just read it when it's needed?

sebrandon1: Okay my latest commit https://github.com/test-network-function/cnf-certification-test/pull/759/commits/7a3e41e4832c0f9e5d12945479d09b46c82fc64d has removed the work done by @greyerof because it was unused.  The preflight suite now just uses a byte array param instead of requiring the file to be present.

internal/clientsholder/clientsholder.go
```diff
@@ -46,16 +44,17 @@ import (
...
-	ready                bool
+	RestConfig          *rest.Config
+	DynamicClient       dynamic.Interface
+	APIExtClient        apiextv1.Interface
+	OlmClient           olmClient.Interface
+	OcpClient           clientconfigv1.ConfigV1Interface
+	K8sClient           kubernetes.Interface
+	K8sNetworkingClient networkingv1.NetworkingV1Interface
+	MachineCfg          ocpMachine.Interface
+	// MergedKubeConfigFile string
```
jmontesi, line 55: Remove comment.

jmontesi (approved):

greyerof (approved):

internal/clientsholder/clientsholder.go
sebrandon1: Fixed.

PR #769, opened 2023-01-09
Fix copyright year

rdavid:


sebrandon1 (approved):

greyerof (approved):

cnf-certification-test/identifiers/identifiers.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (C) 2021-2022 Red Hat, Inc.
+// Copyright (C) 2021-2023 Red Hat, Inc.
```
jmontesi, line 1: So, what's the point of only updating the year in those three files? What about the rest?

sebrandon1: @rdavid Do you want to add it as a "linter" github workflow?

rdavid: @sebrandon1 It is a cool idea. I'll take a look, maybe the script needs to adopt.

PR #771, opened 2023-01-09
Improve tnf-image.yaml; fix deprecations

sebrandon1:
Notable changes:
- Remove the "testing" sections of the image building scripts.  The code should already be tested prior to merging as part of the PR running the same tests.
- Fix the [set-output](https://dev.to/cicirello/how-to-patch-the-deprecated-set-output-in-github-workflows-and-in-container-actions-9co#:~:text=GitHub%20recently%20deprecated%20the%20set,the%20end%20of%20May%202023.) lines that are deprecated.
- Fix the `docker push` section of the 4.0.x job to only push the TNF_VERSION, not the `latest`.

### DIFF
--- a/.github/workflows/tnf-image.yaml
+++ b/.github/workflows/tnf-image.yaml
@@ -23,7 +23,6 @@ env:
   IMAGE_TAG: latest
   TNF_CONTAINER_CLIENT: docker
   TNF_NON_INTRUSIVE_ONLY: false
-  TNF_MINIKUBE_ONLY: true
   TNF_DISABLE_CONFIG_AUTODISCOVER: false
   TNF_CONFIG_DIR: /tmp/tnf/config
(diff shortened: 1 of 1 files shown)


.github/workflows/tnf-image.yaml
```diff
@@ -23,7 +23,7 @@ env:
   IMAGE_TAG: latest
   TNF_CONTAINER_CLIENT: docker
   TNF_NON_INTRUSIVE_ONLY: false
-  TNF_MINIKUBE_ONLY: true
```
edcdavid, line 26: is it still used ?

edcdavid (approved):

.github/workflows/tnf-image.yaml
sebrandon1: Whoops no, I was messing with this prior.  I'll fix.

jmontesi (approved):

PR #777, opened 2023-01-10
Point logrus logs to file

sebrandon1:
* Adds a new log file `cnf-certification-test/tnf-execution.log`.
* Deletes this log file at the beginning of every run (if it exists).
* Adds documentation and gitignore references to this new file.

### DIFF
--- a/cnf-certification-test/suite_test.go
+++ b/cnf-certification-test/suite_test.go
@@ -136,10 +136,9 @@ func TestTest(t *testing.T) {
 	}
 
 	defer f.Close()
-	log.SetOutput(f)
 
 	// Set hooks for logger
-	loghelper.SetHooks()
+	loghelper.SetHooks(f)
 
 	// Set logging level
 	loghelper.SetLogLevel()
--- a/cnf-certification-test/suite_test.go
+++ b/cnf-certification-test/suite_test.go
@@ -50,18 +50,22 @@ import (
 const (
 	claimFileName                 = "claim.json"
 	claimPathFlagKey              = "claimloc"
+	executionLogPathFlagKey       = "execlogloc"
 	CnfCertificationTestSuiteName = "CNF Certification Test Suite"
 	defaultClaimPath              = ".."
+	defaultExecutionLogPath       = ".."
 	defaultCliArgValue            = ""
 	junitFlagKey                  = "junit"
 	TNFJunitXMLFileName           = "cnf-certification-tests_junit.xml"
 	TNFReportKey                  = "cnf-certification-test"
 	extraInfoKey                  = "testsExtraInfo"
+	execLogFile                   = "tnf-execution.log"
 )
 
 var (
-	claimPath *string
-	junitPath *string
+	claimPath   *string
+	junitPath   *string
+	execLogPath *string
 	// GitCommit is the latest commit in the current git branch
 	GitCommit string
 	// GitRelease is the list of tags (if any) applied to the latest commit
@@ -80,6 +84,8 @@ func init() {
 		"the path where the claimfile will be output")
 	junitPath = flag.String(junitFlagKey, defaultCliArgValue,
 		"the path for the junit format report")
+	execLogPath = flag.String(executionLogPathFlagKey, defaultExecutionLogPath,
+		"the path for the execution log")
 }
 
 func getK8sClientsConfigFileNames() []string {
@@ -119,9 +125,8 @@ func TestTest(t *testing.T) {
 	loghelper.SetLogFormat()
 
 	// Delete execution log (if any) prior to run
-	logFile := "tnf-execution.log"
-	_, fileErr := os.Stat(logFile)
-	if os.IsExist(fileErr) {
+	logFile := filepath.Join(*execLogPath, execLogFile)
+	if _, fileErr := os.Stat(logFile); os.IsExist(fileErr) {
 		e := os.Remove(logFile)
 		if e != nil {
 			panic(e)
--- a/cnf-certification-test/suite_test.go
+++ b/cnf-certification-test/suite_test.go
@@ -82,20 +82,6 @@ func init() {
 		"the path for the junit format report")
 }
 
-// setLogLevel sets the log level for logrus based on the "TNF_LOG_LEVEL" environment variable
-func setLogLevel() {
-	params := configuration.GetTestParameters()
-
-	var logLevel, err = log.ParseLevel(params.LogLevel)
-	if err != nil {
-		log.Error("TNF_LOG_LEVEL environment set with an invalid value, defaulting to DEBUG \n Valid values are:  trace, debug, info, warn, error, fatal, panic")
-		logLevel = log.DebugLevel
-	}
-
-	log.Info("Log level set to: ", logLevel)
-	log.SetLevel(logLevel)
-}
-
 func getK8sClientsConfigFileNames() []string {
 	params := configuration.GetTestParameters()
 	fileNames := []string{}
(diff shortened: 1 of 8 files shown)


cnf-certification-test/suite_test.go
```diff
@@ -131,6 +131,27 @@ func TestTest(t *testing.T) {
...
+	}
+
+	// Setup logging to file
+	f, err := os.OpenFile(logFile, os.O_WRONLY|os.O_CREATE|os.O_APPEND, 0o644)
+	if err != nil {
+		log.Errorf("failed to create log file: %v", err)
+		panic(err)
+	}
+	defer f.Close()
+	log.SetOutput(f)
```
edcdavid, line 152: Isn't that removing the console logs and only leaving Ginkgo logs? It would be nice to preserve the console logs at info level and have more detailed file logs at debug or lower level. Logrus supports having several outputs natively with hooks. see https://github.com/sirupsen/logrus/tree/master/hooks/writer

sebrandon1: Whoops yes, adjusted to have the hooks like you suggested.

pkg/loghelper/loghelper.go
```diff
@@ -63,3 +66,36 @@ func SetLogFormat() {
...
+func SetLogLevel() {
+	params := configuration.GetTestParameters()
+
+	var logLevel, err = logrus.ParseLevel(params.LogLevel)
+	if err != nil {
+		logrus.Error("TNF_LOG_LEVEL environment set with an invalid value, defaulting to DEBUG \n Valid values are:  trace, debug, info, warn, error, fatal, panic")
+		logLevel = logrus.DebugLevel
+	}
+
+	logrus.Info("Log level set to: ", logLevel)
```
rdavid, line 99: Should be specifier in the format string?

cnf-certification-test/suite_test.go
```diff
@@ -168,7 +179,7 @@ func TestTest(t *testing.T) {
 
 	configurations, err := claimhelper.MarshalConfigurations()
 	if err != nil {
-		log.Errorf("Configuration node missing because of: %s", err)
+		log.Errorf("Configuration node missing because of: %v", err)
```
rdavid, line 182: Should the specifier be %w for error?

cnf-certification-test/suite_test.go
```diff
@@ -131,7 +117,32 @@ func TestTest(t *testing.T) {
 
 	// Set up logging params for logrus
 	loghelper.SetLogFormat()
-	setLogLevel()
+
+	// Delete execution log (if any) prior to run
+	logFile := "tnf-execution.log"
```
rdavid, line 122: if _, err := os.Stat("/path/to/whatever"); err == nil

See:
 https://stackoverflow.com/questions/12518876/how-to-check-if-a-file-exists-in-go

cnf-certification-test/suite_test.go
edcdavid: I would use the hooks to write all the log levels to file and leave the default mechanism we have now for console, something like:
https://github.com/sebrandon1/cnf-certification-test/pull/109

pkg/loghelper/loghelper.go
```diff
@@ -63,3 +66,36 @@ func SetLogFormat() {
...
+	})
+}
+
+// setLogLevel sets the log level for logrus based on the "TNF_LOG_LEVEL" environment variable
+func SetLogLevel() {
+	params := configuration.GetTestParameters()
+
+	var logLevel, err = logrus.ParseLevel(params.LogLevel)
+	if err != nil {
+		logrus.Error("TNF_LOG_LEVEL environment set with an invalid value, defaulting to DEBUG \n Valid values are:  trace, debug, info, warn, error, fatal, panic")
```
ramperher, line 95: I thinks there's no documentation currently for TNF_LOG_LEVEL variable, may we add it to the docs?
Also, if saying that "defaulting to DEBUG", it looks like "DEBUG" is a correct value but it is "debug", should it be lower case instead?

ramperher:
> Starting dci-check-change job.

I'll include this feature in dci-openshift-app-agent to use this log file, testing it in https://softwarefactory-project.io/r/c/dci-openshift-app-agent/+/27153, that's why you'll see some dcibot comments in the PR

ramperher:
From [this job](https://www.distributed-ci.io/jobs/8cfb0dd9-097b-40db-b1bf-0745eab1941b/jobStates?sort=date&task=55dfe816-4d25-4e00-8947-6bc9eeed11b3), ` "+ cp cnf-certification-test/tnf-execution.log /tmp/dci_logs.iuro56ok", "cp: cannot stat 'cnf-certification-test/tnf-execution.log': No such file or directory"`, `tnf-execution.log` file does not exist. I'm retrieving it from the same folder where `tnf_config.yaml` file is located.

cnf-certification-test/suite_test.go
sebrandon1: Merged.

cnf-certification-test/suite_test.go
sebrandon1: You only need %w if the calling function needs to "unwrap" the underlying error.  As far as I know we don't do any of that.

pkg/loghelper/loghelper.go
sebrandon1: Good question - this was in the old code I just moved it to the `loghelper` package.  I'll adjust it.

pkg/loghelper/loghelper.go
sebrandon1: Adjusted to be lowercase `debug` and added a note in the docs.

rdavid (approved):

sebrandon1:
@ramperher I wonder if it acts different when running in the container, maybe at the top level?  I'm looking into it.

ramperher:
> @ramperher I wonder if it acts different when running in the container, maybe at the top level? I'm looking into it.

Yes, it looks like it's not working in the desired way. May `tnf-execution.log` file be saved within the tnf container? Because, again, I've tested and `tnf-execution.log` is not found: https://www.distributed-ci.io/jobs/d231d1af-14d8-490f-88ea-3535e0b00462/files

pkg/loghelper/loghelper.go
```diff
@@ -63,3 +66,32 @@ func SetLogFormat() {
...
+}
+
+// setLogLevel sets the log level for logrus based on the "TNF_LOG_LEVEL" environment variable
+func SetLogLevel() {
+	params := configuration.GetTestParameters()
+
+	var logLevel, err = logrus.ParseLevel(params.LogLevel)
+	if err != nil {
+		logrus.Error("TNF_LOG_LEVEL environment set with an invalid value, defaulting to debug \n Valid values are: trace, debug, info, warn, error, fatal, panic")
+		logLevel = logrus.DebugLevel
```
greyerof, line 92: I know this is a copy/paste from the old code, but I have to remark this again, as this might be a good moment to change it: typically, in production software, the default log level for any application is (and should be) "info". Only when something went wrong, the user (probably a developer at this point) wants to see the "debug" output, so he/she will run again the test suite with the debug level set. If the debug output is not enough to fix the issue, we also have the "trace" level which should show everything else.

cnf-certification-test/suite_test.go
```diff
@@ -131,7 +117,30 @@ func TestTest(t *testing.T) {
...
```
greyerof, line 122: As Ramon said, this file won't exist outside the container, unless you create it in the same folder as the claim file (see var `claimPath ` in the `init()` function of this file), as that folder is mounted into a local folder in the host using the "-v param of docker/podman. I suggest (maybe in a further PR) changing/unifying those flags to a single "--output" or "--output-folder" and use it for all the output files: junit, claimfile and this new execution log file.

sebrandon1: Yup I think I've almost got it fixed.

pkg/loghelper/loghelper.go
sebrandon1: Done, changed to `info`.

cnf-certification-test/suite_test.go
sebrandon1: @ramperher Can you re-run it with my latest changes?

ramperher: now the file is captured :) you can see it here: https://www.distributed-ci.io/jobs/e51cb6a8-2465-45ef-bb14-a8917107d65a/files

ramperher:
I've retested again and now, I can see tnf-execution.log file submitted: https://www.distributed-ci.io/jobs/e51cb6a8-2465-45ef-bb14-a8917107d65a/files

However, if you compare this with our already generated execution.log file, I found differences between both of them. For example, our execution.log prints the results of each test that is executed, whereas tnf-execution.log doesn't. Is tnf-execution.log expected to only print what it's printing right now? Or should it look like execution.log (which, in the end, is the same that you'd see in the terminal)?

sebrandon1:
Closing this in favor of https://github.com/test-network-function/cnf-certification-test/pull/787

PR #784, opened 2023-01-13
access-control: add test case to check the use of CPU pools for Far Edge

jmontesi:


cnf-certification-test/accesscontrol/resources/resources.go
```diff
@@ -40,3 +40,38 @@ func HasRequestsAndLimitsSet(cut *provider.Container) bool {
 	}
 	return passed
 }
+
+// For more info on cpu mgmt polcies see https://kubernetes.io/docs/tasks/administer-cluster/cpu-management-policies/.
+func HasExclusiveCPUsAssigned(cut *provider.Container) bool {
```
edcdavid, line 45: This seems to be equivalent to pre-exisitng function: IsPodGuaranteed() / GetGuaranteedPods()

jmontesi: Yes, it's very similar but I need more granularity than Pod level because I need to determine if the containers inside a Pod have CPUs in both the exclusive and and the shared pool to fail the test. I understand that only using the shared CPU pool would be fine for this test.

With regards to the function you propose I think it has a misleading name because a Pod can belong to the QoS `Guaranteed` without having integer CPU units. See https://kubernetes.io/docs/tasks/administer-cluster/cpu-management-policies/. I have updated that function in https://github.com/test-network-function/cnf-certification-test/pull/790.

Also, reviewing the documentation I see that I need to check for memory limits as well in order for the Pod to qualify as `Guaranteed` Qos.

sebrandon1 (approved):

cnf-certification-test/accesscontrol/resources/resources.go
```diff
@@ -40,3 +40,38 @@ func HasRequestsAndLimitsSet(cut *provider.Container) bool {
...
+	cpuLimits := cut.Resources.Limits.Cpu()
+	cpuRequests := cut.Resources.Requests.Cpu()
```
bnshr, line 47: LGTM, can Line 47 be put later after Line 57, when you are checking on cpuRequests?

cnf-certification-test/accesscontrol/resources/resources_test.go
```diff
@@ -83,3 +83,78 @@ func TestHasRequestsAndLimitsSet(t *testing.T) {
...
+				},
+			},
+			expectedResult: false,
+		},
+		{ // Test Case #3 - Shared CPU pool selected (requests and limits quantities are not an integer)
+			testContainer: &provider.Container{
+				Container: &corev1.Container{
+					Resources: corev1.ResourceRequirements{
+						Requests: corev1.ResourceList{
+							"cpu": resource.MustParse("512m"),
```
bnshr: Is 512m treated as a fraction?

bnshr (changes requested):

cnf-certification-test/accesscontrol/resources/resources_test.go
bnshr: We can add one more test case with fraction value with limit 1.5

cnf-certification-test/accesscontrol/resources/resources.go
jmontesi: Done.

cnf-certification-test/accesscontrol/resources/resources_test.go
jmontesi: Yes, but it's a strange number for a CPU fraction, I probably was thinking of memory when I wrote that number. I changed it to 500m, which is equivalent to 0.5. I added a new unit test with 1.5 as you suggested in any case.

edcdavid (approved):

edcdavid:
needs rebase

cnf-certification-test/accesscontrol/resources/resources_test.go
```diff
@@ -83,3 +83,122 @@ func TestHasRequestsAndLimitsSet(t *testing.T) {
...
+		testContainer  *provider.Container
+		expectedResult bool
+	}{
+		{ // Test Case #1 - Exclusive CPU pool selected
+			testContainer: &provider.Container{
+				Container: &corev1.Container{
+					Resources: corev1.ResourceRequirements{
+						Requests: corev1.ResourceList{
+							"cpu":    resource.MustParse("2"),
+							"memory": resource.MustParse("512Mi"),
```
bnshr, line 99: I think some constants are defined but not used. Could you verify this please?

jmontesi: Done.

bnshr (approved):
lgtm!

Issue #762, opened 2023-01-05
Potentially uncordon nodes prior to starting pod-recreation tests


sebrandon1:
In-case of a "Ready,SchedulingDisabled" node, we might be encountering a leftover from a previous run.  Should we be checking for these nodes prior to the run so we can uncordon them properly?

Or maybe add an "After" section into the ginkgo that uncordons all nodes upon an unexpected exit/panic.

sebrandon1:
Also, we should FAIL the pod in recreation test if it has a node selector prior to running the cordon, delete, etc.

sebrandon1:
So far the only case we have seen where this test leaves nodes in a `SchedulingDisabled` state is when the test attempts to run against pods that have a `nodeSelector` or a node name attached to them and there are no other nodes to be scheduled to.  This leaves the pod under test in a `Pending` state and the test eventually times out.

#785 helps skip some of these pods that have node assignments.

sebrandon1:
We have not seen any more occurrences (at least reported) since #785.  Closing.

PR #785, opened 2023-01-13
Add check for nodeSelectors in pod-recreation test

sebrandon1:
Related to: https://github.com/test-network-function/cnf-certification-test/issues/762

The pod-recreation test case falls over because _sometimes_ there can be nodeSelectors attached to pods.  This is not allowed, per best practice requirement guidelines but that doesn't stop the test from attempting to run.  The test is now modified to check for these nodeSelectors prior to entering the loop that does the `cordon`.

Notable changes:
- Added `pod.HasNodeSelector()` func and appropriate changes.
- Added logic into the `pod-recreation` test to utilize the above `HasNodeSelector` func and detect any possible failures this test could encounter early.
- Changed `IsRuntimeClassNameSpecified` into a `Pod` struct func because it is used outside of the `isolation` file.

### DIFF
--- a/cnf-certification-test/lifecycle/suite.go
+++ b/cnf-certification-test/lifecycle/suite.go
@@ -452,9 +452,11 @@ func testPodsRecreation(env *provider.TestEnvironment) { //nolint:funlen
 	// Filter out pods with Node Assignments present and FAIL them.
 	// We run into problems with this test when there are nodeSelectors assigned affecting where
 	// pods are scheduled.  Also, they are not allowed in general, see the node-selector test case.
+	// Skip the safeguard for any pods that are using a runtimeClassName.  This is potentially
+	// because pods that are derived from a performance profile might have a built-in nodeSelector.
 	var podsWithNodeAssignement []*provider.Pod
 	for _, put := range env.Pods {
-		if put.HasNodeSelector() {
+		if !put.IsRuntimeClassNameSpecified() && put.HasNodeSelector() {
 			podsWithNodeAssignement = append(podsWithNodeAssignement, put)
 			logrus.Errorf("%s has been found with node selector(s): %v", put.String(), put.Spec.NodeSelector)
 		}
--- a/cnf-certification-test/lifecycle/suite.go
+++ b/cnf-certification-test/lifecycle/suite.go
@@ -280,8 +280,8 @@ func testPodsOwnerReference(env *provider.TestEnvironment) {
 func testPodNodeSelectorAndAffinityBestPractices(testPods []*provider.Pod) {
 	var badPods []*corev1.Pod
 	for _, put := range testPods {
-		if put.HasNodeAssignment() {
-			tnf.ClaimFilePrintf("ERROR: %s has a node assignment. Node selector: %v Node Name: %s", put, &put.Spec.NodeSelector, put.Spec.NodeName)
+		if put.HasNodeSelector() {
+			tnf.ClaimFilePrintf("ERROR: %s has a node selector. Node selector: %v", put, &put.Spec.NodeSelector)
 			badPods = append(badPods, put.Pod)
 		}
 		if put.Spec.Affinity != nil && put.Spec.Affinity.NodeAffinity != nil {
@@ -450,12 +450,13 @@ func testPodsRecreation(env *provider.TestEnvironment) { //nolint:funlen
 	}
 
 	// Filter out pods with Node Assignments present and FAIL them.
-	// We run into problems with this test when there are nodeSelectors or node names assigned affecting where
+	// We run into problems with this test when there are nodeSelectors assigned affecting where
 	// pods are scheduled.  Also, they are not allowed in general, see the node-selector test case.
 	var podsWithNodeAssignement []*provider.Pod
 	for _, put := range env.Pods {
-		if put.HasNodeAssignment() {
+		if put.HasNodeSelector() {
 			podsWithNodeAssignement = append(podsWithNodeAssignement, put)
+			logrus.Errorf("%s has been found with node selector(s): %v", put.String(), put.Spec.NodeSelector)
 		}
 	}
 	if len(podsWithNodeAssignement) > 0 {
--- a/cnf-certification-test/lifecycle/suite.go
+++ b/cnf-certification-test/lifecycle/suite.go
@@ -280,8 +280,8 @@ func testPodsOwnerReference(env *provider.TestEnvironment) {
 func testPodNodeSelectorAndAffinityBestPractices(testPods []*provider.Pod) {
 	var badPods []*corev1.Pod
 	for _, put := range testPods {
-		if len(put.Spec.NodeSelector) != 0 {
-			tnf.ClaimFilePrintf("ERROR: %s has a node selector clause. Node selector: %v", put, &put.Spec.NodeSelector)
+		if put.HasNodeAssignment() {
+			tnf.ClaimFilePrintf("ERROR: %s has a node assignment. Node selector: %v Node Name: %s", put, &put.Spec.NodeSelector, put.Spec.NodeName)
 			badPods = append(badPods, put.Pod)
 		}
 		if put.Spec.Affinity != nil && put.Spec.Affinity.NodeAffinity != nil {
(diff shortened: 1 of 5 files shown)


cnf-certification-test/lifecycle/suite.go
```diff
@@ -448,6 +448,21 @@ func testPodsRecreation(env *provider.TestEnvironment) { //nolint:funlen
 		tnf.ClaimFilePrintf("%s", claimsLog.GetLogLines())
 		ginkgo.Fail("Some deployments or stateful sets are not in a good initial state. Cannot perform test.")
 	}
+
+	// Filter out pods with NodeSelectors present and FAIL them.
+	// We run into problems with this test when there are nodeSelectors affecting where
+	// pods are scheduled.  Also, they are not allowed in general, see the node-selector test case.
+	var podsWithNodeSelectors []*provider.Pod
+	for _, put := range env.Pods {
```
rdavid, line 456: `put` for Putin? `p` or `pod` is better.

pkg/provider/pods.go
```diff
@@ -202,3 +202,7 @@ func (p *Pod) CreatedByDeploymentConfig() (bool, error) {
 	}
 	return false, nil
 }
+
+func (p *Pod) HasNodeSelector() bool {
+	return p.Spec.NodeSelector != nil
```
rdavid, line 207: Should we consider `nodeName` as well?
 https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/

rdavid (approved):

rdavid:
OMG, what is this `Operator hazelcast-platform-operator.v5.5.0 (channel alpha) failed to be certified for OpenShift` in smoke tests?

edcdavid (commented):
add note in test case description about failing this test when node selectors are present

edcdavid (approved):

cnf-certification-test/lifecycle/suite.go
sebrandon1: We use `put` everywhere for "Pod Under Test".

pkg/provider/pods.go
```diff
@@ -202,3 +202,7 @@ func (p *Pod) CreatedByDeploymentConfig() (bool, error) {
 	}
 	return false, nil
 }
+
+func (p *Pod) HasNodeSelector() bool {
```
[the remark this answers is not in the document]
sebrandon1: Sounds good I will add it there as well.

pkg/provider/pods.go
sebrandon1: Good call I'll add this to the func as well.

pkg/provider/pods.go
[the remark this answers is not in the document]
sebrandon1: Done.

pkg/provider/pods.go
```diff
@@ -202,3 +202,13 @@ func (p *Pod) CreatedByDeploymentConfig() (bool, error) {
 	}
 	return false, nil
 }
+
+func (p *Pod) HasNodeAssignment() bool {
```
rdavid, line 206: Is it not the same?

`return p.Spec.NodeSelector != nil || p.Spec.NodeName != ""`

sebrandon1: You're right!  I'll adjust.

rdavid (approved):

ramperher:
Hi @sebrandon1, the DCI job is failing in the `lifecycle-pod-scheduling` and `lifecycle-pod-recreation` tests (I think the second one is the target of this change, but it's also affecting to the first one: https://www.distributed-ci.io/jobs/d134c011-75df-4c15-bc50-f585bfd0727e/tests/5ccaae58-cb85-412e-98d9-1f988ae2fc88. I think it's failing because the pods have a nodeName different than empty.

But, if you see pods specification, which is [here](https://github.com/redhat-cip/dci-openshift-app-agent/blob/master/samples/tnf_test_example/hooks/templates/test_pods.yml.j2), you'll see we're not explicitly defining any nodeSelector or nodeName. We just have a labelSelector under a podAntiAffinity rule (that I remember that we had to include it in order to pass the recreation test, as the condition of nodeSelectors was added at that time). So, we are not hardcoding the node to use in the config (this is what I would consider as failure in this test), but Kubernetes is deciding to assign a nodeName based on the labelSelector we have defined.

So, having said that, I don't know how to pass the test with the definition we have, also knowing that we have this kind of definition with labelSelector in order to test as many tnf tests as possible. What can we do?

edcdavid:
> @ramperher Maybe I take the `node name` check out of this, and only check for the node selector. We can always add/refactor this in the future if we need to specifically node names.
> 
> @edcdavid what do you think?
I agree that we should remove it for now. We need a logic that is able to check whether the nodename field was updated by kubernetes or by a user level operation. Wecould investigate this in a later PR.

sebrandon1:
I have removed the check for the populated `node name` field as it's not entirely straightforward about how this field got set.  Since we've only ever encountered a partner manually entering node selectors we'll just leave it like this for now and deal with potential node name field changes in a subsequent PR if necessary.

sebrandon1:
@ramperher I'm just wondering how the pod specs end up with `node-role.kubernetes.io/worker` in their node selectors.  That's what seems to be failing now.

```
[31mERROR  [0m[Jan 17 20:09:23.394][suite.go: 459] pod: test-0 ns: production-cnf has been found with node selector(s): map[node-role.kubernetes.io/worker:] 
[31mERROR  [0m[Jan 17 20:09:23.394][suite.go: 459] pod: test-1 ns: production-cnf has been found with node selector(s): map[node-role.kubernetes.io/worker:] 
```

In the partner repo, we label all of the workers with that label but I don't know how it translates itself over to the pods?
https://github.com/test-network-function/cnf-certification-test-partner/blob/main/scripts/fix-node-labels.sh

ramperher:
@sebrandon1, I think I've found something.

I've deployed the same workload in two different scenarios: 1) our classical disconnected environment where all the checks for your PR are always run, and where the DCI job is failing, and 2) a connected environment with a virtualized OCP cluster.

In 2), if you see [this job](https://www.distributed-ci.io/jobs/0ca0d849-3b01-499a-a16a-d89ff3384a16/jobStates), which is reported in the comment above, the job is **working** with your patch; `lifecycle-pod-recreation` test is passing.

However, in 1), as we saw, it is not working: `production-cnf` pods have this strange `node-role.kubernetes.io/worker:` label.

The truth is, there are differences between the two scenarios:

- in 1), we have runtimeClass resources created, so that [these tasks retrieves the runtimeClass](https://github.com/redhat-cip/dci-openshift-app-agent/blob/master/samples/tnf_test_example/hooks/install.yml#L2-L11) and the pods in `production-cnf` namespace [include the reference](https://github.com/redhat-cip/dci-openshift-app-agent/blob/master/samples/tnf_test_example/hooks/templates/test_pods.yml.j2#L47-L49) to that runtimeClass in order to pass `lifecycle-cpu-isolation` test.
- in 2), we don't have any runtimeClass created, so that the pods in `production-cnf` namespace do not include the reference to that, and that's why `lifecycle-cpu-isolation` test is failing in that deployment (you can see the job related to 2) to confirm this).

Some logs here:

1) `production-cnf` pods have runtimeClass reference and have the nodeSelector.

```
$ oc get pods -n production-cnf -o json | grep node-role
                    "node-role.kubernetes.io/worker": ""
                    "node-role.kubernetes.io/worker": ""
$ oc get pods -n production-cnf -o json | grep runtimeClass
                "runtimeClassName": "performance-cnf-basic-profile",
                "runtimeClassName": "performance-cnf-basic-profile",

```

2) `production-cnf` pods do not have runtimeClass reference and do not have the nodeSelector.

```
$ oc get pods -n production-cnf -o json | grep runtimeClass
# empty
$ oc get pods -n production-cnf -o json | grep node-role
# empty
```

So, my impression is that `runtimeClass` reference is involved in the creation of the nodeSelector. However, if we remove that reference to pass `lifecycle-pod-recreation` test, then `lifecycle-cpu-isolation` test will fail in the target pods.

ramperher:
I've created this change in our side to use a simpler runtimeClass that does not include the nodeSelector clause for sure: https://softwarefactory-project.io/r/c/dci-openshift-app-agent/+/27218. I've included this as a dependency in this change to run the DCI tests. Now it should pass, I'm re-running

sebrandon1:
@ramperher For pods that are using runtimeClasses, should I just skip enforcing the check for nodeSelectors?  I could just look at all pods that aren't using runtimeClasses for the safeguard check.

cnf-certification-test/lifecycle/suite.go
```diff
@@ -448,6 +448,24 @@ func testPodsRecreation(env *provider.TestEnvironment) { //nolint:funlen
...
 	}
+
+	// Filter out pods with Node Assignments present and FAIL them.
+	// We run into problems with this test when there are nodeSelectors assigned affecting where
+	// pods are scheduled.  Also, they are not allowed in general, see the node-selector test case.
+	// Skip the safeguard for any pods that are using a runtimeClassName.  This is potentially
+	// because pods that are derived from a performance profile might have a built-in nodeSelector.
+	var podsWithNodeAssignement []*provider.Pod
+	for _, put := range env.Pods {
+		if !put.IsRuntimeClassNameSpecified() && put.HasNodeSelector() {
```
sebrandon1, line 459: @ramperher See this proposed change.  I only error out early if there are no runtimeClasses specified.  Does that satisfy some of the issues we are seeing?

ramperher: It makes sense for me, let's try it on DCI! (I'll remove the dependency with my change, so that we can test the current deployment we're doing)

ramperher: It makes sense for me, let's try it on DCI! (I'll remove the dependency with my change, so that we can test the current deployment we're doing)

ramperher (approved):
@sebrandon1 , confirmed it's working fine with your latest patch, thanks for taking care of my suggestions! :)

PR #786, opened 2023-01-13
Added qe variable to catalog entries

theemj:
Also added missing value to storage-required-pods entry, re-generated CATALOG.md

### DIFF
--- a/cnf-certification-test/identifiers/identifiers.go
+++ b/cnf-certification-test/identifiers/identifiers.go
@@ -210,6 +210,7 @@ The label value is not important, only its presence.`,
 		NoDocumentedProcess,
 		VersionOne,
 		bestPracticeDocV1dot4URL, // TODO: link Far Edge spec document
+		false,
 		TagFarEdge)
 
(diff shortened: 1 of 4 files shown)


cnf-certification-test/identifiers/identifiers.go
```diff
@@ -218,8 +229,10 @@ The label value is not important, only its presence.`,
 		`Checks that pods do not place persistent volumes on local storage.`,
 		StorageRequiredPods,
 		InformativeResult,
+		NoDocumentedProcess,
```
theemj, line 232: This entry did not previously have an exception string (if you look at CATALOG.md on main all of the values from here on are off by one). If there is actually an exception process just let me know and I'll add it.

sebrandon1: Good catch!

sebrandon1 (approved):

edcdavid (commented):
Fixing the AddCatalogEntry call in https://github.com/test-network-function/cnfextensions is also needed

edcdavid (approved):

PR #787, opened 2023-01-13
Create log file by mirroring output in bash script

edcdavid:
This is related to https://github.com/test-network-function/cnf-certification-test/pull/777
This is proposing to use tee to mirror the output and captures all the logs(including those added by Ginkgo) in the cli and container scenarios

### DIFF
--- a/run-cnf-suites.sh
+++ b/run-cnf-suites.sh
@@ -146,6 +146,12 @@ fi
 
 cd ./cnf-certification-test || exit 1
 
+# configuring special pipeline mode
+# The exit status of a pipeline is the exit status of the last command in the pipeline, unless the pipefail option
+# is enabled (see The Set Builtin). If pipefail is enabled, the pipeline's return status is the value of the last (rightmost) 
+# command to exit with a non-zero status, or zero if all commands exit successfully.
+set -o pipefail
+
 # Do not double quote.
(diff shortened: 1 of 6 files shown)


run-cnf-suites.sh
```diff
@@ -162,4 +162,7 @@ cd ./cnf-certification-test || exit 1
 	${FOCUS_STRING} \
 	${SKIP_STRING} \
 	"${LABEL_STRING}" \
-	${GINKGO_ARGS}
+	${GINKGO_ARGS} |& tee $OUTPUT_LOC/tnf-execution.log
+
+# preserving the exit status
+exit ${PIPESTATUS[0]}
```
rdavid, line 168: Add the new line at the end. I think your editor doesn't add the new line automatically. What do you use? As far as I know, all *nix text editors do it.

rdavid, line 168: `set -o pipefail` is a more general solution, because it fails if any command fails, `tee` could fail on permission, full disk, etc.

rdavid (approved):

ramperher (approved):
Thanks! Now the output in tnf-execution.log is the expected

sebrandon1 (approved):

run-cnf-suites.sh
edcdavid: fixed

PR #797, opened 2023-01-19
VCP CNF Requirement: PODs must NOT place persistent volumes on local storage

shimritproj:
build-depends: https://github.com/dci-labs/dallas-pipelines/pull/646

VCP CNF Requirement: PODs must NOT place persistent volumes on local storage.

We following pod by pod and check:
1) if PVC is not exist : we don't have what to check and we finished the test.
2) if name of PVC is equal to PVC of the pod:
-we check if provisioner of StorageClasees equal to"kubernetes.io/no-provisioner":
if yes, we found pod that use a local storage enabled storageClass and we add to list of pods that have local storage("podWithLocalStorage")

we check all volumes in all pods under test to see if they are used. If a pod is using a volume that is using a local StorageClass, we should fail the test.

The purpose is to find pods that have PVC&&PV and to use a local storage enabled StorageClasses.

### DIFF
--- a/CATALOG.md
+++ b/CATALOG.md
@@ -503,6 +503,20 @@ Suggested Remediation|Add a liveness probe to deployed containers
 Best Practice Reference|https://connect.redhat.com/sites/default/files/2022-05/Cloud%20Native%20Network%20Function%20Requirements%201-3.pdf Section 5.2.16, 12.1 and 12.5
 Exception Process|There is no documented exception process for this.
 Tags|common,telco
+#### no-pvs-on-localstorage
+
+Property|Description
+---|---
+Test Case Name|no-pvs-on-localstorage
+Test Case Label|lifecycle-no-pvs-on-localstorage
+Unique ID|http://test-network-function.com/testcases/lifecycle/no-pvs-on-localstorage
+Version|v1.0.0
+Description|http://test-network-function.com/testcases/lifecycle/no-pvs-on-localstorage Checks that pods do not place persistent volumes on local storage.
+Result Type|informative
+Suggested Remediation|If the kind of pods is StatefulSet, so we need to make sure that servicename is not local-storage.
+Best Practice Reference|https://TODO Section 4.6.24
+Exception Process|There is no documented exception process for this.
+Tags|extended
 #### persistent-volume-reclaim-policy
 
 Property|Description
--- a/cnf-certification-test/identifiers/identifiers.go
+++ b/cnf-certification-test/identifiers/identifiers.go
@@ -255,6 +255,18 @@ The label value is not important, only its presence.`,
 		false,
 		TagExtended)
 
+	TestStorageRequiredPods = AddCatalogEntry(
+		"no-pvs-on-localstorage",
+		common.LifecycleTestKey,
+		`Checks that pods do not place persistent volumes on local storage.`,
+		StorageRequiredPods,
+		InformativeResult,
+		NoDocumentedProcess,
+		VersionOne,
+		bestPracticeDocV1dot4URL+" Section 4.6.24",
+		false,
+		TagExtended)
+
 	TestStartupIdentifier = AddCatalogEntry(
 		"container-startup",
 		common.LifecycleTestKey,
--- a/cnf-certification-test/lifecycle/suite.go
+++ b/cnf-certification-test/lifecycle/suite.go
@@ -17,7 +17,6 @@
 package lifecycle
 
 import (
-	"context"
 	"fmt"
 	"time"
 
@@ -32,14 +31,12 @@ import (
 	"github.com/test-network-function/cnf-certification-test/cnf-certification-test/lifecycle/tolerations"
 	"github.com/test-network-function/cnf-certification-test/cnf-certification-test/lifecycle/volumes"
 	"github.com/test-network-function/cnf-certification-test/cnf-certification-test/results"
-	"github.com/test-network-function/cnf-certification-test/internal/clientsholder"
 	"github.com/test-network-function/cnf-certification-test/pkg/configuration"
 	"github.com/test-network-function/cnf-certification-test/pkg/postmortem"
 	"github.com/test-network-function/cnf-certification-test/pkg/provider"
 	"github.com/test-network-function/cnf-certification-test/pkg/testhelper"
 	"github.com/test-network-function/cnf-certification-test/pkg/tnf"
 	corev1 "k8s.io/api/core/v1"
-	apiv1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 const (
@@ -584,23 +581,36 @@ func TestPodTolerationBypass(env *provider.TestEnvironment) {
 	testhelper.AddTestResultLog("Non-compliant", podsWithRestrictedTolerationsNotDefault, tnf.ClaimFilePrintf, ginkgo.Fail)
 }
 func testStorageRequiredPods(env *provider.TestEnvironment) {
-	oc := clientsholder.GetClientsHolder()
-	var nonCompliantPods []*provider.Pod
-	for _, pod := range env.Pods {
-		for _, podOwner := range pod.ObjectMeta.GetOwnerReferences() {
-			if podOwner.Kind != statefulSet {
+	var podsWithLocalStorage []string
+	var StorageClasses = env.StorageClasses
+	var Pvc = env.PersistentVolumeClaims
+	for _, put := range env.Pods {
+		for pvIndex := range put.Spec.Volumes {
+			// Skip any nil persistentClaims.
+			volume := put.Spec.Volumes[pvIndex]
+			if volume.PersistentVolumeClaim == nil {
 				continue
 			}
-			statefulset, err := oc.K8sClient.AppsV1().StatefulSets(pod.Namespace).Get(context.TODO(), podOwner.Name, apiv1.GetOptions{})
-			if err != nil {
-				tnf.ClaimFilePrintf(err.Error())
-				continue
-			}
-			if statefulset.Spec.ServiceName == localStorage {
-				nonCompliantPods = append(nonCompliantPods, pod)
-				continue
+			// We have the list of pods/volumes/claims.
+			// Look through the storageClass list for a match.
+			for i := range Pvc {
+				if Pvc[i].Name == put.Spec.Volumes[pvIndex].PersistentVolumeClaim.ClaimName {
+					for j := range StorageClasses {
+						if Pvc[i].Spec.StorageClassName != nil && StorageClasses[j].Name == *Pvc[i].Spec.StorageClassName {
+							tnf.ClaimFilePrintf("%s has been found to use a local storage enabled storageClass.\n Pvc_name: %s, Storageclass_name : %s, Provisionner_name: %s", put.String(), put.Spec.Volumes[pvIndex].PersistentVolumeClaim.ClaimName,
+								StorageClasses[j].Name, StorageClasses[j].Provisioner)
+							podsWithLocalStorage = append(podsWithLocalStorage, put.String())
+							break
+						}
+						tnf.ClaimFilePrintf("%s has been not found to use a local storage enabled storageClass.\n Pvc_name: %s, Storageclass_name : %s, Provisionner_name: %s", put.String(), put.Spec.Volumes[pvIndex].PersistentVolumeClaim.ClaimName,
+							StorageClasses[j].Name, StorageClasses[j].Provisioner)
+					}
+				}
 			}
 		}
 	}
-	testhelper.AddTestResultLog("Non-compliant", nonCompliantPods, tnf.ClaimFilePrintf, ginkgo.Fail)
+
+	if n := len(podsWithLocalStorage); n > 0 {
+		testhelper.AddTestResultLog("Non-compliant", podsWithLocalStorage, tnf.ClaimFilePrintf, ginkgo.Fail)
+	}
 }
--- a/cnf-certification-test/lifecycle/suite.go
+++ b/cnf-certification-test/lifecycle/suite.go
@@ -582,7 +582,7 @@ func TestPodTolerationBypass(env *provider.TestEnvironment) {
 }
 func testStorageRequiredPods(env *provider.TestEnvironment) {
 	var podsWithLocalStorage []string
-	var StorageClasses = env.StorageClasses
+	var StorageClasses = env.StorageList
 	var Pvc = env.PersistentVolumeClaims
 	for _, put := range env.Pods {
 		for pvIndex := range put.Spec.Volumes {
--- a/cnf-certification-test/lifecycle/suite.go
+++ b/cnf-certification-test/lifecycle/suite.go
@@ -582,7 +582,7 @@ func TestPodTolerationBypass(env *provider.TestEnvironment) {
 }
 func testStorageRequiredPods(env *provider.TestEnvironment) {
 	var podsWithLocalStorage []string
-	var StorageClasses = env.StorageList
+	var StorageClasses = env.StorageClassList
 	var Pvc = env.PersistentVolumeClaims
 	for _, put := range env.Pods {
 		for pvIndex := range put.Spec.Volumes {
@@ -597,12 +597,12 @@ func testStorageRequiredPods(env *provider.TestEnvironment) {
 				if Pvc[i].Name == put.Spec.Volumes[pvIndex].PersistentVolumeClaim.ClaimName && Pvc[i].Namespace == put.Namespace {
 					for j := range StorageClasses {
 						if Pvc[i].Spec.StorageClassName != nil && StorageClasses[j].Name == *Pvc[i].Spec.StorageClassName {
-							tnf.ClaimFilePrintf("%s has been found to use a local storage enabled storageClass.\n Pvc_name: %s, Storageclass_name : %s, Provisionner_name: %s", put.String(), put.Spec.Volumes[pvIndex].PersistentVolumeClaim.ClaimName,
+							tnf.ClaimFilePrintf("%s has been found to use a local storage enabled storageClass.\n Pvc_name: %s, Storageclass_name : %s, Provisioner_name: %s", put.String(), put.Spec.Volumes[pvIndex].PersistentVolumeClaim.ClaimName,
 								StorageClasses[j].Name, StorageClasses[j].Provisioner)
 							podsWithLocalStorage = append(podsWithLocalStorage, put.String())
 							break
 						}
-						tnf.ClaimFilePrintf("%s has been not found to use a local storage enabled storageClass.\n Pvc_name: %s, Storageclass_name : %s, Provisionner_name: %s", put.String(), put.Spec.Volumes[pvIndex].PersistentVolumeClaim.ClaimName,
+						tnf.ClaimFilePrintf("%s has been not found to use a local storage enabled storageClass.\n Pvc_name: %s, Storageclass_name : %s, Provisioner_name: %s", put.String(), put.Spec.Volumes[pvIndex].PersistentVolumeClaim.ClaimName,
 							StorageClasses[j].Name, StorageClasses[j].Provisioner)
 					}
 				}
--- a/cnf-certification-test/lifecycle/suite.go
+++ b/cnf-certification-test/lifecycle/suite.go
@@ -594,7 +594,7 @@ func testStorageRequiredPods(env *provider.TestEnvironment) {
 			// We have the list of pods/volumes/claims.
 			// Look through the storageClass list for a match.
 			for i := range Pvc {
-				if Pvc[i].Name == put.Spec.Volumes[pvIndex].PersistentVolumeClaim.ClaimName {
+				if Pvc[i].Name == put.Spec.Volumes[pvIndex].PersistentVolumeClaim.ClaimName && Pvc[i].Namespace == put.Namespace {
 					for j := range StorageClasses {
 						if Pvc[i].Spec.StorageClassName != nil && StorageClasses[j].Name == *Pvc[i].Spec.StorageClassName {
 							tnf.ClaimFilePrintf("%s has been found to use a local storage enabled storageClass.\n Pvc_name: %s, Storageclass_name : %s, Provisionner_name: %s", put.String(), put.Spec.Volumes[pvIndex].PersistentVolumeClaim.ClaimName,
--- a/pkg/autodiscover/autodiscover.go
+++ b/pkg/autodiscover/autodiscover.go
@@ -35,6 +35,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	networkingv1 "k8s.io/api/networking/v1"
 	policyv1 "k8s.io/api/policy/v1"
+	storagev1 "k8s.io/api/storage/v1"
 	apiextv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -83,6 +84,7 @@ type DiscoveredTestData struct {
 	Nodes                  *corev1.NodeList
 	Istio                  bool
 	ValidProtocolNames     []string
+	StorageClasses         []storagev1.StorageClass
 	ServicesIgnoreList     []string
 }
 
@@ -120,6 +122,10 @@ func DoAutoDiscover() DiscoveredTestData {
 		logrus.Fatalf("Cannot load configuration, error: %v", err)
 	}
 	oc := clientsholder.GetClientsHolder()
+	data.StorageClasses, err = getAllStorageClasses()
+	if err != nil {
+		logrus.Fatalf("Failed to retrieve storageClasses - err: %v", err)
+	}
 	data.AllNamespaces, _ = getAllNamespaces(oc.K8sClient.CoreV1())
 	data.AllSubscriptions = findSubscriptions(oc.OlmClient, []string{""})
 	data.AllCsvs = getAllOperators(oc.OlmClient)
--- a/pkg/autodiscover/autodiscover_pv.go
+++ b/pkg/autodiscover/autodiscover_pv.go
@@ -19,7 +19,10 @@ package autodiscover
 import (
 	"context"
 
+	"github.com/sirupsen/logrus"
+	"github.com/test-network-function/cnf-certification-test/internal/clientsholder"
 	corev1 "k8s.io/api/core/v1"
+	storagev1 "k8s.io/api/storage/v1"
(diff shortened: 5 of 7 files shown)


pvs.yaml
```diff
@@ -0,0 +1,55 @@
+apiVersion: v1
```
sebrandon1, line 1: Not needed to be checked in.

pkg/autodiscover/autodiscover_operators.go
```diff
@@ -68,6 +69,19 @@ func findOperatorsByLabel(olmClient clientOlm.Interface, labels []configuration.
 
 	return csvs
 }
+func getAllStorageData(oc corev1client.CoreV1Interface) (data []string, err error) {
```
sebrandon1, line 72: Move this function to https://github.com/test-network-function/cnf-certification-test/blob/main/pkg/autodiscover/autodiscover_pv.go.

This will help us organize the storage related funcs.

CATALOG.md
```diff
@@ -516,7 +530,36 @@ Result Type|normative
 Suggested Remediation|Add a liveness probe to deployed containers
 Best Practice Reference|https://connect.redhat.com/sites/default/files/2022-05/Cloud%20Native%20Network%20Function%20Requirements%201-3.pdf Section 5.2.16, 12.1 and 12.5
 Exception Process|There is no documented exception process for this.
-Tags|common,telco
+Tags|common
+#### net-admin-capability-check
```
sebrandon1, line 534: I think you need to rebase then re-run `make build-catalog-md` to clear these up.

cnf-certification-test/storageclass.yml
```diff
@@ -0,0 +1,6 @@
+apiVersion: storage.k8s.io/v1
```
sebrandon1, line 1: Not needed to be checked in.

sebrandon1 (changes requested):
Just some initial comments.

pkg/autodiscover/autodiscover_operators.go
```diff
@@ -68,6 +69,19 @@ func findOperatorsByLabel(olmClient clientOlm.Interface, labels []configuration.
 
 	return csvs
 }
+func getAllStorageData() (data []string, err error) {
```
sebrandon1, line 72: Move this function to https://github.com/test-network-function/cnf-certification-test/blob/main/pkg/autodiscover/autodiscover_pv.go.

This will help us organize the storage related funcs.

pkg/autodiscover/autodiscover_pv.go
```diff
@@ -39,3 +42,17 @@ func getPersistentVolumeClaims(oc corev1client.CoreV1Interface) ([]corev1.Persis
 	}
 	return pvcs.Items, nil
 }
+
+func getAllStorageData() (data []string, err error) {
```
sebrandon1, line 46: `err` is not handled in the calling func.

pkg/autodiscover/autodiscover_pv.go
```diff
@@ -39,3 +42,17 @@ func getPersistentVolumeClaims(oc corev1client.CoreV1Interface) ([]corev1.Persis
...
+	o := clientsholder.GetClientsHolder()
+	storageclasslist, err := o.K8sClient.StorageV1().StorageClasses().List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		logrus.Errorln("Error when listing", "err: ", err)
+		return data, fmt.Errorf("error getting all storage list, err: %s", err)
+	}
+	fmt.Println(storageclasslist)
```
sebrandon1, line 53: I think this `Println` can be removed.

pkg/autodiscover/autodiscover_pv.go
```diff
@@ -39,3 +42,17 @@ func getPersistentVolumeClaims(oc corev1client.CoreV1Interface) ([]corev1.Persis
 	}
 	return pvcs.Items, nil
 }
+
+func getAllStorageData() (data []string, err error) {
+	o := clientsholder.GetClientsHolder()
+	storageclasslist, err := o.K8sClient.StorageV1().StorageClasses().List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		logrus.Errorln("Error when listing", "err: ", err)
+		return data, fmt.Errorf("error getting all storage list, err: %s", err)
```
sebrandon1, line 51:
```suggestion
		return data, fmt.Errorf("error retrieving storageClass list, err: %s", err)
```

cnf-certification-test/lifecycle/suite.go
```diff
@@ -584,23 +582,25 @@ func TestPodTolerationBypass(env *provider.TestEnvironment) {
...
-				continue
-			}
-			statefulset, err := oc.K8sClient.AppsV1().StatefulSets(pod.Namespace).Get(context.TODO(), podOwner.Name, apiv1.GetOptions{})
-			if err != nil {
-				tnf.ClaimFilePrintf(err.Error())
+	var persistentVolumesBad []string
+	if env.StorageList[index] == "kubernetes.io/no-provisioner" {
+		tnf.ClaimFilePrintf("The class storage is using local storage")
+		return
+	}
```
sebrandon1, line 589: I'm not sure I understand this section of code, is this just a leftover from debugging?

cnf-certification-test/lifecycle/suite.go
```diff
@@ -32,21 +31,20 @@ import (
...
 	corev1 "k8s.io/api/core/v1"
-	apiv1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 const (
 	timeout                    = 300 * time.Second
 	timeoutPodRecreationPerPod = time.Minute
 	timeoutPodSetReady         = 7 * time.Minute
 	minWorkerNodesForLifecycle = 2
+	index                      = 0
```
sebrandon1, line 47: I think this is also a leftover.

sebrandon1 (changes requested):
Some more comments to review!  Thanks!

cnf-certification-test/lifecycle/suite.go
```diff
@@ -584,23 +582,25 @@ func TestPodTolerationBypass(env *provider.TestEnvironment) {
...
+	for _, put := range env.Pods {
+		for pvIndex := range put.Spec.Volumes {
+			if put.Spec.Volumes[pvIndex].PersistentVolumeClaim == nil {
 				continue
 			}
-			if statefulset.Spec.ServiceName == localStorage {
-				nonCompliantPods = append(nonCompliantPods, pod)
-				continue
+			for i := range env.PersistentVolumes {
+				if env.PersistentVolumes[i].Spec.StorageClassName == localStorage {
```
greyerof, line 597: A couple of issues here:
1. I don't see any relationship here between the `env.PersistentVolumes[i]` and the pod you're checking ("put") on each iteration. Could you explain why are you checking this, please?
2. StorageClasses, as all kubernetes resources, can have any name. Why are you specifically checking against "local-storage"?

pkg/autodiscover/autodiscover_pv.go
greyerof, line 46: "Data" suffix is too generic here. My suggestion: `func getStorageClassProvisioners() ([]string, error)`

pkg/autodiscover/autodiscover.go
```diff
@@ -83,6 +83,7 @@ type DiscoveredTestData struct {
 	Nodes                  *corev1.NodeList
 	Istio                  bool
 	ValidProtocolNames     []string
+	Storage                []string
```
greyerof, line 86: Storage -> StorageClassProvisioners

cnf-certification-test/identifiers/identifiers.go
```diff
@@ -250,6 +250,18 @@ The label value is not important, only its presence.`,
 		false,
 		TagExtended)
 
+	TestStorageRequiredPods = AddCatalogEntry(
+		"storage-required-pods",
```
greyerof, line 254: "storage-required-pods" seems weird to me for this test case. What about "no-pvs-on-localstorage".

pkg/autodiscover/autodiscover_pv.go
shimritproj: I think we can leave it as it is.

cnf-certification-test/lifecycle/suite.go
shimritproj: In the begining we check if a storageclass is using local storage, if not we print message and thats it , if yes- we check pod by pod. 
I need to change the code to be like this - 
	if env.StorageList[index] != "kubernetes.io/no-provisioner" {
		tnf.ClaimFilePrintf("The class storage not use in local storage")
		return
	}

cnf-certification-test/lifecycle/suite.go
shimritproj: If local storage exist so I need to check that PVC and also the PV is exist.

In addition, when I did debug I saw the name is local-storage.

@greyerof

pkg/autodiscover/autodiscover_pv.go
shimritproj: you're right! I fixed it.

sebrandon1: I agree with @greyerof I think this could be more specific to the objects it is retrieving.

shimritproj: I changed it. @sebrandon1 @greyerof

cnf-certification-test/lifecycle/suite.go
```diff
@@ -584,23 +581,30 @@ func TestPodTolerationBypass(env *provider.TestEnvironment) {
...
+			// Look through the storageClass list for a match.
+			for _, s := range env.StorageList {
+				if s.Name == put.Spec.Volumes[pvIndex].PersistentVolumeClaim.ClaimName  {
+					// At this point we have found the storage class attached to the pod.
+					// We need to check if this storageClass is no-provisioner.
+					if s.Provisioner == "kubernetes.io/no-provisioner" {
+						tnf.ClaimFilePrintf("This is a bad pod: The pod using in local storage")
+						podsWithLocalStorage = append(podsWithLocalStorage, put.String())
+						// Break out of the storageList loop.
+						continue
```
sebrandon1, line 601: Looking at this again, I think we should keep a "localStorageFound" boolean per pod and then just only do `podsWithLocalStorage = append(podsWithLocalStorage, put.String())` once per pod.

cnf-certification-test/lifecycle/suite.go
```diff
@@ -584,23 +581,30 @@ func TestPodTolerationBypass(env *provider.TestEnvironment) {
...
-				nonCompliantPods = append(nonCompliantPods, pod)
-				continue
+			// We have the list of pods/volumes/claims.
+			// Look through the storageClass list for a match.
+			for _, s := range env.StorageList {
+				if s.Name == put.Spec.Volumes[pvIndex].PersistentVolumeClaim.ClaimName  {
+					// At this point we have found the storage class attached to the pod.
+					// We need to check if this storageClass is no-provisioner.
+					if s.Provisioner == "kubernetes.io/no-provisioner" {
+						tnf.ClaimFilePrintf("This is a bad pod: The pod using in local storage")
```
sebrandon1, line 598:
```suggestion
						tnf.ClaimFilePrintf("%s has been found to use a local storage enabled storageClass", put.String())
```

pkg/provider/provider.go
```diff
@@ -38,6 +38,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	networkingv1 "k8s.io/api/networking/v1"
 	policyv1 "k8s.io/api/policy/v1"
+	v1 "k8s.io/api/storage/v1"
```
sebrandon1, line 41: Whoops, we should change this to `storagev1`.

pkg/provider/provider.go
```diff
@@ -219,6 +221,7 @@ func buildTestEnvironment() { //nolint:funlen
 		env.StatefulSets = append(env.StatefulSets, aNewStatefulSet)
 	}
 	env.HorizontalScaler = data.Hpas
+	env.StorageList = data.StorageClasses
```
edcdavid, line 224: rename StorageList with StorageClasses

cnf-certification-test/lifecycle/suite.go
shimritproj: We can save map that we have details on the pod and also the boolean variable "localStorageFound" and we save it global and then we have information about every pod.
and if we doing like this so, i think that we don't need the message in line 598.

What do you think? @sebrandon1

sebrandon1: I think the `continue` with the `boolean` approach will have less iterations when looping instead of just potentially adding the same pod to the map when/if the first volume fails.

We should keep the message for context for the user.

cnf-certification-test/lifecycle/suite.go
```diff
@@ -584,23 +581,30 @@ func TestPodTolerationBypass(env *provider.TestEnvironment) {
...
+			if put.Spec.Volumes[pvIndex].PersistentVolumeClaim == nil {
 				continue
 			}
-			if statefulset.Spec.ServiceName == localStorage {
-				nonCompliantPods = append(nonCompliantPods, pod)
-				continue
+			// We have the list of pods/volumes/claims.
+			// Look through the storageClass list for a match.
+			for _, s := range env.StorageClasses {
+				if s.Name == put.Spec.Volumes[pvIndex].PersistentVolumeClaim.ClaimName {
```
edcdavid, line 594: you are comparing the storage class name with the PVC name  these will not match. First, you need to retrieve the PVC object corresponding to put.Spec.Volumes[pvIndex].PersistentVolumeClaim.ClaimName and in the PVC you will be able to see the storage class. Looks like we already get the PVCs in env.PersistentVolumeClaims so you can get the PVC object from there

edcdavid (commented):
You should also add a bad test case yaml example to test both passing and failing pods in the partner repo /examples directory

edcdavid (commented):
add a more detailed description for the new test case to describe how it is tested

cnf-certification-test/lifecycle/suite.go
shimritproj: > You should also add a bad test case yaml example to test both passing and failing pods in the partner repo /examples directory

I will do it in another PR (on repo partner)

shimritproj: The changes have been uploaded @sebrandon1 ֿ

cnf-certification-test/lifecycle/suite.go
shimritproj: The changes have been uploaded @edcdavid

shimritproj:
> You should also add a bad test case yaml example to test both passing and failing pods in the partner repo /examples directory

I will do it in another PR (in partner repo) @edcdavid

shimritproj:
@ramperher Issue with DCI , Are using local-storage in your DCI environment?

cnf-certification-test/lifecycle/suite.go
```diff
@@ -584,23 +581,35 @@ func TestPodTolerationBypass(env *provider.TestEnvironment) {
...
+
+			var StorageClasses = env.StorageClasses
+			var Pvc = env.PersistentVolumeClaims
+			for i := range Pvc {
+				if Pvc[i].Name == put.Spec.Volumes[pvIndex].PersistentVolumeClaim.ClaimName {
+					for j := range StorageClasses {
+						if StorageClasses[j].Provisioner == "kubernetes.io/no-provisioner" && !localStorageFound {
+							localStorageFound = true
+							tnf.ClaimFilePrintf("%s has been found to use a local storage enabled storageClass", put.String())
+							podsWithLocalStorage = append(podsWithLocalStorage, put.String())
```
edcdavid, line 603: instead of checking only the first local volume found, find all of them and report that as a result of this test. So in the end we will have a list of bad volumes instead of bad pods. The string could include the namespace/podname/pvc_name/storageclass_name/provisionner_name per non compliant volume

ramperher:
> @ramperher Issue with DCI , Are using local-storage in your DCI environment?

yes, it is our case, sorry, so this means your test works fine. I've added a condition to avoid failing if this test fails. I'll recheck.

cnf-certification-test/lifecycle/suite.go
shimritproj: I updated but tell me if is correct or need more changes. @edcdavid

cnf-certification-test/lifecycle/suite.go
```diff
@@ -584,23 +581,36 @@ func TestPodTolerationBypass(env *provider.TestEnvironment) {
...
+				if Pvc[i].Name == put.Spec.Volumes[pvIndex].PersistentVolumeClaim.ClaimName {
+					for j := range StorageClasses {
+						if StorageClasses[j].Provisioner == "kubernetes.io/no-provisioner" {
+							tnf.ClaimFilePrintf("%s has been found to use a local storage enabled storageClass.\n Pvc_name: %s, Storageclass_name: %s, Provisionner_name: %s", put.String(), put.Spec.Volumes[pvIndex].PersistentVolumeClaim.ClaimName,
+								StorageClasses[j].Name, StorageClasses[j].Provisioner)
+							podsWithLocalStorage = append(podsWithLocalStorage, put.String())
+							// Break out of the storageList loop.
+							continue
+						}
+						tnf.ClaimFilePrintf("%s has been not found to use a local storage enabled storageClass.\n Pvc_name: %s, Storageclass_name: %s, Provisionner_name: %s", put.String(), put.Spec.Volumes[pvIndex].PersistentVolumeClaim.ClaimName,
```
sebrandon1, line 606: I don't believe we need to print when we found a valid pod, only invalid should be in the claim file print.

pkg/autodiscover/autodiscover.go
```diff
@@ -120,6 +122,10 @@ func DoAutoDiscover() DiscoveredTestData {
 		logrus.Fatalf("Cannot load configuration, error: %v", err)
 	}
 	oc := clientsholder.GetClientsHolder()
+	data.StorageClasses, err = getAllStorageClasses()
+	if err != nil {
+		logrus.Fatalf("err: %v", err)
```
sebrandon1, line 127:
```suggestion
		logrus.Fatalf("Failed to retrieve storageClasses - err: %v", err)
```

sebrandon1 (commented):
Some more comments

cnf-certification-test/lifecycle/suite.go
edcdavid: Looks good now, I would also print the name of the PVC and storage class in this log

edcdavid: I see it now, looks good

edcdavid (approved):

cnf-certification-test/lifecycle/suite.go
shimritproj: Yes but look the comment of @edcdavid 
I understood that need print information also in this case.

cnf-certification-test/lifecycle/suite.go
```diff
@@ -584,23 +581,36 @@ func TestPodTolerationBypass(env *provider.TestEnvironment) {
...
-				continue
+			// We have the list of pods/volumes/claims.
+			// Look through the storageClass list for a match.
+
+			var StorageClasses = env.StorageClasses
+			var Pvc = env.PersistentVolumeClaims
+			for i := range Pvc {
+				if Pvc[i].Name == put.Spec.Volumes[pvIndex].PersistentVolumeClaim.ClaimName {
+					for j := range StorageClasses {
+						if StorageClasses[j].Provisioner == "kubernetes.io/no-provisioner" {
```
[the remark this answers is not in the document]
shimritproj: I did debug and I think that we can write like this-
```
for i := range Pvc {
				if Pvc[i].Name == put.Spec.Volumes[pvIndex].PersistentVolumeClaim.ClaimName {
					for j := range StorageClasses {
						if StorageClasses[j].Provisioner == "kubernetes.io/no-provisioner" && StorageClasses[j].Name == *Pvc[i].Spec.StorageClassName {
							tnf.ClaimFilePrintf("%s has been found to use a local storage enabled storageClass.\n Pvc_name: %s, Storageclass_name: %s, Provisionner_name: %s", put.String(), put.Spec.Volumes[pvIndex].PersistentVolumeClaim.ClaimName,
								StorageClasses[j].Name, StorageClasses[j].Provisioner)
							podsWithLocalStorage = append(podsWithLocalStorage, put.String())
							// Break out of the storageList loop.
							continue
						}
						tnf.ClaimFilePrintf("%s has been not found to use a local storage enabled storageClass.\n Pvc_name: %s, Storageclass_name: %s, Provisionner_name: %s", put.String(), put.Spec.Volumes[pvIndex].PersistentVolumeClaim.ClaimName,
							StorageClasses[j].Name, StorageClasses[j].Provisioner)
					}
				}
			}
```

tell me what do you think?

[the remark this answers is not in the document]
sebrandon1: I think this works but I think you also have to compare the `namespace` between the PVC and the Pod as well like in @greyerof's example.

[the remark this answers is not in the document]
greyerof: Although I haven't found explicit mention yet, It seems like Storage Classes are cluster-wide resources, so there's no need to check namespaces here.

But, you should make sure the pointer to the (optional) field storageClassName string in the pvc exists before de-referencing it for comparing:
```
Pvc[i].Spec.StorageClassName != nil && StorageClasses[j].Name == *Pvc[i].Spec.StorageClassName
```
Also, that "continue" on line 604 should be a `continue PV_LOOP`, instead. As you have already found the associated storage class for this pvc-pv, there's no need to keep iterating through all the storage classes in the cluster for the same that pvc.
```
	PV_LOOP:
		for pvIndex := range put.Spec.Volumes {
		    ...
		    ...
		             if ... Pvc[i].Spec.StorageClassName != nil && StorageClasses[j].Name == *Pvc[i].Spec.StorageClassName {
		                 tnf.ClaimFilePrintf(...)
		                 podsWithLocalStorage = append(podsWithLocalStorage, ...)
		                 continue PV_LOOP
		             }
```
There are other refactors that could be made to improve readability/maintenance of this function:
- Check whether the pvc has a non-nil StorageClassName. If nil, there's no need to iterate trough all the storage classes of the cluster.
- Move both `var StorageClasses = env.StorageClasses` and `var Pvc = env.PersistentVolumeClaims` outside all the loops.
- Make use of a couple helper function just to get the PVC and the StorageClass by name. Code would look quite simpler:
```
func testStorageRequiredPods(env *provider.TestEnvironment) {
	var podsWithLocalStorage []string
		for pvIndex := range put.Spec.Volumes {
			volume := put.Spec.Volumes[pvIndex]
			if volume.PersistentVolumeClaim == nil {
				continue
			}

			pvc := getPvcByName(volume.PersistentVolumeClaim.ClaimName, env.PersistentVolumeClaims)
			if pvc.StorageClassName == nil {
				// This pvc does not use any storage class.
				continue
			}

			sc := getStorageClassByName(pvcStorageClassName, env.StorageClasses)
			if sc.Provisioner == "kubernetes.io/no-provisioner" {
				tnf.ClaimFilePrintf(...)
				podsWithLocalStorage = append(podsWithLocalStorage, put.String())
				continue
			}
		}
	}

	testhelper.AddTestResultLog("Non-compliant", podsWithLocalStorage, tnf.ClaimFilePrintf, ginkgo.Fail)
}
```

cnf-certification-test/lifecycle/suite.go
```diff
@@ -584,23 +581,36 @@ func TestPodTolerationBypass(env *provider.TestEnvironment) {
...
-				tnf.ClaimFilePrintf(err.Error())
-				continue
-			}
-			if statefulset.Spec.ServiceName == localStorage {
-				nonCompliantPods = append(nonCompliantPods, pod)
-				continue
+			// We have the list of pods/volumes/claims.
+			// Look through the storageClass list for a match.
+			for i := range Pvc {
+				if Pvc[i].Name == put.Spec.Volumes[pvIndex].PersistentVolumeClaim.ClaimName {
```
greyerof, line 597: Please check also the Pvc[i].Namespace.

greyerof (approved):

cnf-certification-test/identifiers/identifiers.go
```diff
@@ -255,6 +255,18 @@ The label value is not important, only its presence.`,
...
 		TagExtended)
 
+	TestStorageRequiredPods = AddCatalogEntry(
+		"no-pvs-on-localstorage",
+		common.LifecycleTestKey,
+		`Checks that pods do not place persistent volumes on local storage.`,
+		StorageRequiredPods,
+		InformativeResult,
+		NoDocumentedProcess,
+		VersionOne,
```
sebrandon1, line 265: Just a heads up that #843 will cause a merge conflict here.

cnf-certification-test/lifecycle/suite.go
```diff
@@ -584,23 +581,36 @@ func TestPodTolerationBypass(env *provider.TestEnvironment) {
...
+			for i := range Pvc {
+				if Pvc[i].Name == put.Spec.Volumes[pvIndex].PersistentVolumeClaim.ClaimName && Pvc[i].Namespace == put.Namespace {
+					for j := range StorageClasses {
+						if Pvc[i].Spec.StorageClassName != nil && StorageClasses[j].Name == *Pvc[i].Spec.StorageClassName {
+							tnf.ClaimFilePrintf("%s has been found to use a local storage enabled storageClass.\n Pvc_name: %s, Storageclass_name : %s, Provisionner_name: %s", put.String(), put.Spec.Volumes[pvIndex].PersistentVolumeClaim.ClaimName,
+								StorageClasses[j].Name, StorageClasses[j].Provisioner)
+							podsWithLocalStorage = append(podsWithLocalStorage, put.String())
+							break
+						}
+						tnf.ClaimFilePrintf("%s has been not found to use a local storage enabled storageClass.\n Pvc_name: %s, Storageclass_name : %s, Provisionner_name: %s", put.String(), put.Spec.Volumes[pvIndex].PersistentVolumeClaim.ClaimName,
```
sebrandon1, line 605:
```suggestion
						tnf.ClaimFilePrintf("%s has been not found to use a local storage enabled storageClass.\n Pvc_name: %s, Storageclass_name : %s, Provisioner_name: %s", put.String(), put.Spec.Volumes[pvIndex].PersistentVolumeClaim.ClaimName,
```

pkg/provider/provider.go
```diff
@@ -108,6 +109,7 @@ type TestEnvironment struct { // rename this with testTarget
 	IstioServiceMesh       bool
 	ValidProtocolNames     []string
 	DaemonsetFailedToSpawn bool
+	StorageList            []storagev1.StorageClass
```
sebrandon1, line 112:
```suggestion
	StorageClassList            []storagev1.StorageClass
```

sebrandon1 (commented):
Just a few comments, but looks good.

sebrandon1 (approved):

PR #798, opened 2023-01-19
Presentation on CNF and Service mesh

shimritproj:
Presentation on CNF and Service mesh


sebrandon1 (changes requested):
Could we change the title of the PDF to something more descriptive?

Maybe something like `ServiceMesh_CNF_Presentation.pdf`?

shimritproj:
We move the document to folder in google drive.

PR #803, opened 2023-01-19
[FAR-EDGE] Check for the workload running in apps shared CPU pool with cpu scheduling priority 0

bnshr:


cnf-certification-test/faredge/suite.go
```diff
@@ -0,0 +1,127 @@
+// Copyright (C) 2020-2022 Red Hat, Inc.
```
sebrandon1, line 1:
```suggestion
// Copyright (C) 2020-2023 Red Hat, Inc.
```

cnf-certification-test/faredge/suite.go
```diff
@@ -0,0 +1,127 @@
...
+
+		if !existsSchedulingPolicyAndPriority(stdout, ScheduleOtherPolicy, ScheduleOtherPriority) {
+			ginkgo.Fail("failed")
+		}
+	}
+
+	logrus.Println("Testing ended")
+}
+
+func existsSchedulingPolicyAndPriority(output, schedPolicy string, schedPriority int) bool {
```
sebrandon1, line 100: For organization, we could probably use a `scheduling` folder/file structure to hold these "extra" funcs in (and write unit tests for).  Same as we do in other suites.

sebrandon1 (changes requested):
Just some initial comments.

cnf-certification-test/faredge/suite.go
bnshr: Done!

cnf-certification-test/faredge/suite.go
```diff
@@ -0,0 +1,95 @@
...
+	var env provider.TestEnvironment
+	ginkgo.BeforeEach(func() {
+		env = provider.GetTestEnvironment()
+	})
+	ginkgo.ReportAfterEach(results.RecordResult)
+
+	testID, tags := identifiers.GetGinkgoTestIDAndLabels(identifiers.TestSchedOtherInSharedCpuPool)
+	ginkgo.It(testID, ginkgo.Label(tags...), func() {
+		nonGuranteedPods := env.GetNonGuaranteedPods()
+		testhelper.SkipIfEmptyAll(ginkgo.Skip, nonGuranteedPods)
```
edcdavid, line 46: typo "nonGuranteedPods" -> "nonGuaranteedPods"

cnf-certification-test/faredge/suite.go
```diff
@@ -0,0 +1,95 @@
...
+		testSchedOtherPolicyInSharedCPUPool(&env, filterNonGuaranteedContainers(nonGuranteedPods))
+	})
+})
+
+func filterNonGuaranteedContainers(pods []*provider.Pod) []*provider.Container {
```
edcdavid, line 51: move to provider and use in testhelper.SkipIfEmptyAll

cnf-certification-test/identifiers/identifiers.go
```diff
@@ -1518,4 +1523,17 @@ that there are no changes to the following directories:
 		Tags:                  TestContainerPortNameFormat.Tags,
 		Qe:                    false,
 	},
+	// TODO: Later update
+	TestSchedOtherInSharedCpuPool: {
```
edcdavid, line 1527: use AddCatalogEntry instead to add the new entry

pkg/scheduling/scheduling.go
```diff
@@ -0,0 +1,38 @@
...
+
+func GetSchedulingPolicyAndPriority(chrtCommandOutput string) (schedPolicy string, schedPriority int, err error) {
+	/*	Sample output:
+		pid 476's current scheduling policy: SCHED_OTHER
+		pid 476's current scheduling priority: 0*/
+	const (
+		policySubstr   = "current scheduling policy"
+		prioritySubstr = "current scheduling priority"
+	)
+	for _, line := range strings.Split(chrtCommandOutput, "\n") {
```
edcdavid, line 23: optional: why not split with policySubstr then prioritySubstr or better yet, use a multiline regex similar to the one in parsePingResult

cnf-certification-test/faredge/suite.go
```diff
@@ -0,0 +1,95 @@
...
+		ocpContext := clientsholder.NewContext(debugPod.Namespace, debugPod.Name, debugPod.Spec.Containers[0].Name)
+		pid, err := crclient.GetPidFromContainer(container, ocpContext)
+		if err != nil {
+			logrus.Errorf("unable to get container process id due to: %v", err)
+			continue
+		}
+		logrus.Infof("The obtained process id is %d", pid)
+
+		command := fmt.Sprintf("chrt -p %d", pid)
+		stdout, stderr, err := crclient.ExecCommandContainerNSEnter(command, container)
```
edcdavid, line 81: I do not believe this is always correct. pid is just one of the pid in the container pid namespace, we need to test all the pids in the pid namespace:
to get the pid namespace (container pid is 418087 in this example), 4026533163 is the pid namespace with 2 processes (4026533163 pid       **2**):
```
lsns -p 418087 -t pid  -n
4026533163 pid       2 418087 100999 /usr/share/elasticsearch/jdk/bin/java -Des.networkaddress.cache.ttl=60 -Des.networkaddress.cache.negative.ttl=10 -XX:+AlwaysPreTouch -Xss1m -Djava.awt.headless=true -Dfile.encoding=UTF-8 -Djna.nosys=tr

```
to list all the pids in the pid namespace containing this pid, run: 
```
ps -e  -o pidns,pid,args   
```
then search for the "pid namespace". 4026533163 is the namespace, with 2 pids :418087 and 418194. Each can have a different priority and scheduling:
```
4026533163  418087 /usr/share/elasticsearch/jdk/bin/java -Des.networkaddress.cache.ttl=60 -Des.networkaddress.cache.negative.ttl=10 -XX:+AlwaysPreTouch -Xss1m -Djava.awt.headless=true -Dfile.encoding=UTF-8 -Djna.nosys=true -XX:-OmitStackTraceInFastThrow -Dio.netty.noUnsafe=true -Dio.netty.noKeySetOptimization=true -Dio.netty.recycler.maxCapacityPerThread=0 -Dio.netty.allocator.numDirectArenas=0 -Dlog4j.shutdownHookEnabled=false -Dlog4j2.disable.jmx=true -Djava.locale.providers=COMPAT -Xms1g -Xmx1g -XX:+UseConcMarkSweepGC -XX:CMSInitiatingOccupancyFraction=75 -XX:+UseCMSInitiatingOccupancyOnly -Djava.io.tmpdir=/tmp/elasticsearch-13358563585845435259 -XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath=data -XX:ErrorFile=logs/hs_err_pid%p.log -Xlog:gc*,gc+age=trace,safepoint:file=logs/gc.log:utctime,pid,tags:filecount=32,filesize=64m -Des.cgroups.hierarchy.override=/ -XX:MaxDirectMemorySize=536870912 -Des.path.home=/usr/share/elasticsearch -Des.path.conf=/usr/share/elasticsearch/config -Des.distribution.flavor=default -Des.distribution.type=docker -Des.bundled_jdk=true -cp /usr/share/elasticsearch/lib/* org.elasticsearch.bootstrap.Elasticsearch -Ediscovery.type=single-node
4026533163  418194 /usr/share/elasticsearch/modules/x-pack-ml/platform/linux-x86_64/bin/controller
         -  418514 [kworker/4:1-events]
```
so for each of the pids in the pid namespace we should run chrt -p 418087 and chrt -p 418194:
```
deliedit@redhatwork ptp-operator]$ chrt -p 418194
pid 418194's current scheduling policy: SCHED_OTHER
pid 418194's current scheduling priority: 0
[deliedit@redhatwork ptp-operator]$ chrt -p 418087
pid 418087's current scheduling policy: SCHED_FIFO
pid 418087's current scheduling priority: 1

```

cnf-certification-test/faredge/suite.go
```diff
@@ -0,0 +1,95 @@
+// Copyright (C) 2020-2023 Red Hat, Inc.
```
edcdavid, line 1: I would prefer a more generic name for the suite such as "performance" or similar. Faredge are not the only CNFs that could use high performance/low latency configurations

cnf-certification-test/identifiers/identifiers.go
```diff
@@ -735,6 +735,11 @@ var (
 		Url:     formTestURL(common.ManageabilityTestKey, "container-port-name-format"),
 		Version: VersionOne,
 	}
+	TestSchedOtherInSharedCpuPool = claim.Identifier{
+		Tags:    formTestTags(TagExtended),
```
jmontesi, line 739: `TagFarEdge`

cnf-certification-test/identifiers/remediation.go
```diff
@@ -194,5 +194,7 @@ const (
 
 	StorageRequiredPods = "If the kind of pods is StatefulSet, so we need to make sure that servicename is not local-storage."
 
-	ExclusiveCPUPoolRemediation = `Ensure that if one container in a Pod selects an exclusive CPU pool the rest also select this type of CPU pool`
+	SchedOtherInSharedCpuPoolRemediation = `Ensure that the workload application shared CPU pool has SCHED_OTHER CPU schedule policy`
+
+  ExclusiveCPUPoolRemediation = `Ensure that if one container in a Pod selects an exclusive CPU pool the rest also select this type of CPU pool`
```
jmontesi, line 199: Misaligned.

cnf-certification-test/faredge/suite.go
```diff
@@ -0,0 +1,95 @@
...
+
+		command := fmt.Sprintf("chrt -p %d", pid)
+		stdout, stderr, err := crclient.ExecCommandContainerNSEnter(command, container)
+		if err != nil || stderr != "" {
+			logrus.Errorf("unable to run nsenter due to : %v", err)
+		}
+
+		schedulePolicy, schedulePriority, err := scheduling.GetSchedulingPolicyAndPriority(stdout)
+		if err != nil {
+			logrus.Errorf("unable to run nsenter due to : %v", err)
```
jmontesi, line 88: Update error message.

pkg/scheduling/scheduling.go
```diff
@@ -0,0 +1,38 @@
...
+
+import (
+	"strconv"
+	"strings"
+
+	"github.com/sirupsen/logrus"
+)
+
+const (
+	ScheduleOtherPolicy   = "SCHED_OTHER"
```
jmontesi, line 11: As I commented above, I think `SCHED_IDLE` and `SCHED_BATCH `should be taken into account as well. All of them must have the same priority (0).

cnf-certification-test/faredge/suite.go
```diff
@@ -0,0 +1,95 @@
...
+		}
+
+		if schedulePolicy != scheduling.ScheduleOtherPolicy && schedulePriority != scheduling.ScheduleOtherPriority {
```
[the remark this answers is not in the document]
bnshr: Yeah, now I think checking on priority makes sense for being the priority of all 3 being 0.

cnf-certification-test/faredge/suite.go
bnshr: Done!

cnf-certification-test/faredge/suite.go
bnshr: Thanks! I interpreted the requirement differently as checked only for container process id. Any workload run from that container needs to be changed. I would add the suggested part shortly. :)

pkg/scheduling/scheduling.go
bnshr: Priority based checks are considered.

jmontesi:
It looks like you've forgotten to add the performance suite to the PR.

cnf-certification-test/performance/suite.go
```diff
@@ -0,0 +1,106 @@
...
+		logrus.Infof("obtained pidNamespace for %s is %s", testContainer, pidNamespace)
+
+		// Get the list of process ids running in the pid namespace
+		pids := crclient.GetPidsFromPidNamespace(pidNamespace, testContainer)
+
+		// Check for the specified priority for each processes running in that pid namespace
+		for _, pid := range pids {
+			logrus.Infof("checking the scheduling policy for pid=%d", pid)
+			if !doesProcessContainsSharedCPUPolicy(testContainer, pid) {
+				ginkgo.Fail("The cpu scheduling policy did not match.")
```
sebrandon1, line 86: This could probably use the `testhelper.AddTestResultLog` helper function.

And the error could use more context as to which pod and container failed.

sebrandon1 (commented):
One comment about the failure message.

cnf-certification-test/performance/suite.go
```diff
@@ -0,0 +1,106 @@
...
+			logrus.Errorf("unable to get pid namespace due to: %v", err)
+		}
+		logrus.Infof("obtained pidNamespace for %s is %s", testContainer, pidNamespace)
+
+		// Get the list of process ids running in the pid namespace
+		pids := crclient.GetPidsFromPidNamespace(pidNamespace, testContainer)
+
+		// Check for the specified priority for each processes running in that pid namespace
+		for _, pid := range pids {
+			logrus.Infof("checking the scheduling policy for pid=%d", pid)
```
edcdavid, line 84: use tnf.ClaimFilePrintf to print "curated" logs both on console and in claim file. Move other logs to debug level. Log here should be including info about namespace/pod/container/pid namespace/pid/policy/priority

cnf-certification-test/performance/suite.go
edcdavid: instead of stopping at the first failure, scan all the pods/containers and record all bad ones in a badContainers list (see similar tests using AddTestResultLog )

cnf-certification-test/performance/suite.go
```diff
@@ -0,0 +1,106 @@
...
+	var env provider.TestEnvironment
+	ginkgo.BeforeEach(func() {
+		env = provider.GetTestEnvironment()
+	})
+	ginkgo.ReportAfterEach(results.RecordResult)
+
+	testID, tags := identifiers.GetGinkgoTestIDAndLabels(identifiers.TestSchedOtherInSharedCpuPool)
+	ginkgo.It(testID, ginkgo.Label(tags...), func() {
+		var nonGuaranteedPodContainers []*provider.Container
+		for _, pod := range env.GetNonGuaranteedPods() {
```
edcdavid, line 46: add this functionality in filters.go, e.g. env.GetNonGuaranteedPodContainers

edcdavid (approved):

cnf-certification-test/performance/suite.go
bnshr: It's not clear to me what to print with Debug level and what with ClaimFilePrintf. Could we discuss that?

pkg/scheduling/scheduling.go
bnshr: regex can be skipped here as the text has only 2 lines, so we get no performance benefit of using regex.

cnf-certification-test/performance/suite.go
```diff
@@ -0,0 +1,133 @@
...
+		}
+
+		// Get the list of process ids running in the pid namespace
+		pids := crclient.GetPidsFromPidNamespace(pidNamespace, testContainer)
+
+		// Check for the specified priority for each processes running in that pid namespace
+		isCompliant := processPidsCPUScheduling(pids, testContainer, nonCompliantContainers, SharedCPUScheduling)
+		if isCompliant {
+			compliantContainers = append(compliantContainers, testContainer)
+		}
```
sebrandon1, line 77:
```suggestion
		if processPidsCPUScheduling(pids, testContainer, nonCompliantContainers, SharedCPUScheduling) {
		     compliantContainers = append(compliantContainers, testContainer)
		}
```

cnf-certification-test/performance/suite.go
```diff
@@ -0,0 +1,133 @@
...
+		logrus.Debugf("Processed %v", testContainer)
+	}
+	// Add success results to claim
+	if len(nonCompliantContainers) != 0 {
+		testhelper.AddTestResultLog("Non-compliant", nonCompliantContainers, tnf.ClaimFilePrintf, ginkgo.Fail)
+	}
+	tnf.ClaimFilePrintf("Compliant containers", compliantContainers)
+}
+
+func processPidsCPUScheduling(pids []int, testContainer *provider.Container, nonCompliantContainers map[*provider.Container][]int, check string) (hasCPUSchedulingConditionSuccess bool) {
```
sebrandon1, line 87: Could `processPidsCPUScheduling` and `getProcessCPUScheduling` possibly be moved to another file within the new `performance` suite folder structure?  Maybe `performance/scheduling.go`?  I see there's already a `pkg/scheduling.go` so maybe that would be a good place?  Something like that.  In the other suites we have separated out the helper/logic funcs from the `suite.go` code just for organization.

Also would it be possible to build some unit tests around these funcs?

sebrandon1 (approved):
Just some minor comments but looks good.  Unit tests can be added in a subsequent PR if needed.

edcdavid (approved):

cnf-certification-test/performance/suite.go
bnshr: Adding some unit tests.

ramperher:
@bnshr , can you rebase your change to be able to test your PR in DCI?

pkg/scheduling/scheduling_test.go
```diff
@@ -0,0 +1,137 @@
...
+		{
+			mockGetProcessCPUScheduling: func(pid int, container *provider.Container) (string, int, error) {
+				return "SCHED_FIFO", 11, nil
+			},
+			check:                                 ExclusiveCPUScheduling,
+			expectedCPUSchedulingConditionSuccess: false,
+		},
+	}
+	for _, tc := range testCases {
+		GetProcessCPUSchedulingFn = tc.mockGetProcessCPUScheduling
```
sebrandon1, line 51: You should save the original func in a variable then after the test are complete, set it back to the original.  For example: https://github.com/test-network-function/cnf-certification-test/blob/e4a631bb3e91945af53f31e7195622a9ee3eb15b/cnf-certification-test/lifecycle/scaling/deployment_scaling_test.go#L61

sebrandon1 (approved):
Thanks for the tests!  Just one nit.

bnshr:
@ramperher The DCI job still fails. Can we do something for it?

pkg/scheduling/scheduling.go
```diff
@@ -0,0 +1,107 @@
...
+		if err != nil {
+			logrus.Errorf("error getting the scheduling policy and priority : %v", err)
+			return hasCPUSchedulingConditionSuccess
+		}
+
+		switch check {
+		case SharedCPUScheduling:
+			hasCPUSchedulingConditionSuccess = schedulePriority == 0
+		case ExclusiveCPUScheduling:
+			hasCPUSchedulingConditionSuccess = schedulePriority < 10
```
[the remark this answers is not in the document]
bnshr: That refers to some other JIRA issue.

internal/crclient/crclient.go
```diff
@@ -53,6 +54,57 @@ func GetPidFromContainer(cut *provider.Container, ctx clientsholder.Context) (in
...
+	var pids []int
+	for _, line := range strings.Split(strings.TrimSuffix(stdout, newLine), newLine) {
+		if line == "" {
+			continue
+		}
+		tokens := strings.Fields(line)
+		if tokens[0] == pidNamespace {
+			pid, err := strconv.Atoi(tokens[1])
+			if err != nil {
+				logrus.Errorf("err converting %s by strconv %v", tokens[1], err)
```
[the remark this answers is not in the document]
bnshr: Why? token length is not important here, when err is is not present.

internal/crclient/crclient.go
```diff
@@ -53,6 +54,57 @@ func GetPidFromContainer(cut *provider.Container, ctx clientsholder.Context) (in
 	return strconv.Atoi(strings.TrimSuffix(outStr, "\n"))
 }
 
+// To get the pid namespace of the container
+func GetContainerPidNamespace(testContainer *provider.Container, env *provider.TestEnvironment) (string, error) {
+	// Get the container pid
+	nodeName := testContainer.NodeName
+	debugPod := env.DebugPods[nodeName]
+	if debugPod == nil {
+		ginkgo.Fail(fmt.Sprintf("Debug pod not found on Node: %s", nodeName))
```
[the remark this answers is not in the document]
bnshr: No, we decided for continuing with other containers to provide elaborative details on the test result.

cnf-certification-test/performance/suite.go
```diff
@@ -0,0 +1,76 @@
...
+	nonGuaranteedPodContainers []*provider.Container) {
+	nonCompliantContainers := make(map[*provider.Container][]int)
+	var compliantContainers []*provider.Container
+	for _, testContainer := range nonGuaranteedPodContainers {
+		logrus.Infof("Processing %v", testContainer)
+
+		// Get the pid namespace
+		pidNamespace, err := crclient.GetContainerPidNamespace(testContainer, env)
+		if err != nil {
+			logrus.Errorf("unable to get pid namespace due to: %v", err)
```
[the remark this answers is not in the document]
bnshr: Do we put the error in the middle of some processing in the claim file? We are not reporting each details of the processing there, rather logging with the DEBUG level.

ramperher:
> @ramperher The DCI job still fails. Can we do something for it?

the issue with the volume should be fixed starting tomorrow when updating the DCI packages in our clusters. Now, you can try a rerun when having this issue, I'll do it now

pkg/scheduling/scheduling.go
```diff
@@ -0,0 +1,107 @@
...
+		}
+
+		if !hasCPUSchedulingConditionSuccess {
+			tnf.ClaimFilePrintf("pid=%d in %v with cpu scheduling policy=%s, priority=%s did not satisfy cpu scheduling requirements", pid, testContainer, schedulePolicy, schedulePriority)
+			nonCompliantProcessIds, ok := nonCompliantContainers[testContainer]
+			if !ok {
+				nonCompliantContainers[testContainer] = []int{}
```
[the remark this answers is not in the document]
bnshr: Thanks, done!

internal/crclient/crclient.go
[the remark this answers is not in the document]
greyerof: Maybe I didn't express it correctly: if there's only one token and you try to access `tokens[1]` in whatever part of the code, the program will panic with an out-of-range index run-time error. Example [here](https://go.dev/play/p/nClCJJp_MQh). 

Tha would be a weird scenario, I know. I'm more concerned by the fact that you're still adding the pid to the pids list when err != nil.

cnf-certification-test/performance/suite.go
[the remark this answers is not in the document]
greyerof: We can put whatever we want in the claim file that can help partners to fix a failing test case. I proposed to do a `continue` because, if you can't get the pid namespace, the other errors that you will get If you keep processing this container won't be very useful. In other workds: what's the point of calling `crclient.GetPidsFromPidNamespace` if you already know that pidNamespace is not valid?

internal/crclient/crclient.go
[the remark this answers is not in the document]
greyerof: Then I'm confused. If you want to continue with other containers, you should avoid doing `ginkgo.Fail` here. [ginkfo.Fail](https://onsi.github.io/ginkgo/#mental-model-how-ginkgo-handles-failure) stops the execution of the test case immediately when invoked:

> But there's more. The Fail function panics when it is called. This allows Ginkgo to stop the current closure in its tracks - no subsequent assertions or code in the closure will run. Ginkgo is quite opinionated about this behavior - if an assertion has failed then the current spec is not in an expected state and subsequent assertions will likely fail. This fast-fail approach is especially useful when running slow complex integration tests. It cannot be disabled.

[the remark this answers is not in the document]
bnshr: The existence of debug pod is the pre-requisite in many test cases. In this test case too. Hence, fail the test case. We can have a new JIRA ticket to change all test cases, to handle the non-existence of the debug pod. Wdyt?

internal/crclient/crclient.go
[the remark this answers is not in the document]
bnshr: The output is coming from the linux command. It cannot be that the first output is pidNamespace and then comes invalid output, while the error check is already present at first.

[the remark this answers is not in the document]
bnshr: I would add a unit test here for GetPidsFromPidNamespace function.

PR #813, opened 2023-01-24
Add more unit test coverage

sebrandon1:
Changed the references from `.Value()` to `IsZero()` in the requests/limits checks.

Adds some new unit tests in:
- pkg/provider/events_test.go
- pkg/provider/filters_test.go

Previously, there was no unit test coverage in either of these files.

### DIFF
--- a/cnf-certification-test/accesscontrol/resources/resources.go
+++ b/cnf-certification-test/accesscontrol/resources/resources.go
@@ -12,12 +12,12 @@ func HasRequestsAndLimitsSet(cut *provider.Container) bool {
 		tnf.ClaimFilePrintf("Container has been found missing resource limits: %s", cut.String())
 		passed = false
 	} else {
-		if cut.Resources.Limits.Cpu().Value() == 0 {
+		if cut.Resources.Limits.Cpu().IsZero() {
 			tnf.ClaimFilePrintf("Container has been found missing CPU limits: %s", cut.String())
 			passed = false
 		}
 
-		if cut.Resources.Limits.Memory().Value() == 0 {
+		if cut.Resources.Limits.Memory().IsZero() {
 			tnf.ClaimFilePrintf("Container has been found missing memory limits: %s", cut.String())
 			passed = false
 		}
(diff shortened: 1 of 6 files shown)


rdavid (approved):

cnf-certification-test/accesscontrol/resources/resources.go
```diff
@@ -12,12 +12,12 @@ func HasRequestsAndLimitsSet(cut *provider.Container) bool {
 		tnf.ClaimFilePrintf("Container has been found missing resource limits: %s", cut.String())
 		passed = false
 	} else {
-		if cut.Resources.Limits.Cpu().Value() == 0 {
+		if cut.Resources.Limits.Cpu().AsApproximateFloat64() == 0 {
```
jmontesi, line 15: There is a specific function for comparing quantities to zero in the `resource` package: `IsZero()`. The same comment applies to the rest of the change.

sebrandon1: Good call!  I do remember seeing this.  I will fix.

rdavid: Why is there `float`? Float comparing is complicating.

sebrandon1: Well at the time, Value() was only giving an integer, so I changed it to float.  But now I will be changing it to .IsZero() per Juan's recommendation.

rdavid: Take a look on float's problems:
 https://jvns.ca/blog/2023/01/13/examples-of-floating-point-problems/

pkg/provider/isolation.go
```diff
@@ -36,13 +36,13 @@ func AreResourcesIdentical(p *Pod) bool {
 		memoryLimits := cut.Resources.Limits.Memory()
 
 		// Check for mismatches
-		if cpuRequests.Value() != cpuLimits.Value() {
-			logrus.Debugf("%s has CPU requests %d and limits %d that do not match.", cut.String(), cpuRequests.Value(), cpuLimits.Value())
+		if cpuRequests.AsApproximateFloat64() != cpuLimits.AsApproximateFloat64() {
```
jmontesi, line 39: I'd avoid the use of float here. It looks a bit strange to have to perform this casting just for comparing two quantities. You can use the `Equal()` function of the `resource` package.

sebrandon1: I will change it.  Thanks!

PR #815, opened 2023-01-25
performance: add test to check that no RT app uses exec probes

jmontesi:


sebrandon1 (approved):

bnshr (approved):

cnf-certification-test/performance/suite.go
```diff
@@ -82,3 +90,29 @@ func testSchedulingPolicyInCPUPool(env *provider.TestEnvironment,
...
+			schedPolicy, _, err := scheduling.GetProcessCPUScheduling(pid, cut)
+			if err != nil {
+				badContainers = append(badContainers, cut.String())
+				tnf.ClaimFilePrintf("Could not determine the scheduling policy for container %s (pid=%v), err: %v", cut, pid, err)
+				break
+			}
+			if scheduling.PolicyIsRT(schedPolicy) && cut.HasExecProbes() {
+				badContainers = append(badContainers, cut.String())
+				tnf.ClaimFilePrintf("Container %s defines exec probes while having a RT scheduling policy", cut)
+				break
```
edcdavid, line 113: Instead of breaking after finding an offending PID continue testing all PIDs and print a log indicating the pod/container/pid/process_name. Still return  list of non compliant containers.

jmontesi: Agree on performing the iteration for all PIDs and logging the result. But I do not think it's a good idea to modify the function `GetContainerPids()` to return the process command as it is fixed for all inputs (apart from the specific PID number) and there's already an error being returned.

edcdavid: What do you mean by: "it is fixed for all inputs"? I was just thinking on returning a list of some struct [int, string] but there are many other ways.

jmontesi: Ah, I thought you referred to the command used to determine the PID scheduling policy, but I now I see you meant the process name.
I think it's better to add that in another PR as there might be other test cases that could used that improved logging info as well.

PR #816, opened 2023-01-25
Fix pod filter for (non) guaranteed pods

jmontesi:


pkg/provider/pods.go
```diff
@@ -78,6 +78,10 @@ func ConvertArrayPods(pods []*corev1.Pod) (out []*Pod) {
 	return out
 }
 
+func (p *Pod) IsPodGuaranteed() bool {
+	return AreResourcesIdentical(p)
```
greyerof, line 82: When pods are deployed, kubernetes/openshift will set the field "QOSClass" in the status subresource accordingly. Thus, I believe there's no need to do the checks manually. Wo what about removing the whole "AreResourcesIdentical()" function and just check that field:
```
return p.Status.QOSClass == corev1.PodQOSGuaranteed
```

jmontesi: The downside to that approach is that we can only log that the Pod does not have QoS Guaranteed but no reason about why. If we use the requirements about requests and limits we can specify why that Pod does not meed the criteria for having that QoS (_e. g._ the CPU requests are not equal to the CPU limits, absent MEM limits, etc.).

greyerof (approved):

edcdavid (approved):

PR #827, opened 2023-01-30
Dynamically install ginkgo based on go.mod in 'install-tools'

sebrandon1:
Instead of having to manually bump the `install-tools` path in the Makefile every time Ginkgo updates, this should dynamically grab the line that installs whatever is in the go.mod.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -115,7 +115,7 @@ build-cnf-tests-debug:
 
 # Install build tools and other required software.
 install-tools:
-	go install github.com/onsi/ginkgo/v2/ginkgo@v2.7.0
+	go install "$$(awk '/ginkgo/ {printf "%s/ginkgo@%s", $$1, $$2}' go.mod)"
 
 # Install golangci-lint	
 install-lint:


Makefile
```diff
@@ -115,7 +116,7 @@ build-cnf-tests-debug:
 
 # Install build tools and other required software.
 install-tools:
-	go install github.com/onsi/ginkgo/v2/ginkgo@v2.7.0
+	go install ${GINKGO_VERSION}
```
rdavid, line 119: I thinks, this will work.
```
go install "$(awk '/ginkgo/ {printf "%s/ginkgo@%s\n", $1, $2}' go.mod)"
```

sebrandon1: Tried your suggestion, I get:

```
$ make install-tools
go install ""
no Go files in /Users/bpalm/Repositories/go/src/github.com/test-network-function/cnf-certification-test
make: *** [install-tools] Error 1
```

I tried adding `shell` to the line because its Make and:
```
$ make install-tools
awk: syntax error at source line 1
 context is
        /ginkgo/ {printf "%s/ginkgo@%s", >>>  , <<< 
awk: illegal statement at source line 1
go install 
no Go files in /Users/bpalm/Repositories/go/src/github.com/test-network-function/cnf-certification-test
make: *** [install-tools] Error 1
```

This is why I'm just using another script in the first place 😄

rdavid: @sebrandon1 Makefile needs double `$$`, e.g.:
```
go install "$$(awk '/ginkgo/ {printf "%s/ginkgo@%s", $$1, $$2}' go.mod)"
```

sebrandon1: Updated!  Thanks!

PR #804, opened 2023-01-19
Modify scripts to mount docker config

sebrandon1:
Broken out from #631 

Using the `PFLT_DOCKERCONFIG` variable from preflight (found [here](https://github.com/redhat-openshift-ecosystem/openshift-preflight/blob/main/docs/CONFIG.md)) as a prerequisite for running the preflight library.

This change is in the PR #631 but I wanted to make it a separate change for reviewing sake.

If merged, nothing functionally changes it just has a new env variable available in the container.

build-depends: 27115

PR #831, opened 2023-02-01
Safeguard against empty variable

sebrandon1:
If `CONTAINER_TNF_DOCKERCFG` is unset after the discovery phase, just use a default string of `NA`.

#631 will require an actual docker config being created/supplied but as of right now it isn't merged yet.

Related to: #804

### DIFF
--- a/script/run-container.sh
+++ b/script/run-container.sh
@@ -125,6 +125,11 @@ container_tnf_kubeconfig_volumes_cmd_args=$(printf -- "-v %s " "${container_tnf_
 CONTAINER_TNF_DOCKERCFG=$(join_paths "${container_tnf_dockercfg_paths[@]}")
 container_tnf_dockercfg_volumes_cmd_args=$(printf -- "-v %s " "${container_tnf_dockercfg_volume_bindings[@]}")
 
+# Safeguard against an empty variable
+if [ -z "$CONTAINER_TNF_DOCKERCFG" ]; then
+	CONTAINER_TNF_DOCKERCFG=NA
+fi
+
 if [ -n "${LOCAL_TNF_CONFIG}" ]; then
 	CONFIG_VOLUME_MOUNT_ARG="-v $LOCAL_TNF_CONFIG:$CONTAINER_TNF_DIR/config:Z"
 fi


script/run-container.sh
```diff
@@ -125,6 +125,11 @@ container_tnf_kubeconfig_volumes_cmd_args=$(printf -- "-v %s " "${container_tnf_
 CONTAINER_TNF_DOCKERCFG=$(join_paths "${container_tnf_dockercfg_paths[@]}")
 container_tnf_dockercfg_volumes_cmd_args=$(printf -- "-v %s " "${container_tnf_dockercfg_volume_bindings[@]}")
 
+# Safeguard against an empty variable
+if [[ -n $CONTAINER_TNF_DOCKERCFG ]]; then
```
greyerof, line 129: Shouldn't you use "-z" instead? From[ bash manual](https://www.gnu.org/software/bash/manual/html_node/Bash-Conditional-Expressions.html):
```
-z string
True if the length of string is zero.

-n string
string
True if the length of string is non-zero.
```

script/run-container.sh
```diff
@@ -125,6 +125,11 @@ container_tnf_kubeconfig_volumes_cmd_args=$(printf -- "-v %s " "${container_tnf_
...
+	CONTAINER_TNF_DOCKERCFG="$HOME/.docker/config"
```
greyerof, line 130: This makes `$HOME/.docker/config` a mandatory file to exist in the system where our suite runs. What if that file doesn't exist? Wasn't it optional?

script/run-container.sh
rdavid: I'd use single bracket:
```
# Sets default if it is empty.
[ -z "$CONTAINER_TNF_DOCKERCFG" ] && CONTAINER_TNF_DOCKERCFG="$HOME/.docker/config"
```

script/run-container.sh
sebrandon1: Yeah the flag is optional but the `-e` variable in the container run command is still empty.  I thought about making this value `CONTAINER_TNF_DOCKERCFG="NA"` or something.  It doesn't have to be the $HOME/.docker/config.

script/run-container.sh
sebrandon1: Okay, I changed it to: 

```
if [ -z $CONTAINER_TNF_DOCKERCFG ]; then
	CONTAINER_TNF_DOCKERCFG="NA"
fi
```

script/run-container.sh
sebrandon1: Changed to: 
```
if [ -z $CONTAINER_TNF_DOCKERCFG ]; then
	CONTAINER_TNF_DOCKERCFG="NA"
fi
```

script/run-container.sh
```diff
@@ -125,6 +125,11 @@ container_tnf_kubeconfig_volumes_cmd_args=$(printf -- "-v %s " "${container_tnf_
 CONTAINER_TNF_DOCKERCFG=$(join_paths "${container_tnf_dockercfg_paths[@]}")
 container_tnf_dockercfg_volumes_cmd_args=$(printf -- "-v %s " "${container_tnf_dockercfg_volume_bindings[@]}")
 
+# Safeguard against an empty variable
+if [ -z "$CONTAINER_TNF_DOCKERCFG" ]; then
+	CONTAINER_TNF_DOCKERCFG="NA"
```
rdavid, line 130: There is no need for double quotes, `CONTAINER_TNF_DOCKERCFG=NA` is fine.

sebrandon1: Done.

PR #621, opened 2022-10-27
Use privileged deamonset library

bnshr:

PR #833, opened 2023-02-01
Skip tests that require daemonset

sebrandon1:
I _think_ I have covered all of the test cases that require using the debug daemonset pods to retrieve information.

Changes:
- Sets a variable `env.DaemonsetFailedToSpawn` if there's a problem with the daemonset coming up.
- Skips tests that require the daemonset to complete.
- Removes leftover function `DeployPartnerTestDaemonset` which is no longer needed since we are using the #621 change for the privileged daemonset library.  Delete the internal/daemonset package and files.


pkg/provider/provider.go
```diff
@@ -144,12 +144,18 @@ var (
...
 	// Wait for the debug pods to be ready before the autodiscovery starts.
+	oc := clientsholder.GetClientsHolder()
+	k8sPriviledgedDs.SetDaemonSetClient(oc.K8sClient)
 	err := k8sPriviledgedDs.WaitDaemonsetReady(DaemonSetNamespace, DaemonSetName, debugPodsTimeout)
 	if err != nil {
 		logrus.Errorf("Debug daemonset failure: %s", err)
+
+		// Because of this failure, we are only able to run a certain amount of tests that do not rely
+		// on the existence of the daemonset debug pods.
+		env.DaemonsetFailedToSpawn = true
```
rdavid, line 157: Is `env.DebugPods` be empty in case of the failure? It will skip the tests anyway.

sebrandon1: This also explicitly calls `ginkgo.Skip` so it'll show up as a proper skipped test rather than just a pass result.

rdavid: >It'll be partially populated
It is an option to zero all debug pods and remain with current skip mechanism.

>also explicitly calls
The failure prints error message anyway.

Overall, it is just a thought. I agree that more explicit handling - the better.

sebrandon1: Right but we don't want to fail some partner's workload because our debug pods fail to spawn causing confusion.

pkg/provider/provider.go
```diff
@@ -142,14 +146,41 @@ var (
 	loaded = false
 )
 
-func buildTestEnvironment() { //nolint:funlen
+// Build image with version based on environment variables if provided, else use a default value
```
rdavid, line 149: I'd consider:
```
func getEnv(key, fallback string) string {
    if value, ok := os.LookupEnv(key); ok {
        return value
    }
    return fallback
}
```

More details:
 https://stackoverflow.com/questions/40326540/how-to-assign-default-value-if-env-var-is-empty

sebrandon1: Ah yes, this is just a copy/paste from the old function.  I can fix this in a subsequent PR.

sebrandon1: Just kidding, I fixed it.

sebrandon1:
@ramperher Thanks!  Fixed the additional tests in https://github.com/test-network-function/cnf-certification-test/commit/b3375e127e8b7ab8a600c81b3613b9447cbfceaa

pkg/provider/provider.go
```diff
@@ -142,14 +146,41 @@ var (
...
+func buildImageWithVersion() string {
+	tnfPartnerRepo := os.Getenv("TNF_PARTNER_REPO")
```
greyerof, line 151: Parsing/reading config env vars is done through the [pkg/configuration/utils package](https://github.com/test-network-function/cnf-certification-test/blob/main/pkg/configuration/utils.go) `init()` function, which relies on `github.com/kelseyhightower/envconfig` library to do it. I never liked it, but, to keep consistency with the other ones, you should add those env vars [TestParameters struct](https://github.com/test-network-function/cnf-certification-test/blob/a5c7e9c20fb2d23efb8d3682804beea573ec5f05/pkg/configuration/configuration.go#L125).

sebrandon1: Okay!  Makes sense.  This is just a copy paste from another file in our code.  Can we address with a subsequent PR?

greyerof: Of course 👍

pkg/provider/provider.go
```diff
@@ -142,14 +146,41 @@ var (
...
+
+	return tnfPartnerRepo + "/" + supportImage
+}
+
+func buildTestEnvironment() { //nolint:funlen,gocyclo
 	start := time.Now()
+	env = TestEnvironment{}
 	// Wait for the debug pods to be ready before the autodiscovery starts.
-	err := k8sPriviledgedDs.WaitDaemonsetReady(DaemonSetNamespace, DaemonSetName, debugPodsTimeout)
+	oc := clientsholder.GetClientsHolder()
```
jmontesi, line 167: I'd encapsulate all these statements to deploy the daemon set in a function that returns `bool` in order to fill `DaemonsetFailedToSpawn`. Something like: `env.DaemonsetFailedToSpawn = deployDaemonSet()`.

jmontesi (approved):

pkg/provider/provider.go
sebrandon1: Agreed.  This section was copy/pasted from the older `internal/daemonset/daemonset.go` file.  Can be adjusted in a subsequent PR.

rdavid (approved):

PR #834, opened 2023-02-01
POC: Add Failure Reason to claim.json

edcdavid:
This is a proof of concept to add a more meaningful "Failure Reason" to the claim.json results section. 
Example FailureReason field output format. 
When the test succeeds, the reason is at the end of the test log. This is because there is no function equivalent to gingko.Fail for passing test with a reason.
Also using regex to parse the ps command output to retrieve also the command args (to add more context to this particular reason ).

This is just for illustration, in the successful case
 ```
{
  "CompliantObjectsOut": [
    {
      "ObjectType": "ContainerProcess",
      "ObjectFields": {
        "ContainerName": "task-pv-container",
        "Namespace": "tnf",
        "PodName": "task-pv-pod",
        "ProcessCommandLine": "sleep 30",
        "ReasonForCompliance": "process satisfies: SHARED_CPU_SCHEDULING: scheduling priority == 0",
        "SchedulingPolicy": "SCHED_OTHER",
        "SchedulingPriority": "0"
      }
    },
    {
      "ObjectType": "ContainerProcess",
      "ObjectFields": {
        "ContainerName": "task-pv-container",
        "Namespace": "tnf",
        "PodName": "task-pv-pod",
        "ProcessCommandLine": "/bin/bash -c -- while true; do sleep 30; done;",
        "ReasonForCompliance": "process satisfies: SHARED_CPU_SCHEDULING: scheduling priority == 0",
        "SchedulingPolicy": "SCHED_OTHER",
        "SchedulingPriority": "0"
      }
    }
  ],
  "NonCompliantObjectsOut": null
}

}


```
Internal error happening during the testcase are also recorded in the non-compliant list at the level at which it happened (in this case at container level):
```
{
  "CompliantObjectsOut": null,
  "NonCompliantObjectsOut": [
    {
      "ObjectType": "Container",
      "ObjectFields": {
        "ContainerName": "task-pv-container",
        "Namespace": "tnf",
        "PodName": "task-pv-pod",
        "ReasonForNonCompliance": "Internal error, err=unable to run nsenter due to : cannot execute command: \" ps -e -o pidns,pid,args \"  on container: task-pv-container pod: task-pv-pod ns: tnf err:command terminated with exit code 1"
      }
    }
  ]
}
```
Another example for the SCC test highlighting new flexible fields:
```
{
  "CompliantObjectsOut": [
    {
      "ObjectType": "Container",
      "ObjectFields": {
        "Category": "CategoryID1(limited access granted automatically)",
        "ContainerName": "task-pv-container",
        "Namespace": "tnf",
        "PodName": "task-pv-pod",
        "ReasonForNonCompliance": "container category is category 1 or category NoUID0"
      }
    },
    {
      "ObjectType": "Container",
      "ObjectFields": {
        "Category": "CategoryID1(limited access granted automatically)",
        "ContainerName": "test",
        "Namespace": "tnf",
        "PodName": "test-0",
        "ReasonForNonCompliance": "container category is category 1 or category NoUID0"
      }
    },
    {
      "ObjectType": "Container",
      "ObjectFields": {
        "Category": "CategoryID1(limited access granted automatically)",
        "ContainerName": "test",
        "Namespace": "tnf",
        "PodName": "test-1",
        "ReasonForNonCompliance": "container category is category 1 or category NoUID0"
      }
    },
    {
      "ObjectType": "Container",
      "ObjectFields": {
        "Category": "CategoryID1(limited access granted automatically)",
        "ContainerName": "test",
        "Namespace": "tnf",
        "PodName": "test-5c99f464d8-jd7mt",
        "ReasonForNonCompliance": "container category is category 1 or category NoUID0"
      }
    },
    {
      "ObjectType": "Container",
      "ObjectFields": {
        "Category": "CategoryID1(limited access granted automatically)",
        "ContainerName": "test",
        "Namespace": "tnf",
        "PodName": "test-5c99f464d8-nxfpj",
        "ReasonForNonCompliance": "container category is category 1 or category NoUID0"
      }
    },
    {
      "ObjectType": "Cnf",
      "ObjectFields": {
        "Category": "CategoryID1(limited access granted automatically)",
        "ReasonForNonCompliance": "Overall Pod category"
      }
    }
  ],
  "NonCompliantObjectsOut": null

```

### DIFF
--- a/cnf-certification-test/accesscontrol/suite.go
+++ b/cnf-certification-test/accesscontrol/suite.go
@@ -648,26 +648,25 @@ func test1337UIDs(env *provider.TestEnvironment) {
 // an allowed one will pass the test
 
 func testContainerSCC(env *provider.TestEnvironment) {
-	var badContainer []securitycontextcontainer.PodListcategory
-	var goodContainer []securitycontextcontainer.PodListcategory
+	var compliantObjects []*testhelper.ObjectOut
+	var nonCompliantObjects []*testhelper.ObjectOut
 	highLevelCat := securitycontextcontainer.CategoryID1
 	for _, pod := range env.Pods {
 		listCategory := securitycontextcontainer.CheckPod(pod)
 		for _, cat := range listCategory {
 			if cat.Category > securitycontextcontainer.CategoryID1NoUID0 {
-				badContainer = append(badContainer, cat)
+				aContainerOut := testhelper.NewContainerObjectOutBase(cat.NameSpace, cat.Podname, cat.Containername, "container category is NOT category 1 or category NoUID0", false).AddField(testhelper.Category, cat.Category.String())
+				nonCompliantObjects = append(nonCompliantObjects, aContainerOut)
 			} else {
-				goodContainer = append(goodContainer, cat)
+				aContainerOut := testhelper.NewContainerObjectOutBase(cat.NameSpace, cat.Podname, cat.Containername, "container category is category 1 or category NoUID0", false).AddField(testhelper.Category, cat.Category.String())
+				compliantObjects = append(compliantObjects, aContainerOut)
 			}
 			if cat.Category > highLevelCat {
 				highLevelCat = cat.Category
 			}
 		}
 	}
-	logrus.Infof("CNF category (highest container category across all containers):  %s \n", highLevelCat)
-	tnf.ClaimFilePrintf("CNF category (highest container category across all containers):  %s \n", highLevelCat)
-	logrus.Infof("List of containers that are Category1 or CategoryNoUID0 %+v \n", goodContainer)
-	tnf.ClaimFilePrintf("List of containers that are Category1 or CategoryNoUID0 %+v \n", goodContainer)
-	logrus.Infof("List of non-compliant containers that are not from Category1 or CategoryNoUID0 - %+v", badContainer)
-	testhelper.AddTestResultLog("List of non-compliant containers that are not from Category1 or CategoryNoUID0 - ", badContainer, tnf.ClaimFilePrintf, ginkgo.Fail)
+	aCNFOut := testhelper.New("Overall Pod category", false).SetType(testhelper.CnfType).AddField(testhelper.Category, highLevelCat.String())
+	compliantObjects = append(compliantObjects, aCNFOut)
+	testhelper.AddTestResultReason(compliantObjects, nonCompliantObjects, ginkgo.Fail)
 }
--- a/cnf-certification-test/accesscontrol/suite.go
+++ b/cnf-certification-test/accesscontrol/suite.go
@@ -648,17 +648,19 @@ func test1337UIDs(env *provider.TestEnvironment) {
 // an allowed one will pass the test
 
 func testContainerSCC(env *provider.TestEnvironment) {
-	var compliantObjects []*testhelper.ObjectOut
-	var nonCompliantObjects []*testhelper.ObjectOut
+	var compliantObjects []*testhelper.ReportObject
+	var nonCompliantObjects []*testhelper.ReportObject
 	highLevelCat := securitycontextcontainer.CategoryID1
 	for _, pod := range env.Pods {
 		listCategory := securitycontextcontainer.CheckPod(pod)
 		for _, cat := range listCategory {
 			if cat.Category > securitycontextcontainer.CategoryID1NoUID0 {
-				aContainerOut := testhelper.NewContainerObjectOutBase(cat.NameSpace, cat.Podname, cat.Containername, "container category is NOT category 1 or category NoUID0", false).AddField(testhelper.Category, cat.Category.String())
+				aContainerOut := testhelper.NewContainerReportObject(cat.NameSpace, cat.Podname, cat.Containername, "container category is NOT category 1 or category NoUID0", false).
+					SetType(testhelper.ContainerCategory).
+					AddField(testhelper.Category, cat.Category.String())
 				nonCompliantObjects = append(nonCompliantObjects, aContainerOut)
 			} else {
-				aContainerOut := testhelper.NewContainerObjectOutBase(cat.NameSpace, cat.Podname, cat.Containername, "container category is category 1 or category NoUID0", false).AddField(testhelper.Category, cat.Category.String())
+				aContainerOut := testhelper.NewContainerReportObject(cat.NameSpace, cat.Podname, cat.Containername, "container category is category 1 or category NoUID0", false).AddField(testhelper.Category, cat.Category.String())
 				compliantObjects = append(compliantObjects, aContainerOut)
 			}
 			if cat.Category > highLevelCat {
@@ -666,7 +668,7 @@ func testContainerSCC(env *provider.TestEnvironment) {
 			}
 		}
 	}
-	aCNFOut := testhelper.New("Overall Pod category", false).SetType(testhelper.CnfType).AddField(testhelper.Category, highLevelCat.String())
+	aCNFOut := testhelper.NewReportObject("Overall CNF category", testhelper.CnfType, false).AddField(testhelper.Category, highLevelCat.String())
 	compliantObjects = append(compliantObjects, aCNFOut)
 	testhelper.AddTestResultReason(compliantObjects, nonCompliantObjects, ginkgo.Fail)
 }
--- a/pkg/testhelper/testhelper.go
+++ b/pkg/testhelper/testhelper.go
@@ -51,9 +51,12 @@ const (
 	SchedulingPriority     = "SchedulingPriority"
 	ReasonForNonCompliance = "ReasonForNonCompliance"
 	ReasonForCompliance    = "ReasonForCompliance"
+	Category               = "Category"
 )
 
 const (
+	UndefinedType        = "Undefined Type"
+	CnfType              = "Cnf"
 	PodType              = "Pod"
 	ContainerType        = "Container"
 	ContainerProcessType = "ContainerProcess"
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

**Comment by ramperher** [comment]
Just one comment: don't hesitate to check DCI jobs to see if claim.json file is displaying the new information. For the moment, I haven't seen this new failure reason there, just to let you know.

**Comment by jmontesi** [comment]
Looks good in general. But one thing that I'd be very careful about is propagating this `*testhelper.ObjectOut` as a return type for functions processing those objects (which are the simpler types for Container, Pods, etc.). In my opinion, that particular abstract type for objects should only be used at the top level function of the test cases for constructing the compliant/non-compliant slices. The rest of helper/processing functions should continue using the standard types.

**Comment by greyerof** [comment]
Maybe I'm alone here, but I don't think it's necessary to go deeper than standard kubernetes resources, being "pod" the lowest level. Otherwise, we might end up flooding the claim file with too many of these custom objects, with a potencially long (and weird, in my opinion) types list in the code. The testhelper package will have a lot of "NewXXX()" functions... You'll be lucky if you can re-use an existing one.
For this specific example of the process scheduling:
```
{
    ...
    nonCompliantResources : [
      {
        resource="pod",
        name="test-pod-1",
        namespace="test-namespace-1",
        reason="Container test-container-1: process test-process1 has invalid scheduling policy: SCHED_OTHER\nContainer test-container-2:, test-process-2 has invalid scheduling policy: SCHED_RR".
      },
      {
        resource="pod",
        name="test-pod-2",
        namespace="test-namespace-1",
        reason="Container test-container-3:, process test-process1 has invalid scheduling policy: SCHED_OTHER",
      },
      ...
	],   

    compliantResources : [
      {
        resource="pod",
        name="test-pod-3",
        namespace="test-namespace-1"
      },
      ...
    ]
}
```

Another variant:
```
    testedResources: {
       compliant: [
        {
            resource="..."
            name: "...",
            namespace="..."
        }
       ],
       nonCompliant: [
        {
            resource="...",
            name="...",
            namespace="...",
            reason="..."           
        }
       ]
    }
```

**Comment by edcdavid** [comment]
I would still prefer to use the granularity of the the object we are testing. In my opinion, it is still simple, since we don't have to convert much between what we test and how we report it. Also, thinking about the future, it will make it easier to parse in the html web page where we could for instance have a hierarchical view of failed objects. I am not sure about the comment about re-use. Unless I miss something,  most of lowest level objects will be k8s objects such as namespaces, pods,  containers, volumes, etc: these will be reused a lot. Then for the test cases that are going through non k8s objects such as processes I don't think we will have a lot of them but they would still be re-used. I am not suggesting to add an object for every single thing we are testing, only when it makes sense in the way we present the results, for instance the volume object would include info about PV/PVC/storageclass.

File: cnf-certification-test/accesscontrol/suite.go [hunk 1]
```diff
@@ -644,28 +644,27 @@ func Test1337UIDs(env *provider.TestEnvironment) {
...
-	var goodContainer []securitycontextcontainer.PodListcategory
+	var compliantObjects []*testhelper.ObjectOut
+	var nonCompliantObjects []*testhelper.ObjectOut
 	highLevelCat := securitycontextcontainer.CategoryID1
 	for _, pod := range env.Pods {
 		listCategory := securitycontextcontainer.CheckPod(pod)
 		for _, cat := range listCategory {
 			if cat.Category > securitycontextcontainer.CategoryID1NoUID0 {
-				badContainer = append(badContainer, cat)
+				aContainerOut := testhelper.NewContainerObjectOutBase(cat.NameSpace, cat.Podname, cat.Containername, "container category is NOT category 1 or category NoUID0", false).AddField(testhelper.Category, cat.Category.String())
```
jmontesi, line 654: This is not a container. It looks like it needs a new type. Maybe, instead of creating plenty of very specific objects, we could use a "custom" type for objects different from Pod/Container/Process.

pkg/testhelper/testhelper.go
```diff
@@ -27,6 +31,90 @@ const (
...
+	out.ObjectFields[SchedulingPolicy] = aPolicy
+	out.ObjectFields[SchedulingPriority] = aPriority
+	return out
+}
+
+func NewContainerObjectOut(aContainer *provider.Container, aReason string, isCompliant bool) (out *ObjectOut) {
+	return NewContainerObjectOutBase(aContainer.Namespace, aContainer.Podname, aContainer.Name, aReason, isCompliant)
+}
+
+func NewContainerObjectOutBase(aNamespace, aPodName, aContainerName, aReason string, isCompliant bool) (out *ObjectOut) {
```
jmontesi, line 78: It looks unnecessary to have two versions of a function to create container objects.

pkg/testhelper/testhelper.go
```diff
@@ -27,6 +31,90 @@ const (
...
+func NewPodObjectOut(aPod *provider.Pod, aCategory, aReason string, isCompliant bool) (out *ObjectOut) {
+	out = New(aReason, isCompliant)
+	out.ObjectType = PodType
+	out.ObjectFields[Namespace] = aPod.Namespace
+	out.ObjectFields[PodName] = aPod.Name
+	out.ObjectFields[Category] = aCategory
+	return out
+}
+
+func New(aReason string, isCompliant bool) (out *ObjectOut) {
```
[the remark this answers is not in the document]
greyerof: My suggestions: `NewReportObject` or `NewTestTargetReport`.

Also, the ObjectType is a good candidate to be included as a parameter of this function.

pkg/testhelper/testhelper.go
```diff
@@ -27,6 +31,90 @@ const (
...
+
+const (
+	UndefinedType        = "Undefined Type"
+	CnfType              = "Cnf"
+	PodType              = "Pod"
+	ContainerType        = "Container"
+	ContainerProcessType = "ContainerProcess"
+)
+
+func NewProcessObjectOut(aContainer *ObjectOut, aPolicy, aPriority, aCommandLine string) (out *ObjectOut) {
```
greyerof, line 65: Although I don't like the fact that this function can only be used in test cases that deal with process scheduling policies, it looks like a good candidate to make a it a struct's method:
```
func (object *ObjectOut) SetContainerProcessValues(policy, priority, commandLine string) (*ObjectOut) {
   ...
```
Thus, using my name suggestion on line 96 for the base function, you could do:
```
containerProcessReport := NewTestTargetReport(ContainerProcessType, false).SetContainerProcessValues("SCHED_OTHER", 0, "...")
```
or
```
containerProcessReport := NewTestTargetReport(ContainerProcessType, false)
containerProcessReport.SetContainerProcessValues("SCHED_OTHER", 0, "...")
```

cnf-certification-test/accesscontrol/suite.go
sebrandon1: Maybe a `pod/container/process` object func that returns a shared type of struct?

`func (p *Pod) ObjectOut(...params...) ObjectOut {}`

`func (c *Container) ObjectOut(...params...) ObjectOut {}`

pkg/testhelper/testhelper.go
sebrandon1: Might be a use-case for generics?

edcdavid:
> Just one comment: don't hesitate to check DCI jobs to see if claim.json file is displaying the new information. For the moment, I haven't seen this new failure reason there, just to let you know.

@ramperher thanks, sure will double check, the failure reason should be present only when it fails, unfortunately there is no "success" reason so I am just writing it  as the last piece of text in the log for a given test case

pkg/testhelper/testhelper.go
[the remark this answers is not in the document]
edcdavid: Fixed

pkg/testhelper/testhelper.go
```diff
@@ -27,6 +31,90 @@ const (
...
+	out.ObjectFields[ContainerName] = aContainerName
+	return out
+}
+
+func NewPodObjectOut(aPod *provider.Pod, aCategory, aReason string, isCompliant bool) (out *ObjectOut) {
+	out = New(aReason, isCompliant)
+	out.ObjectType = PodType
+	out.ObjectFields[Namespace] = aPod.Namespace
+	out.ObjectFields[PodName] = aPod.Name
+	out.ObjectFields[Category] = aCategory
```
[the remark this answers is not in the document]
edcdavid: fixed

pkg/testhelper/testhelper.go
```diff
@@ -82,3 +170,17 @@ func AddTestResultLog(prefix string, object interface{}, log func(string, ...int
...
+
+func AddTestResultReason(compliantObject, nonCompliantObject []*ObjectOut, fail func(string, ...int)) {
+	var aReason FailureReasonOut
+	aReason.CompliantObjectsOut = compliantObject
+	aReason.NonCompliantObjectsOut = nonCompliantObject
+	bytes, err := json.Marshal(aReason)
+	if err != nil {
+		logrus.Errorf("Could not Marshall FailureReason object, err=%s", err)
+	}
+	logrus.Info(string(bytes))
```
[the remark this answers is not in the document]
edcdavid: Yes, correct, I did not want to add another error handling when calling the reporting function

pkg/testhelper/testhelper.go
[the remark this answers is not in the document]
edcdavid: I left it out on purpose. In my opinion, the PID is not very helpful in the report as it is dynamic. The information identifying better the process is the command line. A Partner looking at it will know exactly which pod/container is identified.

cnf-certification-test/accesscontrol/suite.go
edcdavid: fixed. Used the SetType function to re-set the type a second time, while inheriting all the fields from the base type.

pkg/testhelper/testhelper.go
edcdavid: fixed

pkg/testhelper/testhelper.go
edcdavid: Fixed.

pkg/testhelper/testhelper.go
```diff
@@ -60,42 +59,35 @@ const (
...
-	out.ObjectType = ContainerProcessType
-	out.ObjectFields[ProcessCommandLine] = aCommandLine
-	out.ObjectFields[SchedulingPolicy] = aPolicy
-	out.ObjectFields[SchedulingPriority] = aPriority
-	return out
-}
-
-func NewContainerObjectOut(aContainer *provider.Container, aReason string, isCompliant bool) (out *ObjectOut) {
-	return NewContainerObjectOutBase(aContainer.Namespace, aContainer.Podname, aContainer.Name, aReason, isCompliant)
+func (obj *ReportObject) SetContainerProcessValues(aPolicy, aPriority, aCommandLine string) *ReportObject {
```
sebrandon1, line 65: Since `obj` is already a pointer, we don't need to return the `*ReportObject`.

pkg/testhelper/testhelper.go
```diff
@@ -105,12 +97,12 @@ func New(aReason string, isCompliant bool) (out *ObjectOut) {
 	return out
 }
 
-func (obj *ObjectOut) AddField(aKey, aString string) (out *ObjectOut) {
+func (obj *ReportObject) AddField(aKey, aString string) (out *ReportObject) {
```
sebrandon1, line 100: Same here with the pointer comment.

pkg/testhelper/testhelper.go
```diff
@@ -105,12 +97,12 @@ func New(aReason string, isCompliant bool) (out *ObjectOut) {
...
 	obj.ObjectFields[aKey] = aString
 	return obj
 }
 
-func (obj *ObjectOut) SetType(aType string) (out *ObjectOut) {
+func (obj *ReportObject) SetType(aType string) (out *ReportObject) {
```
sebrandon1, line 105: Same here with the pointer comment.

sebrandon1 (commented):
Some nits about pointers.

pkg/testhelper/testhelper.go
edcdavid: I agree but the reason why we need the return value in this case is to be able to daisy chain calls to similar functions and eventually assign the result to a variable.

pkg/testhelper/testhelper.go
edcdavid: same resolution as above

pkg/testhelper/testhelper.go
edcdavid: same resolution as above

sebrandon1 (approved):

greyerof (approved):

PR #835, opened 2023-02-02
Far Edge TC: check restart-on-reboot label for sriov pods.

greyerof:
Added pod method to return whether a pod is using any sriov interface or not. Reads CNCF's networks annotation on each pod and check if any of the CNIs of that network is sriov type.

### DIFF
--- a/cnf-certification-test/networking/suite.go
+++ b/cnf-certification-test/networking/suite.go
@@ -423,7 +423,6 @@ func testRestartOnRebootLabelOnPodsUsingSriov(sriovPods []*provider.Pod) {
 		if labelValue != "true" {
 			tnf.ClaimFilePrintf("Pod %s is using SRIOV but the %s label value is not true.", pod, restartOnRebootLabel)
 			nonCompliantPods = append(nonCompliantPods, pod.String())
-			continue
 		}
 	}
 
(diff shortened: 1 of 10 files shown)


sebrandon1 (approved):

cnf-certification-test/networking/suite.go
```diff
@@ -394,3 +403,29 @@ func testNetworkPolicyDenyAll(env *provider.TestEnvironment) {
...
+		if !exist {
+			tnf.ClaimFilePrintf("Pod %s is using SRIOV but the label %s was not found.", pod, restartOnRebootLabel)
+			nonCompliantPods = append(nonCompliantPods, pod.String())
+			continue
+		}
+
+		if labelValue != "true" {
+			tnf.ClaimFilePrintf("Pod %s is using SRIOV but the %s label value is not true.", pod, restartOnRebootLabel)
+			nonCompliantPods = append(nonCompliantPods, pod.String())
+			continue
```
jmontesi, line 426: Unnecessary continue.

jmontesi (approved):

cnf-certification-test/networking/suite.go
greyerof: You're right. I've removed it. Thanks!

edcdavid (approved):

edcdavid (commented):
I think this is a better approach compared to current to check SRIOV+DPDK since a SRIOV device might not be attached to a pod yet but we should still be able to run this test.

PR #843, opened 2023-02-08
Remove version/url; Remove cnfextensions reference

sebrandon1:
- Removes the `version` and the `URL` fields from the identifier.
- Removes the cnfextensions external repo reference.  Documentation links still remain.
- Moves all of the identifiers to the `AddCatalogEntry()` func calls rather than manually building the structs.
- Adds the suite name to the list of tags shown in the catalog as well as used in Ginkgo.

Also related to:

https://github.com/test-network-function/test-network-function-claim/releases/tag/v1.0.12


edcdavid (commented):
Why not remove the Test Case Label and replace it with Unique ID / Label and use the former label format to uniquely identify testcases like:  access-control-cluster-role-bindings
--



image

edcdavid (approved):

bnshr (approved):

jmontesi (approved):

PR #861, opened 2023-02-15
Make docker config optional

sebrandon1:
- Moved the check for the Docker Config file into the suite.go.
- Removed the `LOCAL_DOCKERCFG` as a required parameter to the script.
- Removed `required:"true"` from the envconfig variable for the docker config.

### DIFF
--- a/cnf-certification-test/preflight/suite.go
+++ b/cnf-certification-test/preflight/suite.go
@@ -36,6 +36,12 @@ var _ = ginkgo.Describe(common.PreflightTestKey, func() {
 	})
 	ginkgo.ReportAfterEach(results.RecordResult)
 
+	// Add safeguard against running the tests if the docker config doesn't exist.
+	if env.GetDockerConfigFile() == "" {
+		logrus.Debug("Skipping the preflight suite because the Docker Config file is not provided.")
+		return
+	}
+
 	testPreflightContainers(&env)
 	if provider.IsOCPCluster() {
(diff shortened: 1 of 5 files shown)


rdavid (approved):

cnf-certification-test/preflight/suite.go
```diff
@@ -36,6 +36,12 @@ var _ = ginkgo.Describe(common.PreflightTestKey, func() {
 	})
 	ginkgo.ReportAfterEach(results.RecordResult)
 
+	// Add safeguard against running the tests if the docker config doesn't exist.
+	if env.GetDockerConfigFile() == "" {
```
greyerof, line 40: Shouldn't we also remove the `required:"true"` setting of the env var in the TestParameters struct (configuration.go)?
```
	PfltDockerconfig       string `split_words:"true" required:"true" envconfig:"PFLT_DOCKERCONFIG"`
```

jmontesi: Yes, it's still failing when the `PFLT_DOCKERCONFIG` env var is unset:

time="2023-02-16T11:07:58+01:00" level=fatal msg="required key PFLT_DOCKERCONFIG missing value"

sebrandon1: Ah whoops, yup I need to remove that as well good catch.

jmontesi (approved):

PR #868, opened 2023-02-20
Fixed operators discovery to avoid depending on installPlans.

greyerof:
The previous implementation was not adding operators without existing installPlans to the final operators to test list. As we've seen in real OCP clusters, for unknown reasons, OLM only keeps the last 5 InstallPlans in the same namespace.

- Removed a trace warning about CSVs without subscription not being able to be upgraded, which was wrong. As we're listing all the CSVs in the cluster, it means we're also getting the same CSV in every namespace for cluster-wide operators. But for those operators, the subscription exists in one namespace only, so that wrong trace was also appearing too many times. The same issue may happen with opeators installed in namespaces with OperatorGroup configured in multiNamespace mode (several entries in the spec.targetNamespace slice).

- The createOperators function has been changed to work on a list with unique CSVs by name. The final provider.Operator struct will contain a reference to the last CSV with that name that was seen by the function that creates the list of unique CSVs. TODO: Make the provider.Operator struct to have a reference to the CSV deployed in the namespace where the subscription was created.

- Added the subscription namespace to our custom Operator struct so the installPlans can be found there.

- Minor refactors related to CSV/Operators code.


ramperher:
Thanks @greyerof for all your effort here! DCI job is passing but in the testing environment we have. I'll try to run a job in the setup of the partner that is hitting this issue to validate if all operators are now discovered. I'll keep you posted, I'll try to do it on Thursday. But if you want to merge this before, don't hesitate, I can try the tests afterwards and report issues if it were.

sebrandon1 (approved):
`getUniqueCsvListByName` would be a good target for some unit tests but looks good.

rdavid (approved):

edcdavid (approved):

sebrandon1 (approved):

PR #891, opened 2023-02-28
Disable 4.0.x build in tnf-image.yaml

sebrandon1:
Comment out the 4.0.x build temporarily (maybe permanently 😄) because it is [failing](https://github.com/test-network-function/cnf-certification-test/actions/runs/4295224753/jobs/7485340046) the build process.  We don't plan on building any more 4.0.x builds in the future so I don't see a problem with turning it off temporarily.


rdavid:
It could be fixed in less radical way :-). We run 4.0.3 for all Mavenir. I'm looking in it.

rdavid:
@sebrandon1 See related PR:
 https://github.com/test-network-function/cnf-certification-test/pull/894

PR #892, opened 2023-02-28
Add CRC local testing

sebrandon1:


sebrandon1:
So I got this running, however it turns out that OpenShift Local needs 9+ GB of RAM free and the provided github runners only have 7GB of RAM.

sebrandon1:
Closing this, going to take another angle at this.

PR #916, opened 2023-03-09
Add new self-hosted PR tester YAML

sebrandon1:
Adds a new YAML for the new self-hosted runner.  Currently this is still running the same tests we use in the github hosted runners but I hope to expand on the strategy in the future and maybe use the self-hosted runner to run our QE tests.


rdavid:
It looks like a copy, what are the differences?

rdavid:
I guess it could be done without copy-paste and then support multiple instances of the same file. Probably YAML/environment variable is a solution.

PR #923, opened 2023-03-14
Safeguard against preflight tests running if not specified

sebrandon1:
Since the `preflight` tests are marked as `TagCommon` and are living under the `preflight` test suite, any labeling that occurs that includes those two strings allows the preflight tests to run prior to the `It` blocks being created.

If there is a label set such as `platform-alteration-isredhat-release` (like in QE), we gracefully return our way out of the function so as to not run the preflight tests if we don't need them.

### DIFF
--- a/cnf-certification-test/preflight/suite.go
+++ b/cnf-certification-test/preflight/suite.go
@@ -47,7 +47,7 @@ var _ = ginkgo.Describe(common.PreflightTestKey, func() {
 	// Safeguard against running the preflight tests if we are specifically targeting a certain label eg. in QE.
 	ginkgoConfig, _ := ginkgo.GinkgoConfiguration()
 	if !labelsAllowTestRun(ginkgoConfig.LabelFilter, []string{common.PreflightTestKey, identifiers.TagCommon}) {
-		logrus.Debug("LabelFilter is set but 'preflight' or 'common' tests are not targeted.  Issuing skip.")
+		logrus.Warn("LabelFilter is set but 'preflight' or 'common' tests are not targeted. Skipping the preflight tests.")
 		return
 	}
 
--- a/cnf-certification-test/preflight/suite.go
+++ b/cnf-certification-test/preflight/suite.go
@@ -17,6 +17,8 @@
 package preflight
 
 import (
+	"strings"
+
 	"github.com/onsi/ginkgo/v2"
 	plibRuntime "github.com/redhat-openshift-ecosystem/openshift-preflight/certification"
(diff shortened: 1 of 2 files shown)


cnf-certification-test/preflight/suite.go
```diff
@@ -42,6 +44,12 @@ var _ = ginkgo.Describe(common.PreflightTestKey, func() {
 		return
 	}
 
+	// Safeguard against running the preflight tests if we are specifically targeting a certain label eg. in QE.
+	ginkgoConfig, _ := ginkgo.GinkgoConfiguration()
+	if !labelsAllowTestRun(ginkgoConfig.LabelFilter, []string{common.PreflightTestKey, identifiers.TagCommon}) {
+		ginkgo.Skip("LabelFilter is set but 'preflight' or 'common' tests are not targeted.  Issuing skip.")
```
greyerof, line 50: Have you tested this? As per [ginkgo docs](https://onsi.github.io/ginkgo/#skipping-specs), Skip might not run here as this function is called uring the construction phase, not the running phase:

> You can call Skip in any subject or setup nodes. If called in a BeforeEach, Skip will skip the current spec. If called in a BeforeAll, Skip will skip all specs in the Ordered container (however, skipping an individual spec in an Ordered container does not skip subsequent specs). If called in a BeforeSuite, Skip will skip the entire suite.
>
> You cannot call Skip in a container node - Skip only applies during the Run Phase, not the Tree Construction Phase.

sebrandon1: You are correct!  It errored out when attempting the Skip() not inside of an It block.

I have adjusted it to just log and return.

Thanks!

cnf-certification-test/preflight/suite.go
```diff
@@ -42,6 +44,13 @@ var _ = ginkgo.Describe(common.PreflightTestKey, func() {
 		return
 	}
 
+	// Safeguard against running the preflight tests if we are specifically targeting a certain label eg. in QE.
+	ginkgoConfig, _ := ginkgo.GinkgoConfiguration()
+	if !labelsAllowTestRun(ginkgoConfig.LabelFilter, []string{common.PreflightTestKey, identifiers.TagCommon}) {
+		logrus.Debug("LabelFilter is set but 'preflight' or 'common' tests are not targeted.  Issuing skip.")
```
greyerof, line 50: I'd rather use logrus.Warn or logrus.Info here. Also, there's an extra space before the word "Issuing".

Issue #926, opened 2023-03-15
Nodes using latest OCP 4.13 (and 4.14) are based on CentOS Stream CoreOS, so that platform-alteration-ocp-node-os-lifecycle fails


ramperher:
Since today, latest OCP 4.13 (and OCP 4.14) versions are using CentOS Stream CoreOS in the OCP nodes, so that the `platform-alteration-ocp-node-os-lifecycle` test is failing because it does not consider it as a valid OS version.

Example of DCI job where this is happening:

https://www.distributed-ci.io/jobs/a229f9d1-0438-47d5-acc9-e77e4751c521/tests/37d1534f-3332-40d5-b7f3-8b8616126f8e

```
> Enter [BeforeEach] platform-alteration - /usr/tnf/tnf-src/cnf-certification-test/platform/suite.go:49 @ 03/15/23 01:26:36.856
< Exit [BeforeEach] platform-alteration - /usr/tnf/tnf-src/cnf-certification-test/platform/suite.go:49 @ 03/15/23 01:26:36.856 (0s)
> Enter [It] platform-alteration-ocp-node-os-lifecycle - /usr/tnf/tnf-src/cnf-certification-test/platform/suite.go:144 @ 03/15/23 01:26:36.856
STEP: Testing the control-plane and workers in the cluster for Operating System compatibility - /usr/tnf/tnf-src/cnf-certification-test/platform/suite.go:490 @ 03/15/23 01:26:36.856
Master Node worker-1 has been found to be running an incompatible operating system: CentOS Stream CoreOS 413.92.202303061740-0 (Plow)
Master Node worker-2 has been found to be running an incompatible operating system: CentOS Stream CoreOS 413.92.202303061740-0 (Plow)
Master Node worker-3 has been found to be running an incompatible operating system: CentOS Stream CoreOS 413.92.202303061740-0 (Plow)
Master Node master-0 has been found to be running an incompatible operating system: CentOS Stream CoreOS 413.92.202303061740-0 (Plow)
Master Node master-1 has been found to be running an incompatible operating system: CentOS Stream CoreOS 413.92.202303061740-0 (Plow)
Master Node master-2 has been found to be running an incompatible operating system: CentOS Stream CoreOS 413.92.202303061740-0 (Plow)
Master Node worker-0 has been found to be running an incompatible operating system: CentOS Stream CoreOS 413.92.202303061740-0 (Plow)
Number of control plane nodes running non-RHCOS based operating systems: 7
[...]

PR #928, opened 2023-03-15
Add support to CentOS Stream node OS

ramperher:
This is to address https://github.com/test-network-function/cnf-certification-test/issues/926

The idea is to consider that a node OS can be also CentOS Stream CoreOS, starting to be used in OCP 4.13/14 latest.

I have just simplified the inclusion of this new version to the minimum, but it's true we're missing here the following:

- A comparison between the CentOS Stream and OCP version, like the one we're doing for RHCOS with [this file](https://github.com/test-network-function/cnf-certification-test/blob/main/cnf-certification-test/platform/operatingsystem/files/rhcos_version_map).
- As a result, missing IsCSCOSCompatible function [here](https://github.com/test-network-function/cnf-certification-test/blob/main/pkg/compatibility/compatibility.go).

If you can guide me in these two points to have the full picture, I'd really appreciate that.

### DIFF
--- a/cnf-certification-test/platform/suite.go
+++ b/cnf-certification-test/platform/suite.go
@@ -530,6 +530,15 @@ func testNodeOperatingSystemStatus(env *provider.TestEnvironment) {
 					failedWorkerNodes = append(failedWorkerNodes, node.Data.Name)
 					continue
 				}
+			} else if node.IsCSCOS() {
+				// Get the short version from the node
+				shortVersion, err := node.GetCSCOSVersion()
+				if err != nil {
+					tnf.ClaimFilePrintf("Node %s failed to gather CentOS Stream CoreOS version. Error: %v", node.Data.Name, err)
+					failedWorkerNodes = append(failedWorkerNodes, node.Data.Name)
+					continue
+				}
+
 			} else if node.IsRHEL() {
 				// Get the short version from the node
 				shortVersion, err := node.GetRHELVersion()
--- a/pkg/provider/nodes.go
+++ b/pkg/provider/nodes.go
@@ -58,6 +58,10 @@ func (node *Node) IsRHCOS() bool {
 	return strings.Contains(strings.TrimSpace(node.Data.Status.NodeInfo.OSImage), rhcosName)
 }
 
+func (node *Node) IsCSCOS() bool {
+	return strings.Contains(strings.TrimSpace(node.Data.Status.NodeInfo.OSImage), cscosName)
+}
+
 func (node *Node) IsRHEL() bool {
 	return strings.Contains(strings.TrimSpace(node.Data.Status.NodeInfo.OSImage), rhelName)
 }
@@ -91,6 +95,19 @@ func (node *Node) GetRHCOSVersion() (string, error) {
 	}
 
 	return shortVersion, nil
+}
+
+func (node *Node) GetCSCOSVersion() (string, error) {
+	// Check if the node is running CoreOS or not
+	if !node.IsCSCOS() {
+		return "", fmt.Errorf("invalid OS type: %s", node.Data.Status.NodeInfo.OSImage)
+	}
+
+	// CentOS Stream CoreOS 413.92.202303061740-0 (Plow) --> 413.92.202303061740-0
+	splitStr := strings.Split(node.Data.Status.NodeInfo.OSImage, cscosName)
+	longVersionSplit := strings.Split(strings.TrimSpace(splitStr[1]), " ")
+
+	return longVersionSplit[0], nil
 }
 
 func (node *Node) GetRHELVersion() (string, error) {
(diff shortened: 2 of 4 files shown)


sebrandon1:
@ramperher Thanks for looking into this!  We modify the `rhcos_version_map` with this script:
https://github.com/test-network-function/cnf-certification-test/blob/main/script/rhcos_versions.sh

And I just realized the "latest" OCP version we're looking at is 4.10 so that might have to be modified to add the newer available versions.  Can you make that change in this PR as well?

ramperher:
> @ramperher Thanks for looking into this! We modify the `rhcos_version_map` with this script: https://github.com/test-network-function/cnf-certification-test/blob/main/script/rhcos_versions.sh
> 
> And I just realized the "latest" OCP version we're looking at is 4.10 so that might have to be modified to add the newer available versions. Can you make that change in this PR as well?
> 
> EDIT: The script runs daily as part of a cron and just makes a PR with any changes so once the versions are added it should be pretty much automated.

Thanks @sebrandon1 ! I can see your point. However, bad thing is that there's no stable version for OCP 4.13 and OCP 4.14 yet, so when running the script for these versions, it does not return anything:

```
$ ./script/rhcos_versions.sh 
+ set -e
+ CHANNELS=(4.14 4.13)
+ rm -f ./cnf-certification-test/platform/operatingsystem/files/rhcos_version_map
+ for i in "${CHANNELS[@]}"
++ curl -sH 'Accept: application/json' 'https://api.openshift.com/api/upgrades_info/v1/graph?channel=stable-4.14'
++ jq '.nodes[].version' -r
+ VERSIONS=
+ for i in "${CHANNELS[@]}"
++ curl -sH 'Accept: application/json' 'https://api.openshift.com/api/upgrades_info/v1/graph?channel=stable-4.13'
++ jq '.nodes[].version' -r
+ VERSIONS=
```

I can do the change for OCP 4.12 and 4.11, but that's all. For OCP 4.14 and 4.13, something different has to be done. Any ideas? Or should we leave it as it is in my current change?

sebrandon1 (approved):
Looks good!  Thanks!

ramperher:
> Looks good! Thanks!



> Looks good! Thanks!

thanks @sebrandon1 ! tl;dr for the others, we've agreed to include minimum functionality for this new OS until it's officialy used in a stable or candidate OCP version (it would happen in the next weeks, probably), so in that moment, we'll do more or less the same it's being done for rhcos right now: check if the OS version is in the internal database, and if the OCP version retrieved is related to the OS version used.

ramperher:
I'll run a test with OCP 4.13 latest now to see if this is behaving correctly

cnf-certification-test/platform/suite.go
```diff
@@ -530,6 +530,20 @@ func testNodeOperatingSystemStatus(env *provider.TestEnvironment) {
...
+				if err != nil {
+					tnf.ClaimFilePrintf("Node %s failed to gather CentOS Stream CoreOS version. Error: %v", node.Data.Name, err)
+					failedWorkerNodes = append(failedWorkerNodes, node.Data.Name)
+					continue
+				}
+
+				// Warning: CentOS Stream CoreOS has not been released yet in any
+				// OCP RC/GA versions, so for the moment, we cannot compare the
+				// version with the OCP one, or retrieve it on the internal database
+				tnf.ClaimFilePrintf("Node %s is using CentOS Stream CoreOS %s, which is not being used yet in any OCP RC/GA version. Relaxing the conditions to check the OS as a result", node.Data.Name, shortVersion)
```
rdavid, line 545:
```suggestion
				msg := `
Node %s is using CentOS Stream CoreOS %s, which is not being used yet in any
OCP RC/GA version. Relaxing the conditions to check the OS as a result.
`
				tnf.ClaimFilePrintf(msg, node.Data.Name, shortVersion)
```

rdavid:
Oops, I didn't mean to commit, only to suggest. I'm still learning Github UI :-)

rdavid (approved):

ramperher:
> Oops, I didn't mean to commit, only to suggest. I'm still learning Github UI :-)

No problem @rdavid :) I was just to include your proposal. I'll rerun the test with OCP 4.13 and if working, I'll merge.

ramperher:
I've run this DCI job to validate this in OCP versions <=4.12 and it's working: https://www.distributed-ci.io/jobs/4dfa2141-6b6f-405c-80c7-8e8a41a79fa8/jobStates?sort=date, I've rerun a job with OCP 4.13 to validate the CentOS Stream check.

By the way, I don't manage to fix the issues I'm having with the required Github checks, regarding the unit and smoke tests. They are complaining about unit tests that I have not modified in my change. Do you know how can I fix this? @sebrandon1

sebrandon1:
@ramperher Yes it's a pyxis problem in those tests.  Not much we can do as it usually comes back on its own.  `"Pyxis API was unable to create valid GraphQL query."`

PR #932, opened 2023-03-16
Admit that a control plane node can use CentOS Stream starting in OCP 4.13

ramperher:
This minor check was missed in https://github.com/test-network-function/cnf-certification-test/pull/928, OCP 4.13 job was failing in fact: https://www.distributed-ci.io/jobs/f0f8d733-30d5-4994-9fb9-6006c28e210e/tests


ramperher:
@sebrandon1 , I discovered this is not working fine because the condition I created was not right, that's why the DCI job failed :( I'll need to create a new PR for this, sorry for the triple-checking

sebrandon1:
No worries!

PR #933, opened 2023-03-16
Add default NA case for docker config discovery

sebrandon1:
The discovery of the LOCAL_DOCKERCFG variable needs a default option (NA).  If the `PFLT_DOCKERCONFIG` variable does not get set, and the `$HOME/.docker/config.json` does not exist, then the variables get set to `NA`.

Prior to this, we were seeing:

```
Mounting 1 kubeconfig volume(s):
-v /opt/ocp/r207-compact/kubeconfig:/usr/tnf/kubeconfig/config:Z
Mounting 1 dockercfg volume(s):
-v 
```

And the `-v` docker run flag was not populated causing an error: `Error: invalid reference format`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -120,6 +120,8 @@ install-tools:
 # Install linters
 install-lint:
 	curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b ${GO_PATH}/bin ${GOLANGCI_VERSION}
+	
+install-shfmt:
 	go install mvdan.cc/sh/v3/cmd/shfmt@latest
 
 vet:
--- a/script/run-container.sh
+++ b/script/run-container.sh
@@ -97,7 +97,7 @@ done
 
 # Explode loaded DOCKERCFG
 # shellcheck disable=SC2162 # Read without -r will mangle backslashes.
-if [[ $LOCAL_DOCKERCFG != 'NA' ]]; then
+if [[ $LOCAL_DOCKERCFG != NA ]]; then
 	IFS=: read -a local_dockercfg_paths <<<"$LOCAL_DOCKERCFG"
 
 	declare -a container_tnf_dockercfg_paths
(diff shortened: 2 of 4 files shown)


Makefile
```diff
@@ -120,6 +120,8 @@ install-tools:
 # Install linters
 install-lint:
 	curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b ${GO_PATH}/bin ${GOLANGCI_VERSION}
+	
+install-shfmt:
```
rdavid, line 124: It's strange, I've added it here:
 https://github.com/test-network-function/cnf-certification-test/commit/7906f8c9eee95fe03531d86a8c126fe371e5ad69

script/run-container.sh
```diff
@@ -124,8 +129,10 @@ container_tnf_kubeconfig_volumes_cmd_args=$(printf -- "-v %s " "${container_tnf_
 
 # Construct new $DOCKERCFG env variable containing all paths to dockercfgs mounted to the container.
 # This environment variable is passed to the TNF container
-CONTAINER_TNF_DOCKERCFG=$(join_paths "${container_tnf_dockercfg_paths[@]}")
-container_tnf_dockercfg_volumes_cmd_args=$(printf -- "-v %s " "${container_tnf_dockercfg_volume_bindings[@]}")
+if [[ $LOCAL_DOCKERCFG != 'NA' ]]; then
```
rdavid, line 132:
```suggestion
if [[ $LOCAL_DOCKERCFG != NA ]]; then
```

rdavid (approved):

Makefile
sebrandon1: Yeah I wanted to split it out because I use `brew` for my golangci-lint installations.

script/run-container.sh
```diff
@@ -94,19 +97,21 @@ done
 
 # Explode loaded DOCKERCFG
 # shellcheck disable=SC2162 # Read without -r will mangle backslashes.
-IFS=: read -a local_dockercfg_paths <<<"$LOCAL_DOCKERCFG"
+if [[ $LOCAL_DOCKERCFG != 'NA' ]]; then
```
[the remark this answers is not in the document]
sebrandon1: Done.

script/run-container.sh
sebrandon1: Done.

Makefile
rdavid: There is a `brew` formula for `shfmt` as well.

PR #936, opened 2023-03-19
Add Hadolint, a Dockerfile linter

rdavid:
- https://github.com/hadolint/hadolint
- https://github.com/hadolint/hadolint-action

### DIFF
--- a/.github/workflows/pre-main.yaml
+++ b/.github/workflows/pre-main.yaml
@@ -59,6 +59,14 @@ jobs:
       - name: Install golangci-lint
         run: make install-lint
 
+      - name: Install hadolint
+        run: |
+          curl \
+            --location \
+            --output /usr/local/bin/hadolint \
+            https://github.com/hadolint/hadolint/releases/download/v2.12.0/hadolint-Linux-x86_64
+          chmod +x /usr/local/bin/hadolint
+
       - name: Install shfmt
         run: make install-shfmt
 
@@ -68,10 +76,14 @@ jobs:
       - name: make vet
         run: make vet
   shell-linters:
-    name: Shellcheck and shfmt
+    name: Hadolint, shellcheck and shfmt
     runs-on: ubuntu-22.04
     steps:
       - uses: actions/checkout@v3
+      - uses: hadolint/hadolint-action@v3.1.0
+        with:
+          dockerfile: Dockerfile
+          recursive: true
       - uses: ludeeus/action-shellcheck@master
       - uses: mfinelli/setup-shfmt@v2
       - run: shfmt -d *.sh script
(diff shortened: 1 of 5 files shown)


.github/workflows/pre-main.yaml
```diff
@@ -68,10 +76,14 @@ jobs:
       - name: make vet
         run: make vet
   shell-linters:
-    name: Shellcheck and shfmt
+    name: Hadolint, shellcheck and shfmt
     runs-on: ubuntu-22.04
     steps:
       - uses: actions/checkout@v3
+      - uses: hadolint/hadolint-action@v3.1.0
```
sebrandon1, line 83: Is it possible to just use this as `v3`?  So we only have to worry about major versions?  See the other actions for example.

Dockerfile
```diff
@@ -1,38 +1,47 @@
...
 # Install dependencies
-RUN yum install -y gcc git jq make wget
+RUN \
+	yum install --assumeyes \
+		gcc-8.5.0 \
+		git-2.31.1 \
+		jq-1.6 \
+		cmake-3.20.2 \
+		wget-1.19.5 \
+	&& yum clean all
```
sebrandon1, line 16: I'm torn on whether or not we need to pin these versions.  I like that we get the latest versions whenever we build our image.

rdavid: `hadolint` doesn't like `latest`, I don't thing the versions are going to be changed in `ubi:8.7`. What is the profit to work with latest? By the way the latest is ubi9.

sebrandon1: Is there a way to turn off the version checking in hadolint?

rdavid: There is a way, sure.

sebrandon1: Let's do that for now, and decide later if we want to turn on the version checking.

rdavid: In theory every image build is different today because it uses the latest packages. What if we need to reproduce exactly the same run we did a year ago with a certain version of curl, for example. For that we have release versions for go modules, for that packages and images should have versions too. And then we lock all software versions (go modules, dependent libraries, images and packages) together under the TNF version, we test and certify it. Under this lock we can always be sure we run the same software for the release.

sebrandon1: Yup makes sense.  We should probably have some sort of automation that runs weekly(?) for updating these image level dependencies.  I wonder if dependabot has some sort of option to scan Dockerfiles.

sebrandon1: https://dev.to/oracle2025/how-to-keep-a-dockerfile-updated-with-dependabot-1mdn

.github/workflows/pre-main.yaml
rdavid: I've tried without success, it needs the full version number.

ramperher (approved):
@rdavid , feel free to merge this if needed, DCI failure is not related to your change, it's related to https://github.com/test-network-function/cnf-certification-test/pull/937.

PR #937, opened 2023-03-20
add the calssification catagory of test case (farEdge Telco NonTelco)

aabughosh:


sebrandon1 (approved):

cnf-certification-test/identifiers/identifiers.go
```diff
@@ -173,6 +173,12 @@ The label value is not important, only its presence.`,
 		NoDocumentedProcess,
 		bestPracticeDocV1dot3URL+" Section 5.2",
 		true,
+		claim.CategoryClassification{
+			FarEdge:  "Mandatory",
```
edcdavid, line 177: should'n claim.CategoryClassification just be a map[string] string like:
type CategoryClassification struct {
   map[string] string
}
also, we could define FarEdge, Telco, NonTelco, Extended (if using a map) as constants, same with Optional, Mandatory

sebrandon1: Not a bad idea, gives a lot of flexibility if we need to add additional scenarios in the future.

cnf-certification-test/preflight/suite.go
```diff
@@ -154,7 +159,12 @@ func generatePreflightOperatorGinkgoTest(testName, description, suggestion strin
 	// Brute force-ish type of method.
 
 	// Store the test names into the Catalog map for results to be dynamically printed
-	aID := identifiers.AddCatalogEntry(testName, common.PreflightTestKey, description, suggestion, "", "", "", false, identifiers.TagCommon)
+	aID := identifiers.AddCatalogEntry(testName, common.PreflightTestKey, description, suggestion, "", "", "", false, map[string]string{
+		"FarEdge":  "Optional",
```
edcdavid, line 163: add constants for all strings in identifiers.go

edcdavid (approved):

ramperher:
DCI job is failing because this change needs to be rebased, that's why we have ERROR.

PR #934, opened 2023-03-16
Admit that a control plane node can use CentOS Stream starting in OCP 4.13 (II)

ramperher:
This minor check was missed in https://github.com/test-network-function/cnf-certification-test/pull/928, OCP 4.13 job was failing in fact: https://www.distributed-ci.io/jobs/f0f8d733-30d5-4994-9fb9-6006c28e210e/tests

https://github.com/test-network-function/cnf-certification-test/pull/932 was created to fix that but condition was wrong. This should fix everything.

PR #944, opened 2023-03-21
Handle possible errors which are not critical

sebrandon1:
Follow up to #934 

The script was [hitting an error](https://github.com/test-network-function/cnf-certification-test/actions/runs/4474518500/jobs/7863067358) with `oc adm` that was causing it to crash out prior to completion.  We handle the blank string so there's no problem with the error and continuing.

I re-ran the script to bring in the latest values with this PR.


script/rhcos_versions.sh
```diff
@@ -1,7 +1,6 @@
 #!/usr/bin/env bash
 
 set -x
-set -e
```
rdavid, line 4: Leave it, just mask the failure:
```
RHCOSVERSION="$(oc adm release info "${VERSION}" -o 'jsonpath={.displayVersions.machine-os.Version}') 2>&1" ||
  printf 'Continue with the error %s.\n' "$RHCOSVERSION"
```

rdavid: Do you mean the script exits anyway?

sebrandon1: No I copied and pasted your suggestion into my code and it doesn't run the way it should.

rdavid: Try now, show me the output.

sebrandon1: Edits committed.  Thanks!

PR #945, opened 2023-03-21
Add some more logging; safeguard against empty InstallPlans

sebrandon1:
There has been some recent DCI runs where it seems like an empty slice of InstallPlans is throwing an error.  We should probably have a safeguard against this sort of scenario.

### DIFF
--- a/cnf-certification-test/preflight/suite.go
+++ b/cnf-certification-test/preflight/suite.go
@@ -80,6 +80,8 @@ func testPreflightOperators(env *provider.TestEnvironment) {
 		}
 	}
 
+	logrus.Infof("Completed running preflight operator tests for %d operators", len(env.Operators))
+
 	// Handle Operator-based preflight tests
 	// Note: We only care about the `testEntry` variable below because we need its 'Description' and 'Suggestion' variables.
 	for testName, testEntry := range getUniqueTestEntriesFromOperatorResults(env.Operators) {
@@ -100,6 +102,8 @@ func testPreflightContainers(env *provider.TestEnvironment) {
 			logrus.Fatalf("failed running preflight on image: %s error: %v", cut.Image, err)
 		}
 	}
+
+	logrus.Infof("Completed running preflight container tests for %d containers", len(env.Containers))
 
 	// Handle Container-based preflight tests
 	// Note: We only care about the `testEntry` variable below because we need its 'Description' and 'Suggestion' variables.
--- a/pkg/provider/containers.go
+++ b/pkg/provider/containers.go
@@ -84,15 +84,12 @@ func (c *Container) SetPreflightResults(preflightImageCache map[string]plibRunti
 		return nil
 	}
 
-	var results plibRuntime.Results
 	opts := []plibContainer.Option{}
 	opts = append(opts, plibContainer.WithDockerConfigJSONFromFile(env.GetDockerConfigFile()))
 	if env.IsPreflightInsecureAllowed() {
 		logrus.Info("Insecure connections are being allowed to preflight")
 		opts = append(opts, plibContainer.WithInsecureConnection())
 	}
-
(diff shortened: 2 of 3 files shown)


pkg/provider/containers.go
```diff
@@ -108,6 +105,8 @@ func (c *Container) SetPreflightResults(preflightImageCache map[string]plibRunti
 	logger := stdr.New(checklogger)
 	ctx = logr.NewContext(ctx, logger)
 
+	check := plibContainer.NewCheck(c.Image, opts...)
```
sebrandon1, line 108: For reviewers, I made this change to make it look exactly like the `container.SetPreflightResults` func.

pkg/provider/operators.go
```diff
@@ -70,6 +70,11 @@ func (op *Operator) String() string {
 
 //nolint:funlen
 func (op *Operator) SetPreflightResults(env *TestEnvironment) error {
+	if len(op.InstallPlans) == 0 {
+		logrus.Warnf("%s has no InstallPlans. Skipping setting preflight results", op.String())
+		return nil
```
sebrandon1, line 75: I'm not sure if I want to actually throw an error here, but rather just skip the operator and exclude it from the preflight tests at this point.  It's open to debate.

cnf-certification-test/preflight/suite.go
```diff
@@ -80,6 +80,8 @@ func testPreflightOperators(env *provider.TestEnvironment) {
 		}
 	}
 
+	logrus.Infof("Completed running preflight operator tests for %d operators", len(env.Operators))
```
rdavid, line 83: Could `env.Operators` be `nil`? If it is a nil slice, `len` returns `0`. Maybe log zero operators explicitly?

pkg/provider/operators.go
```diff
@@ -95,9 +100,12 @@ func (op *Operator) SetPreflightResults(env *TestEnvironment) error {
 	ctx = logr.NewContext(ctx, logger)
 
 	check := plibOperator.NewCheck(bundleImage, indexImage, oc.KubeConfig, opts...)
-	results, err := check.Run(ctx)
-	if err != nil {
-		return err
+	var results plibRuntime.Results
```
rdavid, line 103: What is a rational to have more lines? `res, err := check.Run(ctx)` is cool.

cnf-certification-test/preflight/suite.go
sebrandon1: It works to just check the length.  `len` handles the nil properly.

https://go.dev/play/p/xuipYDlnHx4

pkg/provider/operators.go
sebrandon1: Fixed!

ramperher (approved):
Thanks Brandon, yes, let's check this to see how it goes

cnf-certification-test/preflight/suite.go
rdavid: `len` doesn't handle `nil` pointer:
 https://go.dev/play/p/5O0uDUlFiU4
 
 The code works. I thought about improving the logging:
```
if ops := len(env.Operators); ops > 0 {
  logrus.Infof("Completed running preflight operator tests for %d operators", ops)
else {
  logrus.Info("There is no operator to run preflight operator tests")
}
```

rdavid (approved):

cnf-certification-test/preflight/suite.go
sebrandon1: No it doesn't.  The variable in your code is `untyped`.  This is why it fails.

rdavid: https://go.dev/play/p/Msjo6-ziBsY

sebrandon1: Yeah Go knows that `*int` isn't a slice so `len` doesn't handle it.

PR #966, opened 2023-03-27
Add CRC local testing

sebrandon1:
Related issues found while testing:
- https://github.com/actions/runner-images/issues/7276


sebrandon1:
Marking this as closed as I think the macos hosted runners are just too undersized and cannot handle this type of workload.

PR #967, opened 2023-03-28
Make the namespace for the debug daemonset configurable.

greyerof:
The namespace for the debug daemonset is now configurable via yaml file. In case the field is not set, the namespace "cnf-suite" will be used.

Example:
debugDaemonSetNamespace: cnf-cert

+ Minor refactors.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -105,3 +105,9 @@ skipScalingTestStatefulSetNames:
   - name: "statefulset1"
     namespace: "tnf"
 ```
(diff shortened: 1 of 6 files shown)


sebrandon1 (approved):

sebrandon1:
Failure looks unrelated to this change: `unable to create session key: disk quota exceeded`.  Seems to be happening to multiple PRs.

edcdavid (approved):

PR #968, opened 2023-03-28
Add test with fake statefulset APIs

sebrandon1:
Using the [fake](https://pkg.go.dev/k8s.io/client-go/kubernetes/fake) package to spoof some APIs.

This brings the overall `provider` package to:
`ok  	github.com/test-network-function/cnf-certification-test/pkg/provider	0.733s	coverage: 52.3% of statements`

`GetUpdatedStatefulset` was uncovered as far as unit testing.  Other funcs will be covered in separate PRs to keep changes separate.

### DIFF
--- a/pkg/provider/statefulsets_test.go
+++ b/pkg/provider/statefulsets_test.go
@@ -22,6 +22,9 @@ import (
 	"github.com/stretchr/testify/assert"
 	appsv1 "k8s.io/api/apps/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/kubernetes/fake"
+	k8stesting "k8s.io/client-go/testing"
 )
 
 func TestStatefulsetToString(t *testing.T) {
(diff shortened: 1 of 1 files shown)


pkg/provider/statefulsets_test.go
```diff
@@ -92,3 +95,35 @@ func TestIsStatefulSetReady(t *testing.T) {
...
+func TestGetUpdatedStatefulset(t *testing.T) {
+	testCases := []struct {
+		testNamespace string
+	}{
+		{ // Test Case #1 - Test with valid namespace 'testNS1'
+			testNamespace: "testNS1",
+		},
+		{ // Test Case #2 - Test with valid namespace 'testNS2'
+			testNamespace: "testNS2",
+		},
```
greyerof, line 108: A test case for the error output of `GetUpdatedStatefulset` would also be nice to have.

greyerof (approved):

pkg/provider/statefulsets_test.go
sebrandon1: Good call, I'll add one.

rdavid (approved):

PR #969, opened 2023-03-28
Add more coverage for containers.go

sebrandon1:
`HasIgnoredContainerName` func was missing unit tests.  This file has everything covered by the `SetPreflightResults` func now.

### DIFF
--- a/pkg/provider/containers_test.go
+++ b/pkg/provider/containers_test.go
@@ -140,3 +140,39 @@ func TestHasExecProbes(t *testing.T) {
 		assert.Equal(t, tc.expectedOutput, tc.testContainer.HasExecProbes())
 	}
 }
+
+func TestHasIgnoredContainerName(t *testing.T) {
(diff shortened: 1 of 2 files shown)


pkg/provider/containers_test.go
```diff
@@ -140,3 +140,40 @@ func TestHasExecProbes(t *testing.T) {
...
+				},
+			},
+			expectedOutput: true,
+		},
+	}
+
+	for _, tc := range testCases {
+		assert.Equal(t, tc.expectedOutput, tc.testContainer.HasIgnoredContainerName())
+	}
+
```
rdavid, line 178: There is no need for vertical space.

rdavid (approved):

pkg/provider/containers_test.go
sebrandon1: Good call, fixed.

greyerof (approved):

PR #985, opened 2023-04-04
Add MacOS to CI matrix

sebrandon1:
Notable changes:
- Changed `make install-lint` to use `go install` instead of curling the binary.
- Added additional `Install hadolint` step in pre-main.yaml that uses brew for macOS installs.
- Adds `macos-latest` to matrix for running tests.

Since we are running these scripts locally as well as in CI, they might as well pass both places.

### DIFF
--- a/.github/workflows/pre-main.yaml
+++ b/.github/workflows/pre-main.yaml
@@ -28,19 +28,14 @@ env:
 jobs:
   lint:
     name: Run Linter and Vet
-    runs-on: ubuntu-22.04
+    runs-on: ${{ matrix.os }}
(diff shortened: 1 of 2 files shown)


sebrandon1:
Removing the `WIP` flag from this PR.  It seems to be working!  I'll edit the `required` checks once this merges to remove the `Run Linter and Vet` and `Run Unit Tests` checks and replace them with the linux and mac versions.

sebrandon1:
Also, DCI seems to be failing on their side with something.

rdavid (approved):

PR #991, opened 2023-04-05
Implement Test Exclusive CPU pool

bnshr:


sebrandon1:
Does this also need a `make build-catalog-md` ran prior?

bnshr:
> Does this also need a `make build-catalog-md` ran prior?

I dont know. What is that for?

sebrandon1:
@bnshr This needs to be rebased against `main` for the DCI job to kick off successfully.

PR #999, opened 2023-04-06
Safeguard against NA docker config

sebrandon1:
#831 added this safeguard but the check for the empty string needs to also include the value "NA".  This check was added in #861.

Should be an easy fix and we can release 4.2.2 to pick it up.

For example, here are the ENV vars from a 4.2.1 run:
```
"Environment: {Home:/root Kubeconfig:/usr/tnf/kubeconfig/config ConfigurationPath:/usr/tnf/config/tnf_config.yml NonIntrusiveOnly:true LogLevel:trace OfflineDB:/usr/offline-db AllowPreflightInsecure:false PfltDockerconfig:NA}"
```

### DIFF
--- a/cnf-certification-test/preflight/suite.go
+++ b/cnf-certification-test/preflight/suite.go
@@ -39,7 +39,7 @@ var _ = ginkgo.Describe(common.PreflightTestKey, func() {
 	ginkgo.ReportAfterEach(results.RecordResult)
 
 	// Add safeguard against running the tests if the docker config doesn't exist.
-	if env.GetDockerConfigFile() == "" {
+	if env.GetDockerConfigFile() == "" || env.GetDockerConfigFile() == "NA" {
 		logrus.Debug("Skipping the preflight suite because the Docker Config file is not provided.")
 		return
 	}


rdavid (approved):

cnf-certification-test/preflight/suite.go
```diff
@@ -39,7 +39,7 @@ var _ = ginkgo.Describe(common.PreflightTestKey, func() {
 	ginkgo.ReportAfterEach(results.RecordResult)
 
 	// Add safeguard against running the tests if the docker config doesn't exist.
-	if env.GetDockerConfigFile() == "" {
+	if env.GetDockerConfigFile() == "" || env.GetDockerConfigFile() == "NA" {
```
edcdavid, line 42: Shouldn't we check if the file exists instead?

sebrandon1: Yeah that's done in the script prior to kicking off the suite.  The bug in this case is just not handling the "NA" that we purposefully add when the dockerconfig is missing.

https://github.com/test-network-function/cnf-certification-test/blob/main/run-tnf-container.sh#L141-L158

edcdavid (approved):

PR #1018, opened 2023-04-12
Adjust Makefile build process

sebrandon1:
To try and simplify the build process I want to remove the `install-tools` as a prerequisite for running `build-cnf-tests`.

Also, there were places where we were building the `classification-js` make path and that shouldn't have anything to do with the actual binary.

We still run `install-tools` as part of the environment setup for the CI so this just decouples it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -101,19 +101,16 @@ build-catalog-json: build-tnf-tool
 	./tnf generate catalog json > catalog.json
 
 # generate the test catalog in Markdown
-build-catalog-md: build-tnf-tool
+build-catalog-md: build-tnf-tool classification-js
 	./tnf generate catalog markdown > CATALOG.md
 
 # build the CNF test binary
-build-cnf-tests: install-tools
+build-cnf-tests:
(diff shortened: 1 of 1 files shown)


rdavid (approved):

Makefile
```diff
@@ -101,19 +101,16 @@ build-catalog-json: build-tnf-tool
 	./tnf generate catalog json > catalog.json
 
 # generate the test catalog in Markdown
-build-catalog-md: build-tnf-tool
+build-catalog-md: build-tnf-tool classification-js
```
greyerof, line 104: Some comments here:
- Where is "make build-catalog-md" done now?
- Is "classification-js" actually needed for building the CATALOG.md? Otherwise I don't think it should be a dependency here. In fact, the "classification-js:" rule (line 162) should actually have the "build-tnf-tool" rule as a dependency (like build-catalog-md).
- Maybe we should include both "make build-catalog-md" and "make classification-js" under the "build:" rule or creating a new one like "artifacts:", or "docs:".

PR #1019, opened 2023-04-12
Adding checkbox by state (passed/skipped/failed) to filter out results by state

edcdavid:
Added one checkbox by state (passed/skipped/failed) in the summary column. checked box mean that the result in the corresponding state are shown, unchecked box means that they are hidden. Also fixed a bug where the "all" table was displayed twice


sebrandon1 (approved):
Tested it out, looks good!

aabughosh (approved):

PR #1027, opened 2023-04-16
Add linter for Makefile

rdavid:


Makefile
```diff
@@ -64,24 +65,15 @@ build-tnf-tool:
...
-	rm -f all-releases.txt
+	go clean && rm -f all-releases.txt cover.out claim.json cnf-certification-test/claim.json \
+		cnf-certification-test/claimjson.js cnf-certification-test/cnf-certification-test.test \
+		cnf-certification-test/cnf-certification-tests_junit.xml \
+		cnf-certification-test/results.html jsontest-cli latest-release-tag.txt \
+		release-tag.txt test-out.json tnf
 
 # Run configured linters
 lint:
+	checkmake Makefile
```
edcdavid, line 76: @line 111, it would be great to add installation for other linters:
# Install linters
install-lint:
	go install github.com/golangci/golangci-lint/cmd/golangci-lint@${GOLANGCI_VERSION}

edcdavid (approved):

sebrandon1 (approved):

rdavid:
@sebrandon1 Why linter job is in waiting status? Should I ignore it?

Makefile
rdavid: @edcdavid The problem is they have different install command for different OS, need to think about, will do another PR.

sebrandon1:
Yup just ignore we'll fix after this merges.

PR #1037, opened 2023-04-20
Add gradetool support

sebrandon1:
Utilizing the new [gradetool](https://github.com/test-network-function/gradetool) repository and image to test our CI's pass and fail.

The `gradetool` takes in the claim.json generated by PRs and tests it against the generated_policy.json which is built by the contents of the CATALOG.md.

### DIFF
--- a/.github/workflows/pre-main.yaml
+++ b/.github/workflows/pre-main.yaml
@@ -13,6 +13,8 @@ env:
   TNF_IMAGE_TAG: unstable
   OCT_IMAGE_NAME: testnetworkfunction/oct
   OCT_IMAGE_TAG: latest
+  GRADETOOL_IMAGE_NAME: testnetworkfunction/gradetool
+  GRADETOOL_IMAGE_TAG: latest
   TNF_CONTAINER_CLIENT: docker
   TNF_NON_INTRUSIVE_ONLY: false
   TNF_ALLOW_PREFLIGHT_INSECURE: false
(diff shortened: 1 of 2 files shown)


.github/workflows/pre-main.yaml
```diff
@@ -245,6 +245,28 @@ jobs:
...
+      - name: Run gradetool on the claim.json
+        run: |
+          make build
+          ./gradetool -p ${GITHUB_WORKSPACE}/generated_policy.json -o ${GITHUB_WORKSPACE}/results.txt -r ${GITHUB_WORKSPACE}/cnf-certification-test/claim.json
+        working-directory: gradetool
+
+      - name: Check that their are 0 failed tests in the gradetool results
+        run: |
+          if $(jq '.[] | .Fail | length' ${GITHUB_WORKSPACE}/results.txt | grep -q 0); then
+            echo "OK"
```
rdavid, line 263:
```suggestion
            echo OK
```

.github/workflows/pre-main.yaml
```diff
@@ -245,6 +245,28 @@ jobs:
...
+          else
+            echo "Gradetool has found failing tests in the following:"
+            jq '.[] | .Fail | .[].id' ${GITHUB_WORKSPACE}/results.txt
```
rdavid, line 266:
```suggestion
            jq '.[] | .Fail | .[].id' "${GITHUB_WORKSPACE}"/results.txt
```

rdavid (approved):

greyerof (approved):
I think there's a lot of room for improvement here, but it's a good starting point.

Just a couple of questions:
- Is it really necessary to have this grade tool in a separate repo?
- If yes, can't we just run the gradetool container directly instead of "make build && ./gradetool ...". I believe that the only thing that we should build here is our own repo's artifacts (tnf cmd and the ginkgo test suite binary).

PR #1045, opened 2023-04-24
Rewrite ClusterRoleBindings test; fix RoleBindings test

sebrandon1:
- `testPodClusterRoleBindings` needed to be re-written to check for the pod using cluster roles in general.  It is **not allowed** per the requirements documents.

- `testPodRoleBindings` needed some updates as far as how it was looking for role bindings attached to the pod.  In the previous implementation we were just comparing the service account name to the role binding name.  We needed to be looking at the `Subjects` of the role binding to see if the service account matches up.

- `GetClusterRoleBindings()` and `GetRoleBindings()` were looking at the wrong object types from the API.  These functions were moved inside of the `autodiscover` portion of the code because we don't need any context for looking these up and we would avoid calling the API multiple times during the actual test.

image


ramperher:
Just note that access-control-cluster-role-bindings is currently failing on DCI: https://www.distributed-ci.io/jobs/3768e67d-bac4-455b-8db4-a8acc5b61083/tests/1a51ab34-803c-43d8-b5c8-3e0f19acb4ae

```
> Enter [BeforeEach] access-control - /usr/tnf/tnf-src/cnf-certification-test/accesscontrol/suite.go:54 @ 04/27/23 21:57:35.097
< Exit [BeforeEach] access-control - /usr/tnf/tnf-src/cnf-certification-test/accesscontrol/suite.go:54 @ 04/27/23 21:57:35.097 (0s)
> Enter [It] access-control-cluster-role-bindings - /usr/tnf/tnf-src/cnf-certification-test/accesscontrol/suite.go:154 @ 04/27/23 21:57:35.097
STEP: Pods should not have ClusterRoleBindings - /usr/tnf/tnf-src/cnf-certification-test/accesscontrol/suite.go:446 @ 04/27/23 21:57:35.097
STEP: Testing cluster role binding for pod: test-0 namespace: production-cnf - /usr/tnf/tnf-src/cnf-certification-test/accesscontrol/suite.go:452 @ 04/27/23 21:57:35.097
pod: test-0 ns: production-cnf is using a cluster role binding
STEP: Testing cluster role binding for pod: test-1 namespace: production-cnf - /usr/tnf/tnf-src/cnf-certification-test/accesscontrol/suite.go:452 @ 04/27/23 21:57:35.101
pod: test-1 ns: production-cnf is using a cluster role binding
Non-compliant []string: [test-0 test-1]
[FAILED] Number of Non-compliant []string = 2
In [It] at: /usr/tnf/tnf-src/pkg/testhelper/testhelper.go:162 @ 04/27/23 21:57:35.103
< Exit [It] access-control-cluster-role-bindings - /usr/tnf/tnf-src/cnf-certification-test/accesscontrol/suite.go:154 @ 04/27/23 21:57:35.103 (6ms)
> Enter [ReportAfterEach] access-control - /usr/tnf/tnf-src/cnf-certification-test/accesscontrol/suite.go:57 @ 04/27/23 21:57:35.103
< Exit [ReportAfterEach] access-control - /usr/tnf/tnf-src/cnf-certification-test/accesscontrol/suite.go:57 @ 04/27/23 21:57:35.103 (0s)
```

ramperher:
> @ramperher We may have to change something on your end as far as making this test pass.
> 
> `pod: test-0 ns: production-cnf is using a cluster role binding`
> 
> From the requirements docs it is stated that `CNFs must not have cluster roles` so I had to rewrite the test to actually do what it's supposed to do. Prior to this change the implementation was incorrect in many spots.

Thanks @sebrandon1 for checking this. We'll need to check this, because we're currently using ClusterRoles, e.g., to attach the VZs recommended SCC to test them: https://github.com/redhat-cip/dci-openshift-app-agent/blob/master/samples/tnf_test_example/hooks/install.yml#L32-L92, so we'll need to check the impact of this change in order to refactor everything.

sebrandon1:
Marking this issue as `blocked`.  We have a couple of things that need to be addressed before this can merge.
- The `crd-operator-scaling` repo needs to be updated to remove the cluster role bindings it is currently using.  This is installed via the partner `make install` path and the tests fail because of it.
- The DCI team is going to work to remove the cluster role bindings on their side as well so the DCI jobs start passing again.

cnf-certification-test/accesscontrol/suite.go
```diff
@@ -410,44 +410,58 @@ func testPodRoleBindings(env *provider.TestEnvironment) {
...
+			logrus.Errorf("failed to determine if pod %s/%s is using a cluster role binding: %v", put.Namespace, put.Name, err)
 			failedPods = append(failedPods, put.Name)
 		}
 
-		if len(clusterRoleBindings) > 0 {
-			logrus.Warnf("Pod: %s/%s has the following cluster role bindings: %s", put.Namespace, put.Name, clusterRoleBindings)
-			tnf.ClaimFilePrintf("Pod: %s/%s has the following cluster role bindings: %s", put.Namespace, put.Name, clusterRoleBindings)
+		// Pod was found to be using a cluster role binding.  This is not allowed.
+		// Flagging this pod as a failed pod.
+		if !result {
```
ramperher, line 461: The condition here should be `if result`, and not `if !result`, shouldn't it?
Because, in the function called `IsUsingClusterRoleBinding`, it returns `true` when a pod is using a ClusterRoleBinding, `else` if not using it. And here, we are checking the opposite.
That's why the DCI job is failing: in fact, we've removed all the ClusterRoleBindings and ClusterRoles from the pods under test, and despite that, the job is failing in the access-control-cluster-role-bindings test. And it's because of this condition.

ramperher (changes requested):

pkg/provider/pods.go
```diff
@@ -379,3 +382,42 @@ func (p *Pod) IsUsingSRIOV() (bool, error) {
...
+	// tied to any of them.  If it is, then it returns true, otherwise it returns false.
+
+	// Short circuit.  If there is no serviceAcccountName set we can skip the rest of the checks.
+	// Note: This currently does not handle where projected volumes are used to access service account tokens.
+	if p.Pod.Spec.ServiceAccountName == "" {
+		return false, nil
+	}
+
+	// Get all of the service accounts in the pod's namespace.
+	serviceAccounts, err := rbac.GetServiceAccountsFromNamespace(p.Pod.Namespace)
```
greyerof, line 397: Is this really necessary? As far as I know, If the pod has a ServiceAccountName, that SA must exist inside the pod's namespace, otherwise the pod wouldn't have been deployed.
```
	podServiceAccountName := p.Pod.Spec.ServiceAccountName

	for crbIndex := range clusterRoleBindings {
		for _, subject := range clusterRoleBindings[crbIndex].Subjects {
			if subject.Kind == rbacv1.ServiceAccountKind && subject.Name == podServiceAccountName && subject.Namespace == p.Namespace{
				tnf.ClaimFilePrintf("Pod %s has service account %s that is tied to cluster role binding %s", p.Pod.Name, podServiceAccountName, clusterRoleBindings[crbIndex].Name)
				return true, nil
			}
		}
	}
```
Also note that I also added the check for the subject.Kind and the namespace, as SA's can have the same name across namespaces.

cnf-certification-test/accesscontrol/suite.go
```diff
@@ -410,44 +410,58 @@ func testPodRoleBindings(env *provider.TestEnvironment) {
...
+				// Determine if that service account is not in the same namespace as the pod.
+				if subject.Kind != "ServiceAccount" || subject.Name != put.Spec.ServiceAccountName {
+					continue
+				}
 
-		if len(roleBindings) > 0 {
-			logrus.Warnf("Pod: %s/%s has the following role bindings: %s", put.Namespace, put.Name, roleBindings)
-			tnf.ClaimFilePrintf("Pod: %s/%s has the following role bindings: %s", put.Namespace, put.Name, roleBindings)
-			failedPods = append(failedPods, put.Name)
+				if rbac.IsRoleBindingOutOfNamespace(put.Namespace, put.Spec.ServiceAccountName, env.RoleBindings[rbIndex].Namespace, env.RoleBindings[rbIndex].Name) {
```
greyerof, line 425: Are these helper functions really needed? My suggestion:
```
		RB_LOOP:
		for rbIndex := range env.RoleBindings {
			if env.RoleBindings[rbIndex].Namespace == put.Namespace {
				continue
			}

			for _, subject := range env.RoleBindings[rbIndex].Subjects {
				if subject.Kind == rbacv1.ServiceAccountKind && subject.Namespace == put.Namespace && subject.Name == put.Spec.ServiceAccountName {
					logrus.Warnf("Pod: %s/%s has the following role bindings that do not live in the same namespace: %s", put.Namespace, put.Name, env.RoleBindings[rbIndex].Name)
					tnf.ClaimFilePrintf("Pod: %s/%s has the following role bindings that do not live in the same namespace: %s", put.Namespace, put.Name, env.RoleBindings)
					failedPods = append(failedPods, put.String())
					break RB_LOOP
				}
			}
		}
```
Also note that in that helper func `roleBindingOutOfNamespace()` you're using the RoleBinding name to compare it against the ServiceAccount name, but I believe that's wrong: you don't that RoleBinding name here. The ServiceAccount name should be only compared with the subject.Name.

cnf-certification-test/accesscontrol/suite.go
```diff
@@ -410,44 +410,58 @@ func testPodRoleBindings(env *provider.TestEnvironment) {
...
+					break
+				}
+			}
+			// Break of out the loop if we found a role binding that is out of namespace
+			if found {
+				break
+			}
 		}
 	}
 	testhelper.AddTestResultLog("Non-compliant", failedPods, tnf.ClaimFilePrintf, ginkgo.Fail)
```
edcdavid, line 439: use the new reason object, see testContainerSCC function.
https://github.com/test-network-function/cnf-certification-test/pull/834

cnf-certification-test/accesscontrol/suite.go
sebrandon1: Okay this makes sense.  Agreed the old carryover code from `roleBindingOutOfNamespace` is not correct I will remove it and use the suggestion above.

cnf-certification-test/accesscontrol/suite.go
sebrandon1: Done.

cnf-certification-test/accesscontrol/suite.go
sebrandon1: Fixed!

pkg/provider/pods.go
sebrandon1: This should be addressed in my latest commit.

cnf-certification-test/accesscontrol/suite.go
sebrandon1: @edcdavid Can you re-review to see if I'm using it the way it is intended?  Thanks!

cnf-certification-test/accesscontrol/suite.go
```diff
@@ -401,53 +402,81 @@ func testPodServiceAccount(env *provider.TestEnvironment) {
...
+	var compliantObjects []*testhelper.ReportObject
+	var nonCompliantObjects []*testhelper.ReportObject
 
 	for _, put := range env.Pods {
 		ginkgo.By(fmt.Sprintf("Testing role binding for pod: %s namespace: %s", put.Name, put.Namespace))
-		if put.Spec.ServiceAccountName == "" {
-			logrus.Infof("%s has an empty serviceAccountName, skipping.", put.String())
+		if put.IsUsingDefaultServiceAccount() {
+			logrus.Infof("%s has an empty or default serviceAccountName, skipping.", put.String())
 			continue
```
greyerof, line 412: I don't think you should filter out pods that use the "default" ServiceAccount. It's ok for pods to use it and create rolebinding/clusterrolebindings that have it as the subject for the roles/clusterroles in order to grant access to the cluster api. Good example here https://stackoverflow.com/questions/52995962/kubernetes-namespace-default-service-account

Maybe my previous comment was a bit misleading. I just wanted to say that comparing with the empty string "" seemed redundant to me, as all the pods will come with, at least, the "default" serviceAccountName set. But the "default" is actually a valid SA that can be used by role/clusterrole-bindings.

ramperher (commented):
With what regards to DCI, the change is working fine now

ramperher (approved):

cnf-certification-test/accesscontrol/suite.go
sebrandon1: Ah gotcha.  Okay yes I will remove that check.

cnf-certification-test/accesscontrol/suite.go
```diff
@@ -401,53 +402,77 @@ func testPodServiceAccount(env *provider.TestEnvironment) {
...
+			for _, subject := range env.RoleBindings[rbIndex].Subjects {
+				// If the subject is a service account and the service account is in the same namespace as the pod, then we have a failure
+				//nolint:gocritic
+				if subject.Kind == rbacv1.ServiceAccountKind && subject.Namespace == put.Namespace && subject.Name == put.Spec.ServiceAccountName {
+					failMsg := fmt.Sprintf("Pod: %s/%s has the following role bindings that do not live in the same namespace: %s", put.Namespace, put.Name, env.RoleBindings[rbIndex].Name)
+					logrus.Warnf(failMsg)
+					tnf.ClaimFilePrintf(failMsg)
+
+					// Add the pod to the non-compliant list
+					nonCompliantObjects = append(nonCompliantObjects, testhelper.NewPodReportObject(put.Namespace, put.Name, failMsg, false))
```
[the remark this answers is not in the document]
sebrandon1: Done.

cnf-certification-test/accesscontrol/suite.go
```diff
@@ -398,60 +399,102 @@ func testPodServiceAccount(env *provider.TestEnvironment) {
...
+			}
+			// Break of out the loop if we found a role binding that is out of namespace
+			if found {
+				break
+			}
+		}
+		// Add pod to the compliant object list
+		if podIsCompliant {
+			compliantObjects = append(compliantObjects,
+				testhelper.NewPodReportObject(put.Namespace, put.Name, "Compliant because the role bindings used by this pod (applied by the service account) live in the same namespace", true))
```
edcdavid, line 457: minor/optional:
Compliant because the role bindings used by this pod (applied by the service account) live in the same namespace -> Compliant because **all** the role bindings used by this pod (applied by the service account**s**) live in the same namespace

edcdavid (approved):

PR #1051, opened 2023-04-26
Added qe-coverage-report subcommand to tnf tool.

greyerof:
Parses the (bool) qe flag from each catalog test case identifier and computes the total and per test suite coverage percentages.


bnshr (approved):
lgtm!

sebrandon1 (approved):

PR #1046, opened 2023-04-24
Accessing kube-api via projected volume bypasses automountToken test

shimritproj:
Testing pods to ensure they are not using projected volumes for service account access.

PR #977, opened 2023-03-30
Add all markdown files to the linter

rdavid:

PR #1062, opened 2023-05-02
Remove unnecessary linter

bnshr:


rdavid:
What is a purpose of removing the Markdown linter? [This PR](https://github.com/test-network-function/cnf-certification-test/pull/1046) shows the real problem with the Markdown generation, it generates `CATALOG.md` with a wrong syntax. Now it looks like the problem is not related to the content of the PR. I agree with you that Shimrit's PR is not related to this linter error. But there is still an issue to be resolved. Meanwhile, Shimrit did a workaround of the Markdown generation by my advice. I will address the Markdown generation issue in a separate PR.

bnshr:
The reasons are already stated in the google chat. The documentation pages do not show CATALOG.md reference and highlights and tips. The screenshots are shared already.

sebrandon1:
I would argue that instead of removing it, we need a make path to be able to run this linter locally like we do the other linters. Just tie it into `make lint` and let it run like any of the others.

rdavid:
@bnshr Please, take a look at the web-site to control, everything is the same as before my Markdown linter's commit. If the site is OK, close the PR. Thank you.

sebrandon1:
Closing as the doc website is now fixed.  The markdown syntax exceptions are added in https://github.com/test-network-function/cnf-certification-test/pull/977

PR #1082, opened 2023-05-15
Manually update machine-config-operator

sebrandon1:
https://github.com/openshift/machine-config-operator/commit/49f32d46538e4387966a22aba703f5e579a37356

The MCO was causing some sort of `git_dependencies_not_reachable` error in the dependabot logs so I'm manually updating it to see if it clears the error.


sebrandon1:
https://docs.github.com/en/code-security/dependabot/working-with-dependabot/troubleshooting-dependabot-errors

PR #1083, opened 2023-05-15
Only build during smoke tests, do not run unit tests

sebrandon1:
Speed up the CI by not running the unit tests as part of `make build`.  Those are already ran in other `jobs` in the CI already and do not need to be repeated.

`make build` currently runs the unit tests first, then builds.

### DIFF
--- a/.github/workflows/self-hosted.yml
+++ b/.github/workflows/self-hosted.yml
@@ -51,8 +51,8 @@ jobs:
       - name: Install ginkgo
         run: make install-tools
 
-      - name: Execute `make build`
-        run: make build
+      - name: Build CNF test suite binary
+        run: make build-cnf-tests
 
       # Create a Kind cluster for testing.
       - name: Check out `cnf-certification-test-partner`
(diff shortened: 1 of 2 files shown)


.github/workflows/self-hosted.yml
```diff
@@ -51,8 +51,8 @@ jobs:
       - name: Install ginkgo
         run: make install-tools
 
-      - name: Execute `make build`
-        run: make build
+      - name: Execute `make build-cnf-tests`
```
rdavid, line 54: Give a better name.
```suggestion
      - name: Build CNF tests
```

rdavid (approved):

.github/workflows/self-hosted.yml
sebrandon1: I kind of like the explicit naming where you know exactly what's going on in the log.

rdavid: Rule number one:
 https://stackoverflow.blog/2021/12/23/best-practices-for-writing-code-comments/
 
Bad example:
```
// Increase the counter.
counter++;
```

sebrandon1: Okay I'll change.  😄

PR #1087, opened 2023-05-15
Add CRC local testing

sebrandon1:
There's a new 2.18 release of CRC.  I wanted to see if this works any better.  I will close if there is no change.

https://github.com/crc-org/crc/releases/tag/v2.18.0


sebrandon1:
Closing until we get bigger runners.

PR #1094, opened 2023-05-16
Runasuser tests

sebrandon1:
* Modify the existing RunAsUser 1337 test and move the logic within the `Pod` provider struct.
* Utilize the new compliant/non-compliant objects.


pkg/provider/pods.go
```diff
@@ -400,3 +400,15 @@ func (p *Pod) IsUsingClusterRoleBinding(clusterRoleBindings []rbacv1.ClusterRole
...
+	if p.Pod.Spec.SecurityContext == nil {
+		return false
+	}
+
+	if p.Pod.Spec.SecurityContext.RunAsUser != nil && *p.Pod.Spec.SecurityContext.RunAsUser == 1337 {
+		return true
+	}
+
+	return false
+}
```
rdavid, line 414:
```suggestion
func (p *Pod) IsRunAsUser1337() bool {
	if p.Pod.Spec.SecurityContext == nil || p.Pod.Spec.SecurityContext.RunAsUser == nil {
		return false
	}

	return *p.Pod.Spec.SecurityContext.RunAsUser == 1337
}
```

rdavid (approved):

cnf-certification-test/accesscontrol/suite.go
```diff
@@ -717,18 +717,20 @@ func testPodRequestsAndLimits(env *provider.TestEnvironment) {
...
 	// Note this test is only ran as part of the 'extended' test suite.
+	var compliantObjects []*testhelper.ReportObject
+	var nonCompliantObjects []*testhelper.ReportObject
 	ginkgo.By("Testing pods to ensure none are using UID 1337")
-	const leetNum = 1337
-	var badPods []string
 	for _, put := range env.Pods {
 		ginkgo.By(fmt.Sprintf("checking if pod %s has a securityContext RunAsUser 1337 (ns= %s)", put.Name, put.Namespace))
-		if put.Spec.SecurityContext.RunAsUser != nil && *put.Spec.SecurityContext.RunAsUser == int64(leetNum) {
+		if put.IsRunAsUser1337() {
```
greyerof, line 725: It seems a bit weird to me adding a pod's method for that custom number when you can just make it generic with:
```
func (p *Pod) IsRunAsUserID(uid int) bool
```
And then:
```
const serviceMeshUID = 1337
...
put.IsRunAsUserID(serviceMeshUID)
```
This way we could also use it in `testSecConRootUser()` to check [root's uid (0)](https://github.com/test-network-function/cnf-certification-test/blob/d365e02bb1a9de2293fa239362222432b8784ea9/cnf-certification-test/accesscontrol/suite.go#L294).

pkg/provider/pods.go
sebrandon1: Done!

cnf-certification-test/accesscontrol/suite.go
sebrandon1: Done.

greyerof (approved):

PR #1096, opened 2023-05-17
Bump python version

rdavid:


greyerof (approved):

sebrandon1 (approved):
This is a good temporary measure but this either needs to be automated or we need to remove the package versions until a solution is found.

rdavid:
This file is not checked by dependabot, probably because it is in a dot folder.

rdavid:
In any case dependabot updates only docker image versions and not package versions.

PR #1098, opened 2023-05-17
Add a link to Dockerfile for dependabot monitoring

rdavid:
The link allows `Dependabot` to monitor the `Dockefile` inside a dot directory. There is an unfortunate downside. The PR created by `Dependabot` will replace the link with the actual file with a new version. We should not merge this PR (produced by `Dependabot`), but bump versions in the `Dockefile` manually.

UPDATE: The better solution is to move the `Dockerfile` to root with another name `Dockerfile-documentation` and create `Dockerfile` link to it.

### DIFF
--- a/.github/actions/documentation/Dockerfile
+++ b/.github/actions/documentation/Dockerfile
@@ -1,12 +1,1 @@
-FROM registry.access.redhat.com/ubi8/ubi:8.7
-
-RUN dnf install -y python39-3.9.16 && dnf clean all
-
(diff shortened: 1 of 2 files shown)


sebrandon1:
Is this really the best way to do this?  I feel like others have had to encounter this sort of thing before.

sebrandon1 (approved):
I think we need to put the `hadolint` ignore in our other dockerfiles as well.

PR #1110, opened 2023-05-18
Add CRC local testing

sebrandon1:
2.19 has been released, testing.


sebrandon1:
Same problem with 2.19, the smaller Mac runners can't handle it I don't think.

PR #1126, opened 2023-05-23
Block host network access pods

bnshr:


cnf-certification-test/performance/suite.go
```diff
@@ -47,33 +47,41 @@ var _ = ginkgo.Describe(common.PerformanceTestKey, func() {
...
+	ginkgo.It(testID, ginkgo.Label(tags...), func() {
+		var guaranteedPodContainersWithExclusiveCPUs = env.GetGuaranteedPodContainersWithExlusiveCPUs()
+		testhelper.SkipIfEmptyAll(ginkgo.Skip, guaranteedPodContainersWithExclusiveCPUs)
+		testRtAppsNoExecProbes(&env, guaranteedPodContainersWithExclusiveCPUs)
+	})
+
+	// Scheduling related tests begins here
 	testID, tags = identifiers.GetGinkgoTestIDAndLabels(identifiers.TestSharedCPUPoolSchedulingPolicy)
 	ginkgo.It(testID, ginkgo.Label(tags...), func() {
+		env.Pods = env.GetPodsWithoutHostPID() // Consider only pods with no access to host machine
```
sebrandon1, line 60: We don't want to overwrite the `env.Pods` as this will affect other tests in the suite.

bnshr: The tests which might have got impacted are executed before scheduling tests.

edcdavid: I also think we should not do this

PR #1131, opened 2023-05-24
Improve the SCC package

sebrandon1:
Notable changes:
- Removed `subslice` and `contains` function and created a `stringhelper.SubSlice` helper function with corresponding unit tests.
- Re-ordered the subslice check to take the `drop` slice first followed by the `required` slice.  Because you want to check your SCCs against the required SCCs.
- Adjust camel case of `PodListcategory` to `PodListCategory`.
- Added a comment about the `NOK` value on the RunAsNonRoot in category 1.

### DIFF
--- a/pkg/stringhelper/stringhelper.go
+++ b/pkg/stringhelper/stringhelper.go
@@ -36,21 +36,14 @@ func StringInSlice[T ~string](s []T, str T, contains bool) bool {
 	return false
 }
 
-// SubSlice checks if a slice exists within a slice
-func SubSlice[T ~string](s, sub []T) bool {
-	for i := range s {
-		if len(s)-i < len(sub) {
+// SubSlice checks if a slice's elements all exist within a slice
+func SubSlice(s, sub []string) bool {
+	for _, v := range sub {
+		if !StringInSlice(s, v, false) {
 			return false
 		}
-		for j := range sub {
-			if s[i+j] != sub[j] {
-				break
-			} else if j == len(sub)-1 {
-				return true
-			}
-		}
 	}
-	return false
+	return true
 }
 
 func RemoveEmptyStrings(s []string) []string {
--- a/pkg/stringhelper/stringhelper.go
+++ b/pkg/stringhelper/stringhelper.go
@@ -36,6 +36,23 @@ func StringInSlice[T ~string](s []T, str T, contains bool) bool {
 	return false
 }
 
+// SubSlice checks if a slice exists within a slice
+func SubSlice[T ~string](s, sub []T) bool {
+	for i := range s {
+		if len(s)-i < len(sub) {
+			return false
+		}
+		for j := range sub {
+			if s[i+j] != sub[j] {
+				break
+			} else if j == len(sub)-1 {
+				return true
+			}
+		}
+	}
+	return false
+}
+
 func RemoveEmptyStrings(s []string) []string {
 	var r []string
 	for _, str := range s {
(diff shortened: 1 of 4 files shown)


pkg/stringhelper/stringhelper.go
```diff
@@ -36,6 +36,23 @@ func StringInSlice[T ~string](s []T, str T, contains bool) bool {
 	return false
 }
 
+// SubSlice checks if a slice exists within a slice
+func SubSlice[T ~string](s, sub []T) bool {
+	for i := range s {
```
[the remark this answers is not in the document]
sebrandon1: `range` handles zero length slices just fine.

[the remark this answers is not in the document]
sebrandon1: Had to take a second look.  Yeah it'll just skip right over the loops if there's a zero length `sub` slice.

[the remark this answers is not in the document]
rdavid: It skips, but returns false. I assumed it needs true.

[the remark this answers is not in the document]
sebrandon1: Let me play with it.  👍

pkg/stringhelper/stringhelper.go
```diff
@@ -36,6 +36,23 @@ func StringInSlice[T ~string](s []T, str T, contains bool) bool {
 	return false
 }
 
+// SubSlice checks if a slice exists within a slice
+func SubSlice[T ~string](s, sub []T) bool {
```
greyerof, line 40: I don't think this new implementation is doing the same as the old one, as it will only return true if the elements of the "sub" slice appear consecutive (and in the same order) in the slice "s". But this is not what the SCC test case is actually trying to check. See this UT:
```
		{ 
			testSliceA:     []string{"MKNOD", "NET_RAW", "SETUID", "SETGID", "KILL"},
			testSliceB:     []string{"MKNOD", "SETUID", "SETGID", "KILL"},
			expectedOutput: true,
		},
```
That unit test fails with your new version (SubSlice() returns false), but it should return true. All the elements in testSliceB appear in testSliceA, which is what we are actually checking in this SCC tc (if I'mt not wrong).
This other unit test should also make the SubsSlice() to return true:
```
		{
			testSliceA:     []string{"KILL", "SETUID", "MKNOD", "NET_RAW", "SETGID"},
			testSliceB:     []string{"MKNOD", "SETUID", "SETGID", "KILL"},
			expectedOutput: true,
		},
```

sebrandon1: Okay I re-did the SubSlice function to look at the individual elements and added another unit test case.

pkg/stringhelper/stringhelper.go
```diff
@@ -36,6 +36,16 @@ func StringInSlice[T ~string](s []T, str T, contains bool) bool {
 	return false
 }
 
+// SubSlice checks if a slice's elements all exist within a slice
+func SubSlice[T ~string](s, sub []T) bool {
```
rdavid, line 40: Is it a good taste to use generics? It is a string helper.

sebrandon1: Works either way.  I can switch it just accept `[]string` variables if needed.

rdavid: I think it should be string type in this case.

sebrandon1: Generics removed.

rdavid (approved):

greyerof (approved):

PR #1136, opened 2023-05-25
Prune the self-hosted runner prior to building cluster

sebrandon1:
This only applies to the self-hosted runner setup.

### DIFF
--- a/.github/workflows/self-hosted.yml
+++ b/.github/workflows/self-hosted.yml
@@ -65,6 +65,20 @@ jobs:
         run: make bootstrap-cluster
         working-directory: cnf-certification-test-partner
 
+      - name: Preemptively delete the Kind cluster
+        uses: nick-fields/retry@v2
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/self-hosted.yml
```diff
@@ -65,6 +65,20 @@ jobs:
...
         working-directory: cnf-certification-test-partner
 
+      - name: Prune the local docker system
+        uses: nick-fields/retry@v2
+        with:
+          timeout_minutes: 10
+          max_attempts: 3
+          command: docker system prune -f
+
+      - name: Preemptively delete the Kind cluster
```
greyerof, line 75: I would delete the kind cluster first. Then, docker system prune to remove everything that is left.

sebrandon1: Good call, I'll do that.

rdavid (approved):

PR #1145, opened 2023-05-30
adding failure reason to icmp connectivity tests

edcdavid:
This PR adds the following: 
- failure reason for connectivity tests
- better rendering of failure reason fields. Now the fields are ordered by the addition order. A list is used instead of a map to store the order. Fixed html to better scale
- added fast rendering of configuration items. Added a progress bar when loading large set of objects
Note: the format of the failure reason is changed so new claim files would not be able to be parsed with older pages

### DIFF
--- a/cnf-certification-test/accesscontrol/pidshelper_test.go
+++ b/cnf-certification-test/accesscontrol/pidshelper_test.go
@@ -22,6 +22,7 @@ import (
 
 	"github.com/stretchr/testify/assert"
 	"github.com/test-network-function/cnf-certification-test/internal/clientsholder"
+	"github.com/test-network-function/cnf-certification-test/pkg/testhelper"
 )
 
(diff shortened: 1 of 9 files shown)


ramperher:
@edcdavid , just to warn you, there are a lot of tnf tests that are now skipped with this code: https://www.distributed-ci.io/jobs/bbdfb302-1c0b-4620-b117-75e9acc83fbe/jobStates, don't know if something has changed to cause that

ramperher:
> @edcdavid , just to warn you, there are a lot of tnf tests that are now skipped with this code: https://www.distributed-ci.io/jobs/bbdfb302-1c0b-4620-b117-75e9acc83fbe/jobStates, don't know if something has changed to cause that

Nevermind, this does not come from this change, but it comes from https://github.com/test-network-function/cnf-certification-test/pull/1140

sebrandon1 (approved):

greyerof (approved):

PR #1173, opened 2023-06-14
QE Coverage : Add more details in result

bnshr:
Screenshot 2023-06-14 at 12 39 03

### DIFF
--- a/cmd/tnf/generate/qe_coverage/qe_coverage.go
+++ b/cmd/tnf/generate/qe_coverage/qe_coverage.go
@@ -86,10 +86,10 @@ func runGenerateQeCoverageReport(_ *cobra.Command, _ []string) error {
 	fmt.Printf("Total QE Coverage: %.f%%\n\n", qeCoverage.TotalCoveragePercentage)
 
 	// Per test suite QE coverage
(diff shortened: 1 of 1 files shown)


greyerof (approved):

cmd/tnf/generate/qe_coverage/qe_coverage.go
```diff
@@ -86,10 +86,10 @@ func runGenerateQeCoverageReport(_ *cobra.Command, _ []string) error {
 	fmt.Printf("Total QE Coverage: %.f%%\n\n", qeCoverage.TotalCoveragePercentage)
 
 	// Per test suite QE coverage
-	fmt.Printf("%-30s\t%-10s\t%s\n", "Test Suite", "Test Cases", "QE Coverage")
+	fmt.Printf("%-30s\t%-20s\t%-20s\t%s\n", "Test Suite Name", "QE Coverage", "Total Test Cases", "Not Covered Test Count")
```
edcdavid, line 89: can we also add the list of test cases that are not covered yet? This would be helpful to know which ones need to be worked on

bnshr: Good idea. We can have that feature which can be done by extending the tool with a **subcommand** which would need test suite name. I would create a JIRA ticket for that.

edcdavid (approved):

PR #1176, opened 2023-06-15
Add a single label for running ALL tests instead of having to choose individual labels

aabughosh:


run-cnf-suites.sh
```diff
@@ -79,6 +80,9 @@ done
 # Strips the leading whitespace.
 LABEL="$(echo -e "${LABEL}" | sed -e 's/^[[:space:]]*//')"
 
+if [[ $LABEL == "all" ]]; then
+	LABEL='common,extended,faredge,telco'
+fi
```
rdavid, line 85:
```suggestion
[ "$LABEL" = all ] && LABEL=common,extended,faredge,telco
```

run-cnf-suites.sh
```diff
@@ -79,6 +80,9 @@ done
...
 # List the specs (filtering by suite).
```
rdavid, line 86:
```suggestion
[ "$LABEL" = all ] && LABEL=common,extended,faredge,telco

# List the specs (filtering by suite).
```

rdavid (approved):

sebrandon1 (approved):

sebrandon1:
Agreed we should change this for both files.

cnf-certification-test/identifiers/identifiers.go
```diff
@@ -560,7 +560,7 @@ func InitCatalog() map[claim.Identifier]claim.TestCaseDescription {
 		SecConNonRootUserRemediation,
 		SecConNonRootUserExceptionProcess,
 		bestPracticeDocV1dot3URL+" Section 5.2",
-		false,
+		true,
```
greyerof, line 563: Is this change intentional? This is not related with the PR title at all.

sebrandon1 (approved):

run-cnf-suites.sh
sebrandon1: I think the way it is has a good amount of readability.  We can keep it this way.

PR #1186, opened 2023-06-15
Adding new test case for CRD provided roles

edcdavid:
build-depends: 28703
Adds a test verifying the following:
- In the namespaces under test, look for roles with at least one CRDs under test present
- for each of these roles, make sure that no other resource is present that do not belong to the CRDs under test
- if only CRDs under test are defined in the role, add the role to compliant roles
- otherwise add the role to non compliant role
- if no role contain CRDs under test, fail the test

Failure Reason output:
Failure reason unrolls every compliant rules in compliant and partially compliant roles.
Failure reason unrolls every non-compliant rule in non-compliant roles
Failure reason list compliant and non compliant roles


pkg/testhelper/testhelper.go
```diff
@@ -180,6 +189,14 @@ func (obj *ReportObject) AddField(aKey, aValue string) (out *ReportObject) {
 	return obj
 }
 
+func NewNamespacedReportObject(aReason, aType string, isCompliant bool, aNamespace string) (out *ReportObject) {
```
sebrandon1, line 192: This got me thinking, what if we just baked the `ReportObject` into the objects themselves.  Such as like Pod.NewReportObject()?

sebrandon1 (commented):
Just some initial comments.

cnf-certification-test/accesscontrol/suite.go
```diff
@@ -791,3 +797,47 @@ func testNodePort(env *provider.TestEnvironment) {
...
+	var nonCompliantObjects []*testhelper.ReportObject
+	crdResources := rbac.GetCrdResources(env.Crds)
+	for roleIndex := range env.Roles {
+		if !provider.IsNamespaceUnderTest(env.Roles[roleIndex].Namespace, env.Namespaces) {
+			continue
+		}
+
+		allRules := rbac.GetAllRules(&env.Roles[roleIndex])
+
+		matchingRules, nonMatchinRules := rbac.FilterRulesNonMatchingResources(allRules, crdResources)
```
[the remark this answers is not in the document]
edcdavid: fixed

cnf-certification-test/accesscontrol/rbac/roles.go
```diff
@@ -0,0 +1,110 @@
...
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License along
+// with this program; if not, write to the Free Software Foundation, Inc.,
+// 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+package rbac
```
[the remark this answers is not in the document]
edcdavid: fixed. Didn't get to it yet :-)

pkg/testhelper/testhelper.go
edcdavid: not sure if it could work, the report object is really custom based on the specific processing of the test, not just the content of the base structures: we might need more than one per object or even one mixing base objects. I fell if we add it there would be harder to maintain. In my opinion, the failure reason should feel like writing text annotating the actual tests we are running.

cnf-certification-test/accesscontrol/rbac/roles.go
```diff
@@ -0,0 +1,110 @@
...
+		}
+	}
+	return ruleList
+}
+
+// Checks the resource name in the role against plural, singular and short names
+func equalResource(aResource CrdResource, aRule RoleRule) bool {
+	// remove subresources to keep only resource name
+	ruleBaseResourceName := strings.Split(aRule.Resource.Name, "/")[0]
+	return aResource.Group == aRule.Resource.Group && (aResource.SingularName == ruleBaseResourceName ||
```
greyerof, line 73: I think most of these comparisons are not needed. I don't think you can have different CRDs with the same pluralName and ApiGroup but with different SingularName. Same with the short names. If I'm not wrong, this implementation should be good enough:
```
func isResourceInRoleRule(crd CrdResource, roleRule RoleRule) bool {
	// remove subresources to keep only resource (plural) name
	ruleResourcePluralName := strings.Split(roleRule.Resource.Name, "/")[0]

	return crd.Group == roleRule.Resource.Group && crd.PluralName == ruleResourcePluralName
}
```

cnf-certification-test/accesscontrol/rbac/roles.go
```diff
@@ -0,0 +1,110 @@
...
+		aResource.PluralName == ruleBaseResourceName ||
+		stringhelper.StringInSlice(aResource.ShortNames, ruleBaseResourceName, false))
+}
+
+func FilterRulesNonMatchingResources(ruleList []RoleRule, resourceList []CrdResource) (matching, nonMatching []RoleRule) {
+	for _, aRule := range ruleList {
+		for _, aResource := range resourceList {
+			if equalResource(aResource, aRule) {
+				matching = append(matching, aRule)
```
greyerof, line 82: Here, you've confirmed that this role's rule has the crd resource as "target", but shouldn't you also check that its parent role has only one rule? Otherwise, the role could have more rules for other resources, which is what we don't want, as per the second part of the TC description "If an application creates CRDs it must supply a role to access those CRDs **and no other API resources/permission**.".

edcdavid: This part only matches the valid rules in the role, e.g. the ones that apply to CRDs under test.  The rules applying to non-CRDs under test in the same role are picked up by:
nonMatching = SliceDifference(ruleList, matching)
in my understanding "If an application creates CRDs it must supply a role to access those CRDs and no other API resources/permission." means "If an application creates CRDs it must supply a role to access those CRDs only and no API resources/permission other than what has been created by the application." It does not limit the number of rules in a role or the number of roles

cnf-certification-test/accesscontrol/rbac/roles.go
edcdavid: This code is comparing the following between a CRD and a rule (inside a role):
- rule's resource group with CRD group
- rule's resource name with either CRD plural (more usual) or Singular or short (any in the list ) names 
For instance, if comparing only the plural form and the role's rule is using the short name this function will not match but it should.

greyerof: Ok, I just thought that as per k8s rbac docs [referring to resources](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#referring-to-resources) section, every item in the resources list in role's rules should use the plural form always. It can also be seen doing:
```
oc get roles.rbac.authorization.k8s.io -A -o json | jq -r '.items[] | .rules[] | .resources[]
```

cnf-certification-test/accesscontrol/rbac/roles.go
greyerof: You're right, your code looks good.

greyerof (approved):

cnf-certification-test/accesscontrol/rbac/roles.go
edcdavid: I think you are right, just did a test with a new user and creating a single restrictive rule to allow pod listing. When using the singular or the short resource names, the user was not able to list pods. only the plural resource worked. Looks like singular and short are used only by kubectl. I will use you suggestion

PR #1187, opened 2023-06-16
Fix for hardcoded registry registry.redhat.io

edcdavid:
build-depends: 28703
The registry.redhat.io is hardcoded internally in Pyxis to registry.access.redhat.com. Rewrites registry when processing requests.
Note: Awaiting clarifications from Pyxis team on the on how to use the docker_image_id and image_id fields to consistently get the image digest value

### DIFF
--- a/internal/certdb/config/config.go
+++ b/internal/certdb/config/config.go
@@ -1,4 +1,4 @@
-// Copyright (C) 2020-2022 Red Hat, Inc.
+// Copyright (C) 2020-2023 Red Hat, Inc.
 //
 // This program is free software; you can redistribute it and/or modify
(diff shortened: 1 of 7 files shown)


internal/certdb/config/config.go
```diff
@@ -0,0 +1,19 @@
+// Copyright (C) 2020-2022 Red Hat, Inc.
```
sebrandon1, line 1:
```suggestion
// Copyright (C) 2020-2023 Red Hat, Inc.
```

edcdavid: fixed

greyerof (approved):

PR #1200, opened 2023-06-21
Generate tar.gz archive file with results & web artifacts.

greyerof:
build-depends: 28703

After running all the test cases, a compressed file will be created with all the results files and web artifacts to review them. The file has a UTC date-time prefix and looks like this:

20230620-110654-cnf-test-results.tar.gz

The "20230620-110654" sample prefix means "June-20th 2023, 11:06:54"

This is the content of the tar.gz file:
- claim.json
- cnf-certification-tests_junit.xml
- claimjson.js
- classification.js
- results-embed.html
- results.html

Now, by default, only the claim.js, the junit xml file and this new tar.gz file are created after the test suite has finished, as this is probably all that normal partners/users will need.

This change includes two extra (debug/developer mode) env vars to control the web artifacts and the the new tar.gz file generation:

TNF_OMIT_ARTIFACTS_ZIP_FILE=true/false
  Defaulted to false in the launch scripts. If set to true, the tar.gz
  generation will be skipped.

TNF_INCLUDE_WEB_FILES_IN_OUTPUT_FOLDER=true/false
  Defaulted to false in the launch scripts. If set to true, the web
  viewer/parser files will also be copied to the output (claim) folder.

Notes:
The files classification.js, results-embed.html and results.html have been moved into the cnf-certification-test/html (new) folder so they can be used by the golang's embed package.

As the claimjson.js is just the same content as the claim.json file prefixed by a "var initjson = " text, I removed the bash script that creates it and made it in go.

ToDos in further PRs:
- The classification.js file can (easily) also be built dynamically in go after the test suite has finished and added to the gz file.
- Include the execution log (tnf-execution.log) in the tar.gz file.
- UTs


Dockerfile
```diff
@@ -54,16 +54,17 @@ RUN \
 # Copy all of the files into the source directory and then switch contexts
 COPY . ${TNF_SRC_DIR}
 WORKDIR ${TNF_SRC_DIR}
-RUN make install-tools && make build-cnf-tests
+RUN \
+	make install-tools \
+	&& make classification-js \
+	&& make build-cnf-tests
```
rdavid, line 60:
```suggestion
RUN make build-cnf-tests classification-js install-tools
```

Make works with multiple arguments, put the argument in alphabetical order.

cnf-certification-test/results/archiver.go
```diff
@@ -0,0 +1,88 @@
...
+
+	"github.com/sirupsen/logrus"
+)
+
+const (
+	tarGzFileNameSuffix = "cnf-test-results.tar.gz"
+)
+
+func generateZipFileName() string {
+	dateTimePrefixFormat := "%02d%02d%02d-%02d%02d%02d"
```
rdavid, line 20: Move it to const section.

cnf-certification-test/results/archiver.go
```diff
@@ -0,0 +1,88 @@
...
+
+	header, err := tar.FileInfoHeader(info, info.Name())
+	if err != nil {
+		return nil, fmt.Errorf("failed to get file info header for %s: %v", file, err)
+	}
+
+	return header, nil
+}
+
+// Creates a zip file in the outputDir containing each file in the filePaths slice.
```
edcdavid, line 44: zip -> compressed

cnf-certification-test/results/html/results.html
edcdavid, line 1: rebase to get latest html files

edcdavid (approved):

Dockerfile
greyerof: Fixed in https://github.com/test-network-function/cnf-certification-test/pull/1200/commits/b173b199ea7787ccd16153c2d3e151f690156035
install-tools should run before build-cnf-tests

cnf-certification-test/results/archiver.go
greyerof: Done. I followed the layout format suggestion from @edcdavid in https://github.com/test-network-function/cnf-certification-test/pull/1200/commits/93040781ec5db491aee76d5bd8ad94f9f144211a

cnf-certification-test/results/archiver.go
```diff
@@ -0,0 +1,88 @@
...
+const (
+	tarGzFileNameSuffix = "cnf-test-results.tar.gz"
+)
+
+func generateZipFileName() string {
+	dateTimePrefixFormat := "%02d%02d%02d-%02d%02d%02d"
+
+	now := time.Now().UTC()
+
+	return fmt.Sprintf(dateTimePrefixFormat+"-"+tarGzFileNameSuffix,
```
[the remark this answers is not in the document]
greyerof: Thx! Done in https://github.com/test-network-function/cnf-certification-test/pull/1200/commits/93040781ec5db491aee76d5bd8ad94f9f144211a

cnf-certification-test/results/html/results.html
greyerof: Done.

ramperher:
@greyerof I've added some code in dci-openshift-app-agent to include your new tar.gz, that's why DCI job was failing, because we were handling the Javascript files in a specific way but now it's changed. Code is here: https://softwarefactory-project.io/r/c/dci-openshift-app-agent/+/28703 (it is included as build-depends in the description of the PR). With this, the DCI job should pass.

rdavid (approved):

sebrandon1 (approved):

PR #1201, opened 2023-06-21
Adding support for skip reason and HTML rendering

edcdavid:
build-depends: 28703
This is adding an indication for the reason for skipping a test in the Failure Reason field of the claim file. Also added display of aborted test cases.
HTML page updated to display aborted test cases and skip reason

### DIFF
--- a/cnf-certification-test/results/html/results-embed.html
+++ b/cnf-certification-test/results/html/results-embed.html
@@ -19,7 +19,12 @@
   <script src="https://cdn.jsdelivr.net/npm/bootstrap@5.3.0/dist/js/bootstrap.bundle.min.js"></script>
   <script src="https://unpkg.com/ansi_up@5.1.0/ansi_up.js" crossorigin="anonymous"></script>
   <style type="text/css">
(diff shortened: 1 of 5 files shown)


script/results.html
```diff
@@ -613,10 +638,18 @@
             for (const item of input[key]) {
               const duration = dayjs.duration(item.duration / 1000000)
               formattedDuration = duration.format('D[d] H[h] m[m] s[s] SSS[ms]')
+              skippedReason = ""
+              if (item.state == "skipped") {
+                skippedReason = item.failureReason
+                if (skippedReason == "") {
+                  skippedReason = "Test case skipped by user"
```
sebrandon1, line 645:
```suggestion
                  skippedReason = "Test case skipped by configuration"
```

Minor: Maybe change this to "configuration" instead of placing blame on the user themselves.

sebrandon1 (approved):

rdavid (approved):

script/results.html
edcdavid: fixed

PR #1205, opened 2023-06-23
Add Failure Reason for undeclared port usage

edcdavid:
build-depends: 28703
Added Pod level and Port level reports

### DIFF
--- a/cnf-certification-test/networking/suite.go
+++ b/cnf-certification-test/networking/suite.go
@@ -18,6 +18,7 @@ package networking
 
 import (
 	"fmt"
+	"strconv"
 
 	"github.com/onsi/ginkgo/v2"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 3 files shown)


cnf-certification-test/networking/suite.go
```diff
@@ -188,20 +191,38 @@ func testUndeclaredContainerPortsUsage(env *provider.TestEnvironment) {
...
+					put, listeningPort.PortNumber, listeningPort.Protocol)
 				failedPod = true
+				nonCompliantObjects = append(nonCompliantObjects,
+					testhelper.NewPodReportObject(put.Namespace, put.Name,
+						"Listening port was declared in no container spec", false).
+						SetType(testhelper.ListeningPortType).
+						AddField(testhelper.PortNumber, strconv.Itoa(listeningPort.PortNumber)).
+						AddField(testhelper.PortProtocol, listeningPort.Protocol))
 			}
+			compliantObjects = append(compliantObjects,
```
sebrandon1, line 210: Should this be inside an `else`?  Is it possible to get both a compliant and non-compliant object from the same `put`?

edcdavid:
closed in favor of https://github.com/test-network-function/cnf-certification-test/pull/1207

cnf-certification-test/networking/suite.go
edcdavid: fixed in https://github.com/test-network-function/cnf-certification-test/pull/1207

PR #1209, opened 2023-06-25
Fix multiple typos

rdavid:


jmontesi (approved):

cnf-certification-test/accesscontrol/rbac/automount.go
```diff
@@ -43,7 +43,7 @@ func EvaluateAutomountTokens(put *corev1.Pod) (bool, string) {
 	// if no service account is configured, then the pod will use the configuration
 	// of the default service account in that namespace
 	// the token defined in the pod has takes precedence
-	// the test would pass iif token is explicitly set to false
+	// the test would pass if token is explicitly set to false
```
sebrandon1, line 46: Are we sure this isn't [if and only iff](https://en.wikipedia.org/wiki/If_and_only_if)?

sebrandon1 (approved):

PR #1210, opened 2023-06-26
CNFCERT-541 : Make Pod Role Bindings fail with default service account

bnshr:


cnf-certification-test/accesscontrol/suite.go
```diff
@@ -461,16 +461,21 @@ func testPodRoleBindings(env *provider.TestEnvironment) {
 	var compliantObjects []*testhelper.ReportObject
 	var nonCompliantObjects []*testhelper.ReportObject
 
+	podIsCompliant := true
```
sebrandon1, line 464: I think this should be inside the loop, right?

bnshr: No, why? We do not change its value if the default service account is used.

sebrandon1: `podIsCompliant` is never set back to true.  This would cause every pod to be non-compliant if one was ever found to be non-compliant.

bnshr: I see. I checked the code and it seems we are testing for all pods. It is not fail-fast upon getting one non-compliant pod. I changed the flow. Please review again!

sebrandon1 (approved):
Looks good!  Thanks!

PR #1215, opened 2023-06-27
Add access-control failure reason objects

sebrandon1:
Covers all of the tests in the `access-control` test suite with compliant/non-compliant objects.

Bonus contraction removal `doesn't` to `does not` similar to #264

### DIFF
--- a/cnf-certification-test/accesscontrol/suite.go
+++ b/cnf-certification-test/accesscontrol/suite.go
@@ -504,7 +504,7 @@ func testPodRoleBindings(env *provider.TestEnvironment) {
 			// Add the pod to the non-compliant list
 			nonCompliantObjects = append(nonCompliantObjects,
 				testhelper.NewPodReportObject(put.Namespace, put.Name,
-					"Non-compliant because the pod namespace is either empty or default", false))
+					"The pod namespace is either empty or default", false))
 			podIsCompliant = false
 		} else {
 			logrus.Infof("%s has a serviceAccountName: %s, checking role bindings.", put.String(), put.Spec.ServiceAccountName)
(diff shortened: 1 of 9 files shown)


rdavid (approved):

pkg/testhelper/testhelper.go
```diff
@@ -157,6 +158,8 @@ const (
 	ListeningPortType            = "Listening Port"
 	DeclaredPortType             = "Declared Port"
 	ContainerPort                = "Container Port"
+	HostPort                     = "Host Port"
```
jmontesi, line 161: It looks like object types are used to define the type of a report object and are only used as parameters to the `SetType()` function whereas field types add new data to a report object and are used as parameters to the `AddField()` function. See how `DeclaredPortType` is used for reference.

cnf-certification-test/accesscontrol/suite.go
```diff
@@ -649,14 +694,17 @@ func testSysPtraceCapability(shareProcessPods []*provider.Pod) {
 		}
 		if !sysPtraceEnabled {
 			tnf.ClaimFilePrintf("Pod %s has process namespace sharing enabled but no container allowing the SYS_PTRACE capability.", put.String())
-			podsWithoutSysPtrace = append(podsWithoutSysPtrace, put.String())
+			nonCompliantObjects = append(nonCompliantObjects, testhelper.NewPodReportObject(put.Namespace, put.Name, "Non-compliant because it has process namespace sharing enabled but no container allowing the SYS_PTRACE capability", false))
```
jmontesi, line 697: As we are appending objects to a non-compliant/compliant category adding "Non-compliant because..."/"Compliant because..." sounds a bit redundant to me.

sebrandon1: You are correct I shouldn't have to add that part of the string. I'll adjust where I use that.

pkg/testhelper/testhelper.go
sebrandon1: Can you re-review to see if I'm doing it correct?

jmontesi: Yes. That's the way I think those types are meant to be used.

jmontesi (approved):

PR #1219, opened 2023-06-29
Fix no sshd implementation

bnshr:


cnf-certification-test/accesscontrol/suite.go
```diff
@@ -691,11 +692,23 @@ func testNoSSHDaemonsAllowed(env *provider.TestEnvironment) {
 	var badPods []string
 	var errPods []string
 
-	sshPortInfo := netutil.PortInfo{PortNumber: sshServicePortNumber, Protocol: sshServicePortProtocol}
-
 	for _, put := range env.Pods {
 		cut := put.Containers[0]
 
+		port, err := netutil.GetSSHDaemonPort(cut)
+		if err != nil { // This is fine, as it may be that SSH is not running, hence no sshd_config found
```
sebrandon1, line 699: Would this be a `compliantObject` then?

cnf-certification-test/networking/netutil/netutil.go
```diff
@@ -77,3 +77,13 @@ func GetListeningPorts(cut *provider.Container) (map[PortInfo]bool, error) {
...
 	return parseListeningPorts(outStr)
 }
+
+func GetSSHDaemonPort(cut *provider.Container) (string, error) {
+	const findSSHDaemonPort = "grep -w 'Port' /etc/ssh/sshd_config"
+	outStr, errStr, err := crclient.ExecCommandContainerNSEnter(findSSHDaemonPort, cut)
+	if err != nil || errStr != "" {
+		return "", fmt.Errorf("failed to execute command %s on %s, err: %v", findSSHDaemonPort, cut, err)
+	}
+	portNumber := strings.Fields(outStr)[1]
```
sebrandon1, line 87: Should there be a check for length to make sure it has enough elements first?  Just as a safeguard.

sebrandon1 (commented):
Some comments.

cnf-certification-test/networking/netutil/netutil.go
bnshr: I can put a condition of min 2 length, because in the config file really it could be anything, like Port 2023 # changed by Banashri

cnf-certification-test/networking/netutil/netutil.go
```diff
@@ -77,3 +77,13 @@ func GetListeningPorts(cut *provider.Container) (map[PortInfo]bool, error) {
 
 	return parseListeningPorts(outStr)
 }
+
+func GetSSHDaemonPort(cut *provider.Container) (string, error) {
+	const findSSHDaemonPort = "grep -w 'Port' /etc/ssh/sshd_config"
```
jmontesi, line 82: "[sshd](https://linux.die.net/man/8/sshd)(8) reads configuration data from /etc/ssh/sshd_config (**or the file specified with -f on the command line**)."
So, what if the sshd config was specified using the -f option?

cnf-certification-test/accesscontrol/suite.go
bnshr: This is not non-compliant object. I would not call it compliantObject as we are finding only the non-compliant ones.

cnf-certification-test/networking/netutil/netutil.go
bnshr: Could you show me a command with -f option?

jmontesi: `sshd -f /home/user/.ssh/sshd_config` where `/home/user/.ssh/sshd_config` is a customized config file.

jmontesi: Also the `-p port` option can be used to override the port specified in the `sshd_config` file.

bnshr: user specific sshd config can be used which can override global sshd config. But we cannot know which user is being used. Can we? We can just check if .ssh/sshd_config is present and get the Port from there.

cnf-certification-test/networking/netutil/netutil.go
```diff
@@ -77,3 +77,17 @@ func GetListeningPorts(cut *provider.Container) (map[PortInfo]bool, error) {
...
```
rdavid, line 82:
```suggestion
	const findSSHDaemonPort = "sudo netstat -tpln | grep sshd | head -1 | awk '{ print $4 }' | awk -F : '{ print $2 }'"
```

cnf-certification-test/accesscontrol/suite.go
```diff
@@ -683,19 +684,31 @@ func testNamespaceResourceQuota(env *provider.TestEnvironment) {
...
+			logrus.Error("error occurred while finding ssh port from sshd_config")
+		} else {
+			sshServicePortNumber, err = strconv.Atoi(port)
+			if err != nil {
+				logrus.Error("error occurred while convering port number from string to integer")
```
rdavid, line 705:
```suggestion
				logrus.Error("error occurred while converting port number from string to integer")
```

cnf-certification-test/networking/netutil/netutil.go
bnshr: True, but could not we assume that the user is following best practices? Like keeping the sshd_config in its usual location, under .ssh folder in the home?

cnf-certification-test/networking/netutil/netutil.go
bnshr: I would have loved to use netstat, but this is deprecated in favour of the suite iproute2.

rdavid:
```suggestion
	const findSSHDaemonPort = "sudo ss -tpln | grep sshd | head -1 | awk '{ print $4 }' | awk -F : '{ print $2 }'"
```

cnf-certification-test/networking/netutil/netutil.go
jmontesi: I'm not sure that's a best practice. In any case I keep seeing weak points to this solution. Even if there's a `sshd_config` file in the user's home dir, you cannot be sure that the `-f` option was used with that file or that it was not overwritten with the `-p` option. In order to check this the command used to launch the ssh daemon could be analyzed. This would cover the use of a different port that 22 but not a different name for the daemon process.

bnshr: RH admins either set this config at global level or at the user's home dir .ssh folder, as per the common / best practices. If anywhere else is placed, there must be some reason, that I am not familiar with. But I agree, that case would not be covered.

cnf-certification-test/networking/netutil/netutil.go
```diff
@@ -77,3 +77,13 @@ func GetListeningPorts(cut *provider.Container) (map[PortInfo]bool, error) {
 
 	return parseListeningPorts(outStr)
 }
+
+func GetSSHDaemonPort(cut *provider.Container) (string, error) {
+	const findSSHDaemonPort = "ss -tpln | grep sshd | head -1 | awk '{ print $4 }' | awk -F : '{ print $2 }'"
+	outStr, errStr, err := crclient.ExecCommandContainerNSEnter(findSSHDaemonPort, cut)
+	if err != nil || errStr != "" {
+		return "", fmt.Errorf("failed to execute command %s on %s, err: %v", findSSHDaemonPort, cut, err)
+	}
```
rdavid, line 86:
```suggestion
	if err != nil {
		return "", fmt.Errorf("Failed to execute command %s on %s, err code: %v, err str: %s.", findSSHDaemonPort, cut, err, errStr)
	}
```

cnf-certification-test/accesscontrol/suite.go
```diff
@@ -737,19 +737,32 @@ func testNamespaceResourceQuota(env *provider.TestEnvironment) {
...
-	sshPortInfo := netutil.PortInfo{PortNumber: sshServicePortNumber, Protocol: sshServicePortProtocol}
+	sshServicePortNumber := 22
 
 	for _, put := range env.Pods {
 		cut := put.Containers[0]
 
+		// 1. Find SSH port
+		port, err := netutil.GetSSHDaemonPort(cut)
+		if err != nil {
+			logrus.Errorf("error occurred while finding ssh port on %s, err %v", cut, err)
```
rdavid, line 755: Need to break here, right?

cnf-certification-test/networking/netutil/netutil.go
bnshr: That condition is present in the repo in multiple places from ExecCommandContainer. I don't know what case would give no error, but errStr, but I think there may be some case. Hence, that condition was with OR.

cnf-certification-test/accesscontrol/suite.go
bnshr: No, continue with port default 22 to check if it is listening to be put into non-compliant list in the end.

rdavid (approved):

sebrandon1 (approved):

PR #1222, opened 2023-06-30
chaos-testing: comment out the existing suite as only contains an example test case

jmontesi:
example test case


sebrandon1 (approved):
Looks good.  The `gradetool` config needs to remove the reference to the chaos testing pod-delete test as well.

greyerof (approved):

PR #1233, opened 2023-07-07
Fix for HTML js: avoid crashing when ObjectFieldsKeys or ObjectFields values are null.

greyerof:
The js code was crashing because ObjectFieldsKeys was null in one tc.
Just a workaround to make the html parser to work. Needs further investigation.

### DIFF
--- a/cnf-certification-test/results/html/results-embed.html
+++ b/cnf-certification-test/results/html/results-embed.html
@@ -1537,21 +1537,25 @@ classification=  {
           // Create table header
           var thead = document.createElement('thead');
           var headerRow = document.createElement('tr');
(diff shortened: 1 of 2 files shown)


sebrandon1:
This seems to be a problem that presented itself in #1145.  I worked through the history and this is when the problem started.  This null check might be a part of the problem but is probably something bigger.

sebrandon1:
image

sebrandon1 (approved):
Approved.  We should keep the null check.

Further PRs will have to add a disclaimer at the top warning that the parser tool is from a different released version than the claim file being uploaded.

edcdavid (approved):

PR #1260, opened 2023-07-18
dumps claim file as CSV with telco/non-telco

edcdavid:
New TNF command t dump claim results as CSV with classification information.
Required inputs:
- a list of "telco" CNF names to identify which CNF names are "telco"
- a claim file
- the name of the CNF to which the claim file belongs to
The test case catalog, classification table and telco CNF list are used to do the following:
- the catalog is used to add a remediation column to each test case in the generated CSV
- the classification is used to determine whether a test case is mandatory or optional (and add a column). The determination is based on the test name and CNF type (telco/non-telco)
- add a telco/non-telco column based on the telco CNF list


cmd/tnf/claim/show/csvtelco/csvtelco.go
```diff
@@ -0,0 +1,202 @@
...
+			claimScheme.Claim.Results[testID][0].Description,
+			claimScheme.Claim.Results[testID][0].State,
+			claimScheme.Claim.Results[testID][0].StartTime,
+			claimScheme.Claim.Results[testID][0].EndTime,
+			claimScheme.Claim.Results[testID][0].FailureReason,
+			claimScheme.Claim.Results[testID][0].Output,
+			catalogMap[testID].Remediation,
+		)
+		// if the name of this CNF matches a name in the telco CNF list, all the results are tagged as "telco"
+		if _, ok := telcoCNFMap[CNFNameFlag]; ok {
```
greyerof, line 150: A couple of suggestions:
1. In Golang, maps return the default value for the entry type when the key does not exist, so you don't need to do `if _, ok  := telcoCNFMap[CNFNameFlag]; ok` here. Since the entries are bools, you can simply write:
```
if telcoCNFMap[CNFNameFlag] {
```

2. Also, If I'm not wrong, I think you could get rid of the entire if/else block using a helper string var to store the cnf type:
```
    ...
	cnfType := identifiers.NonTelco
	if telcoCNFMap[CNFNameFlag] {
		cnfType = identifiers.Telco
	}

	for testID := range claimScheme.Claim.Results {
		// get classification map for current test case
		classificationForTestID := identifiers.Classification[testID]
		// initialize record
		record := []string{}
		// creates and appends new CSV record
		record = append(record,
			CNFNameFlag,
			testID,
			claimScheme.Claim.Results[testID][0].Description,
			claimScheme.Claim.Results[testID][0].State,
			claimScheme.Claim.Results[testID][0].StartTime,
			claimScheme.Claim.Results[testID][0].EndTime,
			claimScheme.Claim.Results[testID][0].FailureReason,
			claimScheme.Claim.Results[testID][0].Output,
			catalogMap[testID].Remediation,
		)
		
		// Append the Telco/NonTelco field and its corresponding result type (Mandatory/Optional)
		record = append(record, cnfType, classificationForTestID[cnfType])

		resultsCSVRecords = append(resultsCSVRecords, record)
	}
```

greyerof (approved):
I'm ok if you want to add this new command, but I must say that it sounds too specific to me. The subcommand name itself "csv-telco" restricts it to just a csv output for either telco or non-telco workloads. But what about the other workload types?

I believe we should merge both the "tnf claim show failures" and this new "tnf claim show csv-telco" to create a more generic but unified subcommand to show results from a claim file. This is my proposal:

```
./tnf claim show results --claim/-c path/to/claim.json --output/-o text/json/csv  (default text).
```
If no other modifier flags are provided, this command should print all the results from all the test cases of all test suites, either failed, skipped or passed. For text and json outputs, I'd use the same format as [the current "tnf claim show failures" uses](https://github.com/test-network-function/cnf-certification-test/blob/5b8e10b592af32d6b8d20348254dd28ae2539fe4/cmd/tnf/claim/show/failures/failures.go#L29), although it could be changed in case more/less info should be printed.

Useful, but optional, modifier flags:
Filter test cases based on the test case status. These three optional flags could be combined to show only the failed test cases, both failed and skipped, passed and skipped... let the user decide.
 + --failed/-f: prints only results for failed test cases, as the existing "tnf claim show failures" is doing.
 + --skipped/-s: prints "results" for skipped test cases only.
 + --passed/-p: prints results for passed test cases only.

More optional flags:
 + --cnf-type/-t [telco/non-telco/far-edge/extended]: the user selects the type of workload that was tested. This can be used to add a field on each printed test case to indicate whether it is a Mandatory or an Optional test case for that workload type.
 + --cnf-name/-n : just to add it to the output report.
 + --test-suites/-s "ts1,ts2..." : prints results from the these test suites only. Very useful for debugging. This filter is already implemented in the existing "tnf claim show failures" command.

cmd/tnf/claim/show/csvtelco/csvtelco.go
```diff
@@ -0,0 +1,202 @@
...
+		"Required: claim file.",
+	)
+
+	err := CSVDumpCommand.MarkFlagRequired("claim-file")
+	if err != nil {
+		log.Fatalf("Failed to mark claim file path as required parameter: %v", err)
+		return nil
+	}
+
+	CSVDumpCommand.Flags().StringVarP(&CNFNameFlag, "cnf-name", "o", "",
```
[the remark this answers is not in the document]
edcdavid: fixed

cmd/tnf/claim/show/csvtelco/csvtelco.go
edcdavid: fixed

edcdavid:
> I'm ok if you want to add this new command, but I must say that it sounds too specific to me. The subcommand name itself "csv-telco" restricts it to just a csv output for either telco or non-telco workloads. But what about the other workload types?

using JSON file to import the CNF name to type mapping (in latest commit) makes this command more generic -- to support all types. It is renamed "show csv". In the future, we can update the "show failures" command to support an even more generic export to CSV and various modifiers beyond the simple merge and dump of this command or include the classification/catalog to the claim file directly.

greyerof (approved):

PR #1269, opened 2023-07-20
Fix error handling logic for no SSHd check

bnshr:
Merge it after DCI is green!


ramperher (approved):
LGTM, the issue is now fixed

sebrandon1:
Merging as the problem with the SSHd check seems fixed but the failure in the check seems to be coming from the tarball change.  No space left on device.  Will fix in another PR.

PR #1274, opened 2023-07-21
Use imageID in  all cases for the Digest

edcdavid:
The affiliated certification tests used the Image container field to get the digest where it could be present or not. Now using the imageID field which should always match the value in Pixys database.
Also refactored parsing of the URL to use regex.

### DIFF
--- a/pkg/provider/provider_test.go
+++ b/pkg/provider/provider_test.go
@@ -17,16 +17,15 @@
 package provider
 
 import (
-	"os"
-	"testing"
-
 	"errors"
-
-	corev1 "k8s.io/api/core/v1"
+	"os"
+	"reflect"
+	"testing"
 
 	olmv1Alpha "github.com/operator-framework/api/pkg/operators/v1alpha1"
 	"github.com/stretchr/testify/assert"
 	"github.com/test-network-function/cnf-certification-test/pkg/configuration"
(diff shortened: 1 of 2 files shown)


pkg/provider/provider_test.go
```diff
@@ -302,31 +343,47 @@ func TestIsWorkerNode(t *testing.T) {
 		},
 		{
 			node: &corev1.Node{
-				ObjectMeta: metav1.ObjectMeta{Labels: map[string]string{"node-role.kubernetes.io/master": ""}},
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: map[string]string{"node-role.kubernetes.io/master": ""},
```
sebrandon1, line 347: Should this be `control-plane` instead of `master`?  The string has been deprecated.

sebrandon1 (commented):
One comment about the `master` string deprecation.

ramperher (approved):
LGTM, thanks!

pkg/provider/provider_test.go
edcdavid: This is change is just re-formatting of the current code with golines. Maybe we can address this in another PR ?

sebrandon1: Yes of course!

sebrandon1 (approved):

PR #1290, opened 2023-07-27
Update the error message in some access-control tests

ramperher:
The error message for `access-control-pod-service-account` is a bit confusing, because it makes you think that you are not using a service account, but it may happen that the pods are attached to the default service account.

As described in the [docs](https://github.com/test-network-function/cnf-certification-test/blob/main/CATALOG.md#access-control-pod-service-account), the service account must be a valid one. So, updating the message consequently to emphasise that

Same for `access-control-pod-role-bindings`; error message said namespace but it should say serviceAccountName


rdavid (approved):

greyerof (approved):

ramperher:
Let me add another message that is wrong to this PR

ramperher:
Modiifcation done

PR #1294, opened 2023-07-28
Claim comparison tool improvement: node's CNIs tables.

greyerof:
Refactored the code so the relevant information associated to each node on both claim files is parsed and compared.

This change shows the differences in the CNI networks and its plugins of each node. These tables only make sense when both node names appear in both files.

For each node, a table will show comma separated list of strings, where each string is a category/relevant section that has been compared and it's different. If the node doesn't appear in one file, that will be the only difference.

Example 1:
Both claim file have the same nodes, but one of them has different CNI networks config:
```
NODE                          DIFFERENCES
worker-2.my-cluster.com       CNIs

NODE: worker-2.my-cluster.com
CNI-NETWORK                   DIFFERENCES
crio                          plugins

NODE: worker-2.my-cluster.com, CNI-NETWORK: crio
PLUGIN                        DIFFERENCES
bridge                        ipMasq,ipam
```
The worker-2 node appears in both claim files but its CNI configuration seems to be different. The second table shows which fields of the cni network are different. In this case, there's some plugin in the crio network whose config is different. The last table shows each plugin of the crio network of the worker-2 node and the fields of its config whose value has been found to be different.

Example 2:
```
NODE                          DIFFERENCES
master-1.my-cluster.com       not found in claim1
master-2.my-cluster.com       not found in claim1
master-3.my-cluster.com       not found in claim1
kind-control-plane            not found in claim2
kind-worker                   not found in claim2
kind-worker2                  not found in claim2
```
None of the workers of each file have been found in the other file, so that's the only table we'll get. This could happen if, by mistake, the selected claim files were generated by runs of the CNF Cert Suite in totally unrelated clusters.


cmd/tnf/claim/compare/nodes/cnis/cnis.go
```diff
@@ -0,0 +1,311 @@
...
+// crio                          cniVersion,plugins
+// loopback						 not found in claim2
+//
+// That example shows that the crio network is not the same in both claim files
+// for a given node. The difference column shows a list of "fields" whose value
+// is different. Plugins means that there's at least one crio plugin whose config
+// is different.
+// In case the network appears only in one claim file, the only difference will
+// be "not found in claim[1|2]"
+func (networksDiff CNINetworksDiffReports) String() string {
```
edcdavid, line 65: I feel like  the table format output is missing details to be clear enough. Why not print the json path of the exact object that has differences at every level for instance at plugin level (there might be a better/simple jq command) :
'.claim.nodes.cniPlugins|to_entries[]| select (.key | contains("clus0-0")).value[0].cniVersion' or 
'.claim.nodes.cniPlugins|to_entries[]| select (.key | contains("clus0-0")).value[0].plugins[].ipam'

The user could use this output to get more details about the differences without overloading the output of this command.

edcdavid (approved):

cmd/tnf/claim/compare/nodes/cnis/cnis.go
greyerof: I still want to keep the table as a summary of the fields whose value don't match, but I could add the json paths right after it. What do you think?

greyerof: Ok. As discussed in private, this extra improvements will be done in further PRs.

greyerof:
Closing this PR as I'll upload a new one soon with a different and more flexible approach based on @edcdavid suggestion to traverse json objects without static structs representation.

PR #1303, opened 2023-08-02
add a check for hyperthreading enable

aabughosh:


cnf-certification-test/identifiers/identifiers.go
```diff
@@ -293,6 +294,22 @@ func InitCatalog() map[claim.Identifier]claim.TestCaseDescription {
 		},
 		TagFarEdge)
 
+	TestHyperThreadEnable = AddCatalogEntry(
+		"hyperthread-enable",
+		common.PlatformAlterationTestKey,
+		`Check that pods using hyperthread enable`,
```
sebrandon1, line 300:
```suggestion
		`Check that baremetal workers have hyperthreading enabled`,
```

cnf-certification-test/platform/ishyperthread/ishyperthread.go
```diff
@@ -0,0 +1,48 @@
...
+	for _, match := range matches {
+		num, err = strconv.Atoi(match)
+		if err == nil {
+			fmt.Println("Number:", num)
+		}
+	}
+	if num > 1 {
+		return true, nil
+	}
+	return false, nil
```
sebrandon1, line 42:
```suggestion
	return num > 1, nil
```

cnf-certification-test/platform/suite.go
```diff
@@ -169,6 +175,23 @@ var _ = ginkgo.Describe(common.PlatformAlterationTestKey, func() {
...
 
+func testHyperThreadEnable(env *provider.TestEnvironment) {
+	var compliantObjects []*testhelper.ReportObject
+	var nonCompliantObjects []*testhelper.ReportObject
+	for _, node := range env.Nodes {
+		nodeName := node.Data.Name
+		if ishyperthread.IsBareMetal(node.Data.Spec.ProviderID) {
+			enable, _ := ishyperthread.IsHyperThread(env, nodeName)
+			if enable {
+				compliantObjects = append(compliantObjects, testhelper.NewNodeReportObject(nodeName, "Node has hyperthread enable", true))
```
sebrandon1, line 186:
```suggestion
				compliantObjects = append(compliantObjects, testhelper.NewNodeReportObject(nodeName, "Node has hyperthreading enabled", true))
```

cnf-certification-test/platform/suite.go
```diff
@@ -169,6 +175,23 @@ var _ = ginkgo.Describe(common.PlatformAlterationTestKey, func() {
...
+			} else {
+				nonCompliantObjects = append(nonCompliantObjects, testhelper.NewNodeReportObject(nodeName, "Node has hyperthread disable", false))
```
sebrandon1, line 188:
```suggestion
				nonCompliantObjects = append(nonCompliantObjects, testhelper.NewNodeReportObject(nodeName, "Node has hyperthreading disabled", false))
```

sebrandon1 (changes requested):
Some string changes, looks good though!

Is there some screenshots you could put in this PR with testing it against a baremetal cluster and testing it against Kind?

sebrandon1 (changes requested):
Some comments

cnf-certification-test/platform/ishyperthread/ishyperthread.go
```diff
@@ -0,0 +1,45 @@
...
+	matches := re.FindAllString(str, -1)
+	num := 0
+	// Loop through the matches and convert them to integers (assuming there's only one number)
+	for _, match := range matches {
+		num, err = strconv.Atoi(match)
+		if err == nil {
+			fmt.Println("Number:", num)
+		}
+	}
+	return num > 1, nil
```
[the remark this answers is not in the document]
aabughosh: thats with exec command so im not sure how to do unit test for IsHyperThread function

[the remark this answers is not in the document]
sebrandon1: Just this section with the regex parsing.  The exec command is harder to spoof.

cnf-certification-test/platform/ishyperthread/ishyperthread.go
```diff
@@ -0,0 +1,45 @@
...
+
+func IsHyperThread(env *provider.TestEnvironment, nodeName string) (bool, error) {
+	o := clientsholder.GetClientsHolder()
+	ctx := clientsholder.NewContext(env.DebugPods[nodeName].Namespace, env.DebugPods[nodeName].Name, env.DebugPods[nodeName].Spec.Containers[0].Name)
+	cmdValue, errStr, err := o.ExecCommandContainer(ctx, isHyperThreadCommand)
+	if err != nil || errStr != "" {
+		return false, fmt.Errorf("cannot execute %s on debug pod %s, err=%s, stderr=%s", isHyperThreadCommand, env.DebugPods[nodeName], err, errStr)
+	}
+	str := cmdValue // Replace with your input string containing the number
+	num := extractNumber(str)
```
sebrandon1, line 25:
```suggestion
	num := extractNumber(cmdValue)
```

cnf-certification-test/platform/suite.go
```diff
@@ -169,6 +175,23 @@ var _ = ginkgo.Describe(common.PlatformAlterationTestKey, func() {
 
 })
 
+func testHyperThreadEnable(env *provider.TestEnvironment) {
+	var compliantObjects []*testhelper.ReportObject
+	var nonCompliantObjects []*testhelper.ReportObject
+	for _, node := range env.Nodes {
+		nodeName := node.Data.Name
+		if ishyperthread.IsBareMetal(node.Data.Spec.ProviderID) {
+			enable, err := ishyperthread.IsHyperThread(env, nodeName)
```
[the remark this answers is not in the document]
aabughosh: its going with enable so i added if it ==nil (also its ok without)but if u see the return value is true and nill if it ok
otherwise false

cnf-certification-test/platform/suite.go
```diff
@@ -169,6 +175,26 @@ var _ = ginkgo.Describe(common.PlatformAlterationTestKey, func() {
 
 })
 
+func testHyperThreadEnable(env *provider.TestEnvironment) {
+	var compliantObjects []*testhelper.ReportObject
+	var nonCompliantObjects []*testhelper.ReportObject
+	for _, node := range env.Nodes {
+		nodeName := node.Data.Name
+		if ishyperthread.IsBareMetal(node.Data.Spec.ProviderID) {
```
edcdavid, line 183: move this function to provider/nodes.go

cnf-certification-test/platform/ishyperthread/ishyperthread.go
```diff
@@ -0,0 +1,44 @@
...
+
+	"github.com/test-network-function/cnf-certification-test/internal/clientsholder"
+	"github.com/test-network-function/cnf-certification-test/pkg/provider"
+)
+
+const (
+	isHyperThreadCommand = "lscpu | grep \"Thread(s) per core\""
+)
+
+func IsHyperThread(env *provider.TestEnvironment, nodeName string) (bool, error) {
```
edcdavid, line 17: move this function to provider/nodes.go

cnf-certification-test/platform/ishyperthread/ishyperthread.go
```diff
@@ -0,0 +1,44 @@
...
+	"regexp"
+	"strconv"
+	"strings"
+
+	"github.com/test-network-function/cnf-certification-test/internal/clientsholder"
+	"github.com/test-network-function/cnf-certification-test/pkg/provider"
+)
+
+const (
+	isHyperThreadCommand = "lscpu | grep \"Thread(s) per core\""
```
edcdavid, line 14: use only lscpu and create a function similar to the extractNumber function to parse the output directly with regex
or
use lscpu --json and unmarshall the json output to get the number of threads per core

cnf-certification-test/platform/ishyperthread/ishyperthread.go
edcdavid: Hyperthreading is a property of a node, so we can have a node.IsHyperthread() member function

pkg/provider/provider.go
```diff
@@ -576,3 +576,12 @@ func createNodes(nodes []corev1.Node) map[string]Node {
 
 	return wrapperNodes
 }
+func GetBaremetalNodes(env *TestEnvironment) []Node {
+	var baremetalNodes []Node
+	for _, node := range env.Nodes {
+		if strings.HasPrefix(node.Data.Spec.ProviderID, "baremetalhost://") {
```
sebrandon1, line 582: I feel like this would be a good built-in function for our `Node` objects similar to the other ones [here](https://github.com/test-network-function/cnf-certification-test/blob/main/pkg/provider/nodes.go).

Then you could write `if node.IsBaremetal`.

aabughosh: david asked to be like this and i think its better becuse of the skip if empty

pkg/provider/provider.go
```diff
@@ -576,3 +576,12 @@ func createNodes(nodes []corev1.Node) map[string]Node {
 
 	return wrapperNodes
 }
+func GetBaremetalNodes(env *TestEnvironment) []Node {
```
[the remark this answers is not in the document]
aabughosh: to implement a test env? will be huge i know that all will be empty but what u think?

pkg/provider/provider_test.go
```diff
@@ -862,3 +862,72 @@ func Test_buildContainerImageSource(t *testing.T) {
 		})
 	}
 }
+
+func TestTestEnvironment_GetBaremetalNodes(t *testing.T) {
```
sebrandon1, line 866:
```suggestion
func TestGetBaremetalNodes(t *testing.T) {
```

pkg/provider/provider_test.go
```diff
@@ -862,3 +862,72 @@ func Test_buildContainerImageSource(t *testing.T) {
...
+	type fields struct {
+		Nodes map[string]Node
```
sebrandon1, line 868: Why use a `map`?  Seems like you would just need a `[]Node`.

sebrandon1: Then just compare the slices.  Actually in Go 1.21 they added a `slices` comparison tool which we could take advantage of after we move to 1.21.

sebrandon1 (commented):
Some comments about the `map`

pkg/provider/nodes.go
```diff
@@ -122,3 +125,25 @@ func (node *Node) GetRHELVersion() (string, error) {
 
 	return longVersionSplit[0], nil
 }
+
+const (
+	expectedValue        = 2
+	isHyperThreadCommand = "lscpu"
```
edcdavid, line 131: needs access to the host filesystem "lscpu" -> "chroot /host lscpu"

edcdavid (approved):

pkg/provider/provider_test.go
aabughosh: i need it as that type since that what we are using in env

sebrandon1 (approved):

PR #1306, opened 2023-08-02
Remove gradetool image to save space

sebrandon1:
Similar to how we cleanup the `oct` image.

PR #1309, opened 2023-08-03
Adapt the remaining networking tests to the new report objects

jmontesi:


sebrandon1:
@jmontesi I _think_ I fixed the space issue in the runners with #1306.  Can you rebase this PR and see?

cnf-certification-test/networking/suite.go
```diff
@@ -322,41 +323,52 @@ func testNetworkPolicyDenyAll(env *provider.TestEnvironment) {
 		}
```
edcdavid, line 323: line 316: add a network policy level reporting object (new type ), to list all the policies that are triggering the non-compliance  and compliance.

edcdavid: as I understand it LabelMatch (checks that podSelectorLabels is empty or matches the pod) and IsNetworkPolicyCompliant (checks that egress, ingress or both rules are present) together check that rules are enabled for the pod, so only the specified rules will be allowed.  
If a pod is not compliant to this test, then we would not know which network policy were tested and non compliant to the test. For instance if there are rules present but not with both ingress and ingress. In my opinion this would pinpoint the issue. On the other hand, if the test passes we can visually check the rules that are configured

edcdavid: Adding report objects per pod and per network policy would create a lot of reports objects which we are not ready to properly handle now in the result page. If needed we can come back to add it.

edcdavid (approved):

sebrandon1 (approved):

PR #1319, opened 2023-08-07
Run docker and clean in Github action

rdavid:


sebrandon1 (approved):

rdavid:
@sebrandon1 Something wrong is [here](https://github.com/test-network-function/cnf-certification-test/actions/runs/5790703780/job/15694260514?pr=1319), I've changed parameter order, the same line works on Local Smoke.

sebrandon1:
@rdavid Remember the env vars: ${{ env.TNF_OUTPUT_DIR }}/claim.json in the smoke tests section

.github/workflows/pre-main.yaml
```diff
@@ -214,24 +214,26 @@ jobs:
...
-
+      # Update the CNF containers, helm charts and operators DB.
       - name: Update the CNF DB
         run: |
-          docker pull ${REGISTRY}/${OCT_IMAGE_NAME}:${OCT_IMAGE_TAG}
           mkdir -p "${GITHUB_WORKSPACE}"/offline-db
-          docker run -v "${GITHUB_WORKSPACE}"/offline-db:/tmp/dump:Z --env OCT_DUMP_ONLY=true ${REGISTRY}/${OCT_IMAGE_NAME}:${OCT_IMAGE_TAG}
+          docker run \
+            --env OCT_DUMP_ONLY=true \
+            --rm \
```
jmontesi, line 223: Probably unnecessary if after this docker command you issue `docker system prune -a`.

sebrandon1 (approved):

jmontesi (approved):

.github/workflows/pre-main.yaml
sebrandon1: Honestly we're probably good with #1320 but these changes can't hurt.

rdavid: @jmontesi `--rm` came first. `podman` has `--rmi` parameter for the image. `docker` doesn't, that is why `docker system prune -a` for image.

jmontesi: I meant that `docker system prune -a` removes the container as well, that's why I think `--rm` might be unnecessary there. But it's not a big deal anyway.

rdavid: I got you. Agree on 'But it's not a big deal anyway.' :-) Thanks.

PR #1324, opened 2023-08-08
remove deprecated focus strings feature

edcdavid:


jmontesi:
I believe the _skip_ filtering mechanism should also be removed. From the Ginkgo's doc:

> The description-based `--focus` and `--skip` flags were Ginkgo's original command-line based filtering mechanism and will continue to be supported - however we recommend using labels when possible as the label filter language is more flexible and easier to reason about.

sebrandon1 (approved):

script/run-container.sh
```diff
@@ -179,4 +179,4 @@ ${TNF_CONTAINER_CLIENT} run --rm $DNS_ARG \
 	-e TNF_INCLUDE_WEB_FILES_IN_OUTPUT_FOLDER=$TNF_INCLUDE_WEB_FILES_IN_OUTPUT_FOLDER \
 	-e PATH=/usr/bin:/usr/local/oc/bin \
 	$TNF_IMAGE \
-	$TNF_CMD $OUTPUT_ARG $CONTAINER_TNF_DIR/claim $FOCUS_ARG $TNF_FOCUS_SUITES $SKIP_ARG $TNF_SKIP_SUITES $LABEL_ARG $TNF_LABEL "$@"
+	$TNF_CMD $OUTPUT_ARG $CONTAINER_TNF_DIR/claim $SKIP_ARG $TNF_SKIP_SUITES $LABEL_ARG $TNF_LABEL "$@"
```
jmontesi, line 182: Delete the SKIP references as well.

jmontesi (approved):

script/run-container.sh
edcdavid: fixed

PR #1332, opened 2023-08-10
Remove 'make install' call from QE jobs

sebrandon1:
We don't need the -partner repo resources installed for QE testing.  The QE repo installs all of its necessary testing resources itself.

Builds on: https://github.com/test-network-function/cnf-certification-test-partner/pull/323


sebrandon1:
Closing this, working on other things we can revisit later.

PR #1340, opened 2023-08-15
Mark lifecycle with 100% QE test coverage

bnshr:


.github/actions/setup/action.yml
```diff
@@ -19,3 +21,14 @@ runs:
...
+    - name: Install cr-scale-operator
+      run: |
+        rm -rf cr-scale-operator
+        git clone https://github.com/test-network-function/cr-scale-operator
+        cd cr-scale-operator
+        IMG="quay.io/testnetworkfunction/cr-scale-operator"
+        make deploy IMG=$IMG
+        cd ..
+        rm -rf cr-scale-operator
+      shell: bash
```
sebrandon1, line 34: I don't think this is going to work as the Kind clusters are not ready at this point.

sebrandon1 (changes requested):
Remove change to the `.github/actions/setup/action.yml`, otherwise looks good!

.github/actions/setup/action.yml
bnshr: I think we should not run QE tests in Kind cluster from this repo. Can we turn this off entirely? The reasons are as follows:
1. Delay in the feedback of the test development.
2. We cannot have all pre-conditions of QE tests fulfilled from for QE tests (here, for example this preinstallation). There may be many more that QE can share.
3. I do not think that we have sufficient infrastructure ready to run the code of two repos in a pipeline. Please correct me if I am wrong.
4. If we want to keep QE code testing in this pipeline, we can run that only upon merging the PR to the main branch. Normal, harmless PR are not blocked in that case due to slow feedback cycle.

sebrandon1: 1. Currently, the QE tests only get ran once a day (maybe) which doesn't give us traceability into what breaks QE.  The slowest QE suite is the lifecycle suite and it takes ~40 minutes to run which isn't terrible.  Prior to the QE PR checks our slowest smoke tests runs were ~20 minutes anyways.
2. Installing an operator to test in QE should be done as part of the QE test in the `BeforeEach` section.  It should all be self-contained.  We don't install any tested resources as part of the QE PR pipeline.
3. We currently have [5 self-hosted runners](https://github.com/test-network-function/cnfcert-tests-verification/blob/main/playbooks/inventory.ini) handling requests.  There are only [4 suites](https://github.com/test-network-function/cnf-certification-test/blob/main/.github/workflows/qe.yaml#L19) currently enabled.  This should be fine until we add more test suites to test against.
4. Once PRs are merged that break QE they will need reverts anyways, why not just test them initially to prevent breakage?

sebrandon1: I have temporarily disabled the QE PR tests here #1341 while I fine tune these to get faster.

bnshr: 2. We are not installing the operator from OLM. So, operator installation is done from the cloned repo. We decided to have that as part of QE script and check for the operator namespace.

.github/actions/setup/action.yml
```diff
@@ -19,3 +21,14 @@ runs:
...
+    #- name: Install cr-scale-operator
+    #  run: |
+    #    rm -rf cr-scale-operator
+    #    git clone https://github.com/test-network-function/cr-scale-operator
+    #    cd cr-scale-operator
+    #    IMG="quay.io/testnetworkfunction/cr-scale-operator"
+    #    make deploy IMG=$IMG
+    #    cd ..
+    #    rm -rf cr-scale-operator
+    #  shell: bash
```
sebrandon1, line 34: Can you remove the commented section?  Also the changes to the `description`?

bnshr: I think I have to add it back when you enable QE tests.

bnshr: What is description?

sebrandon1: Even if we uncommented this, the clusters at this point aren't spawned so the command to deploy would fail.

sebrandon1: We'll add this as part of the BeforeEach in the QE repo so it can run anywhere smoothly.

sebrandon1:
Can you remove the changes here as well?  

image


Just leave the QE flag change and that should be good-to-go.

bnshr:
Screenshot 2023-08-31 at 15 48 16
This is how the lines look like.

PR #1342, opened 2023-08-15
Add envconfig option for OmitArtifactsZipFile

sebrandon1:
Add the ability to set this variable via the CLI.

### DIFF
--- a/pkg/configuration/configuration.go
+++ b/pkg/configuration/configuration.go
@@ -157,5 +157,5 @@ type TestParameters struct {
 	AllowPreflightInsecure        bool   `split_words:"true"`
 	PfltDockerconfig              string `split_words:"true" envconfig:"PFLT_DOCKERCONFIG"`
 	IncludeWebFilesInOutputFolder bool   `split_words:"true" default:"false"`
-	OmitArtifactsZipFile          bool   `split_words:"true" default:"false"`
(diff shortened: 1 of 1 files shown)


rdavid (approved):

pkg/configuration/configuration.go
```diff
@@ -157,5 +157,5 @@ type TestParameters struct {
 	AllowPreflightInsecure        bool   `split_words:"true"`
 	PfltDockerconfig              string `split_words:"true" envconfig:"PFLT_DOCKERCONFIG"`
 	IncludeWebFilesInOutputFolder bool   `split_words:"true" default:"false"`
-	OmitArtifactsZipFile          bool   `split_words:"true" default:"false"`
+	OmitArtifactsZipFile          bool   `split_words:"true" default:"false" envconfig:"OMIT_ARTIFACTS_ZIP_FILE"`
```
greyerof, line 160: Why is this change needed? It is already possible to set this variable through the env var `TNF_OMIT_ARTIFACTS_ZIP_FILE` (see [doc](https://test-network-function.github.io/cnf-certification-test/test-standalone/#output-targz-file-with-results-and-web-viewer-files)). If I'm not wrong, your change will only remove the "TNF_" prefix of the expected env var.

PR #1353, opened 2023-08-21
Add tnf.Log() function to allow log level selection

jmontesi:
It avoids the common pattern of calling `tnf.ClaimFilePrint()` followed or preceded by a `logrus` function to print a log with a different level than `TRACE`. It also avoids log duplication when doing this.

The issue is fixed in a couple of places as an example, not with the intent of being exhaustive.

### DIFF
--- a/pkg/tnf/status.go
+++ b/pkg/tnf/status.go
@@ -9,11 +9,16 @@ import (
 
 // ClaimFilePrintf prints to claim and junit report files.
 func ClaimFilePrintf(format string, args ...interface{}) {
+	Log(logrus.TraceLevel, format, args)
+}
+
+// Log prints to stdout and claim and junit report files.
+func Log(level logrus.Level, format string, args ...interface{}) {
 	message := fmt.Sprintf(format+"\n", args...)
 	_, err := ginkgo.GinkgoWriter.Write([]byte(message))
 	if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/tnf/status.go
```diff
@@ -9,11 +9,16 @@ import (
...
 func ClaimFilePrintf(format string, args ...interface{}) {
+	Log(logrus.TraceLevel, format, args)
+}
+
+// Log prints to stdout and claim and junit report files.
+func Log(level logrus.Level, format string, args ...interface{}) {
 	message := fmt.Sprintf(format+"\n", args...)
 	_, err := ginkgo.GinkgoWriter.Write([]byte(message))
 	if err != nil {
 		logrus.Errorf("Ginkgo writer could not write msg '%s' because: %s", message, err)
```
rdavid, line 20: Should it return on error? Or put trace to the else section?

pkg/tnf/status.go
```diff
@@ -9,11 +9,16 @@ import (
...
-	} else {
-		logrus.Trace(message)
 	}
+
+	logrus.StandardLogger().Log(level, message)
```
rdavid, line 23: I'd move it to right under the message, line 18.

rdavid (approved):

sebrandon1 (approved):

edcdavid (approved):

PR #1381, opened 2023-09-04
GitHub actions use Depends-On: strings in the PR description

fredericlepied:
This allows to have dependencies between PR in the Github actions like in DCI.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -64,7 +64,7 @@ build:
 		build-cnf-tests \
 		test
 
-build-tnf-tool:
+build-tnf-tool: go.sum
 	go build -o tnf -v cmd/tnf/main.go
 
 # Cleans up auto-generated and report files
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -117,9 +117,12 @@ build-cnf-tests-debug: results-html
 	PATH=${PATH}:${GOBIN} ginkgo build -gcflags "all=-N -l" -ldflags "${LINKER_TNF_RELEASE_FLAGS} -extldflags '-z relro -z now'" ./cnf-certification-test
 
 # Installs build tools and other required software.
-install-tools:
+install-tools: go.sum
 	go install "$$(awk '/ginkgo/ {printf "%s/ginkgo@%s", $$1, $$2}' go.mod)"
 
+go.sum: go.mod
+	go mod tidy
+
```
sebrandon1, line 125: Are these changes needed?

fredericlepied: yes because when the dependency is injected you need to run this command.

sebrandon1: The go.sum shouldn't need to be recreated at this point.  Unless I'm missing something that has to do with the new depends-on action?

fredericlepied: The process of injecting a change is modifying `go.mod` so `go.sum` needs to be regenerated.

sebrandon1 (approved):

PR #1384, opened 2023-09-05
Remove automount and projected volume tests

sebrandon1:
We have determined that the existing pod-service-account test covers the scenario quite well where we only enforce that the user must have a non-default service account specified.

We will have to remove the QE tests and the references to the best practice document as well.


edcdavid (approved):

sebrandon1:
@edcdavid should we close this PR in favor of #1385?

sebrandon1:
After some discussion about the relevance of these tests and some reading of the documentation, it seems that we probably do not need to keep the projected volume test.  See the following raw YAML:

image

If you are are attempting to create a pod with a service account similar to the above, in k8s 1.22+ the service account tokens are automatically added as a projected volume mount in pods:

image

It ends up looking like:

image

sebrandon1:
Closing in favor of: https://github.com/test-network-function/cnf-certification-test/pull/1428

PR #1385, opened 2023-09-06
Check for default SA in projected volume test

sebrandon1:
Changes the projected volume test to only look for `system:serviceaccount:default:default` service account "audiences" to allow for users to use projected service account tokens that aren't default.  This matches other tests where we prevent users from using default service account token references.

We have also discussed completely removing the projected SA and automount tests in #1384.

This discussion is ongoing but we wanted to see the different options.


sebrandon1:
Marked WIP because the `audience` field is considered optional according to the docs so we will have to figure out a new way to tell which service account is bound via projected volume.

sebrandon1:
Closing in favor of: https://github.com/test-network-function/cnf-certification-test/pull/1428

PR #1389, opened 2023-09-07
config: use a string slice to hold the target namespaces

jmontesi:
It substitutes the previous struct which only had one field for the namespace name.

Also, targetNamespaces is used instead of targetNameSpaces in the config file.


sebrandon1 (approved):

edcdavid (approved):
Mark this PR as non backward compatible

PR #1404, opened 2023-09-11
Use debug pod for getCurrentKernelCmdlineArgs

sebrandon1:
`getCurrentKernelCmdlineArgs` was previously using the partner workload containers to issue the command "cat /proc/cmdline" instead of the debug pods.

### DIFF
--- a/cnf-certification-test/platform/bootparams/bootparams.go
+++ b/cnf-certification-test/platform/bootparams/bootparams.go
@@ -29,7 +29,7 @@ import (
 
 const (
 	grubKernelArgsCommand = "cat /host/boot/loader/entries/$(ls /host/boot/loader/entries/ | sort | tail -n 1)"
-	kernelArgscommand     = "cat /proc/cmdline"
+	kernelArgscommand     = "cat /host/proc/cmdline"
 )
 
 func TestBootParamsHelper(env *provider.TestEnvironment, cut *provider.Container) (claimsLog loghelper.CuratedLogLines, err error) {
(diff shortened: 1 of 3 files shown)


cnf-certification-test/platform/bootparams/bootparams.go
```diff
@@ -94,12 +94,12 @@ func getGrubKernelArgs(env *provider.TestEnvironment, nodeName string) (aMap map
 	return arrayhelper.ArgListToMap(grubSplitKernelConfig), nil
 }
 
-func getCurrentKernelCmdlineArgs(cut *provider.Container) (aMap map[string]string, err error) {
+func getCurrentKernelCmdlineArgs(env *provider.TestEnvironment, nodeName string) (aMap map[string]string, err error) {
 	o := clientsholder.GetClientsHolder()
-	ctx := clientsholder.NewContext(cut.Namespace, cut.Podname, cut.Name)
+	ctx := clientsholder.NewContext(env.DebugPods[nodeName].Namespace, env.DebugPods[nodeName].Name, env.DebugPods[nodeName].Spec.Containers[0].Name)
 	currentKernelCmdlineArgs, errStr, err := o.ExecCommandContainer(ctx, kernelArgscommand)
```
edcdavid, line 100: why not use "outStr, errStr, err := crclient.ExecCommandContainerNSEnter(getListeningPortsCmd, cut)" instead. We would keep the same logic but not rely on commands installed on the container

sebrandon1: Yes this make sense, I will switch.

greyerof: @edcdavid I don't think we need `nsenter` here, as we're cat'ing the cmdline file from each OCP node's OS using the "cat" command that (we know) should exist in our debug pods (UBI-based images). I believe we should also add the `/host` prefix to the `kernelArgsCommand`:
```
	kernelArgscommand     = "cat /host/proc/cmdline"
```

greyerof (approved):

edcdavid (approved):

greyerof (approved):

PR #1414, opened 2023-09-13
add new test thats cheking of control plan has disabled isecure-port if it has the test will fail

aabughosh:


sebrandon1 (commented):
After doing some reading into this, I'm not sure this is something we can test for through the API.  This is an api-server flag that has to be included during the startup of the api-server itself rather than checking any running pod on the cluster.

This issue [here](https://github.com/kubernetes/kubernetes/issues/91506) discusses the insecure-port feature and mentions that in k8s 1.20 the feature is completely removed.

The PR that actually removed the insecure port was [here](https://github.com/kubernetes/kubernetes/pull/95856) for reference.

I don't think we need to test this as OCP 4.7 was the version that brought in k8s 1.20 api-server according to this page: https://access.redhat.com/solutions/4870701

OCP 4.7 is in end-of-life and reached end of maintenance on August 24, 2022.

https://access.redhat.com/support/policy/updates/openshift

PR #1417, opened 2023-09-13
Fix linting error

jmontesi:


sebrandon1:
I think this was covered by #1415

PR #1218, opened 2023-06-28
Add affiliated-certification failure reason objects

sebrandon1:
Added failure reason objects for:
- `affiliated-certification-helm-version`
- `affiliated-certification-container-is-certified`
- `affiliated-certification-operator-is-certified`
- `affiliated-certification-helmchart-is-certified`
- `affiliated-certification-container-is-certified-digest`

Added two new funcs in the testhelper:
- `NewHelmChartReportObject`
- `NewOperatorReportObject`

PR #1418, opened 2023-09-13
Add report objects for testContainerCertificationStatus

sebrandon1:
This is a leftover from #1218 where this modifications to this test were missing.

* Created a new func `NewCertifiedContainerReportObject` that will print the details of the certified container object to the user.

### DIFF
--- a/pkg/testhelper/testhelper.go
+++ b/pkg/testhelper/testhelper.go
@@ -22,6 +22,7 @@ import (
 	"reflect"
 
 	"github.com/sirupsen/logrus"
+	"github.com/test-network-function/cnf-certification-test/pkg/configuration"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


pkg/testhelper/testhelper.go
```diff
@@ -214,6 +219,15 @@ func NewContainerReportObject(aNamespace, aPodName, aContainerName, aReason stri
 	return out
 }
 
+func NewCertifiedContainerReportObject(cii configuration.ContainerImageIdentifier, aReason string, isCompliant bool) (out *ReportObject) {
+	out = NewReportObject(aReason, ContainerType, isCompliant)
```
jmontesi, line 223: It may be better to use a new ContainerImageType.

jmontesi (approved):

pkg/testhelper/testhelper.go
sebrandon1: Good idea, I'll add that.

sebrandon1: Done.

greyerof (approved):

PR #1428, opened 2023-09-14
Remove projected volume test

sebrandon1:
We are removing the projected volume test as projected volumes using service account tokens are automatically generated and added to pods via Kubernetes 1.22+.  The existing `pod-service-account` test covers the "non-default" criteria for passing/failing.

From the [official documentation](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#manually-create-an-api-token-for-a-serviceaccount): 
image

For example, if you build a raw pod YAML such as:
image

It ends up looking like this after you apply it:
image

The service account token is automatically added to the pod.


ramperher (approved):
We had experience with partners with this test case and it's happening that these projected volumes are automatically created and there's no way to pass this test, so it looks fine to the Telco CI team that this is eventually removed here

edcdavid (approved):

sebrandon1:
@ramperher Yes we'll be releasing a new version with this change.  👍

PR #1439, opened 2023-09-15
Fix: run container-is-certified also in the case of digest only images

edcdavid:
This is a regression compared with previous logic. Containers with digest **only** images would not be tested in the container-is-certified test, or in the container-is-certified-digest test. 

For digest images, the repo and name fields are no longer populated since they could be anything, only the digest counts. However, since we skip containers with images that have both empty repository and name we would also skip digest only images. This same check in the container-is-certified-digest test would skip containers that have only a digest defined, only testing the image digest when it was defined using a repository, name and tag. This fix remove the unnecessary check in both container-is-certified and container-is-certified-digest tests.
PR causing regression: https://github.com/test-network-function/cnf-certification-test/pull/1274/files

build-depends: 29479

### DIFF
--- a/cnf-certification-test/certification/suite.go
+++ b/cnf-certification-test/certification/suite.go
@@ -116,10 +116,6 @@ func testContainerCertificationStatus(env *provider.TestEnvironment, validator c
 	ginkgo.By(fmt.Sprintf("Getting certification status. Number of containers to check: %d", len(containersToQuery)))
 	allContainersToQueryEmpty := true
 	for c := range containersToQuery {
-		if c.Repository == "" || c.Registry == "" {
-			tnf.ClaimFilePrintf("Container name = \"%s\" or repository = \"%s\" is missing, skipping this container to query", c.Repository, c.Registry)
-			continue
-		}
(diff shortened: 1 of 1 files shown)


sebrandon1 (approved):

ramperher:
I've included a dependency with DCI code where we included an update in the containter certification info provided to tnf_config.yml, now the job is passing here

aabughosh (approved):

edcdavid:
> Seems like the PR tests are failing with:
> 
> ```
> TRACE  [Sep 18 15:24:44.204][status.go: 18] {"CompliantObjectsOut":null,"NonCompliantObjectsOut":[{"ObjectType":"Operator","ObjectFieldsKeys":["Reason For Non Compliance","Namespace","Name","OCP Version","OCP Channel"],"ObjectFieldsValues":["Operator failed to be certified for OpenShift","tnf","hazelcast-platform-operator.v5.7.0","","alpha"]}]}
> ```

I am assuming it is a temporary issue with the online db? Seems to pass now

PR #1440, opened 2023-09-15
Add the name of the object responsible for skipping in addition to type

edcdavid:
Currently only the type of objects responsible for skipping test cases is displayed, add also the name of the object
before we had []*Pods, which becomes []*Pods ( env.GetHugepagesPods() )
Screenshot from 2023-09-15 15-09-02

### DIFF
--- a/cnf-certification-test/accesscontrol/suite.go
+++ b/cnf-certification-test/accesscontrol/suite.go
@@ -65,121 +65,121 @@ var _ = ginkgo.Describe(common.AccessControlTestKey, func() {
 	ginkgo.ReportAfterEach(results.RecordResult)
 	testID, tags := identifiers.GetGinkgoTestIDAndLabels(identifiers.TestSecContextIdentifier)
 	ginkgo.It(testID, ginkgo.Label(tags...), ginkgo.Label(tags...), func() {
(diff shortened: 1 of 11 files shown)


greyerof (approved):

aabughosh:
why not to emplent somthing that printing type of specific object? and not adding that string 
like this 
// Define an interface to retrieve the object's type.
type TypedObject interface {
	GetType() string
}

// Implement the GetType() method for Container.
func (c Container) GetType() string {
	return "env.container"
}

// Implement the GetType() method for Pods.
func (p Pods) GetType() string {
	return "env.pods"
}

// Implement the GetType() method for Deployment.
func (d Deployment) GetType() string {
	return "env.deployment"
}

func PrintObjectType(obj TypedObject) {
	fmt.Printf("Type: %s\n", obj.GetType())
}

edcdavid:
> why not to emplent somthing that printing type of specific object? and not adding that as a string like this // Define an 
The previous code was already printing the type (e.g. Pods, Deployments, etc). This PR adds a manual string to describe the object instance, not type, as well as we can. For instance, "Pod with DPDK configued", or "Debug pods ". Both are still Pods type, but are different instances

sebrandon1 (approved):

PR #1449, opened 2023-09-19
Bug fix: reduce the TNF image size

jmontesi:
In a Docker multi-stage build the value of the environment variables is not preserved between different stages, so if they want to be used they must be redefined. In this case the env var OSDK_BIN was used with an empty value, causing the COPY operation to bring the whole filesystem of the builder image to the final TNF image.

The fix reduces the TNF image size from 787MB to 355MB.


rdavid (approved):

rdavid:
Nice catch!

sebrandon1 (approved):

PR #1450, opened 2023-09-19
New tool to verify the certification status of container images

jmontesi:
New subcommand under `tnf ` to manually verify the certification status of container images. To use it, provide the `name`, `registry` and `tag` (defaults to latest) of an image as parameters or simply the `digest`. For example:

`./tnf check image-cert-status --name rocketchat/rocketchat --registry registry.connect.redhat.com`

`./tnf check image-cert-status --digest sha256:03f7f2499233a302351821d6f78f0e813c3f759258184f4133144558097c57b0`

It can be launched as a standalone binary if compiled directly with `make build-tnf-tool` or using the TNF image with Docker or Podman as follows:
`
docker run <tnf-image> tnf check image-cert-status --digest sha256:03f7f2499233a302351821d6f78f0e813c3f759258184f4133144558097c57b0`

The change also makes available all the commands under` tnf `directly from the TNF image.


cmd/tnf/check/check.go
```diff
@@ -0,0 +1,19 @@
+package check
+
+import (
+	"github.com/spf13/cobra"
+	imagecert "github.com/test-network-function/cnf-certification-test/cmd/tnf/check/image_cert_status"
+)
+
+var (
+	checkCmd = &cobra.Command{
+		Use:   "check",
```
edcdavid, line 10: I feel check is to vague, what about certification?

edcdavid (approved):

cmd/tnf/check/check.go
sebrandon1: I believe that `preflight` also uses the "check":

image

sebrandon1 (approved):

cmd/tnf/check/check.go
sebrandon1: https://github.com/redhat-openshift-ecosystem/openshift-preflight#usage

PR #1451, opened 2023-09-19
CNFCERT-612:Cluster-admin permission: Check which subjects have cluster-admin

aabughosh:
CNFCERT-612: Cluster-admin permission: Check which subjects have cluster-admin RBAC permissions


cnf-certification-test/accesscontrol/suite.go
```diff
@@ -236,8 +236,44 @@ var _ = ginkgo.Describe(common.AccessControlTestKey, func() {
...
+	var nonCompliantObjects []*testhelper.ReportObject
+	for _, pod := range env.Pods {
+		serviceAccountName := pod.Spec.ServiceAccountName
+		tnf.ClaimFilePrintf("Pod: %s, ServiceAccount: %s\n", pod.Name, serviceAccountName)
+
+		// Check if there's a "cluster-admin" ClusterRoleBinding for the service account.
+		hasClusterAdminBinding := false
+		for i := range env.ClusterRoleBindings {
+			for _, subject := range env.ClusterRoleBindings[i].Subjects {
+				if subject.Name == serviceAccountName && subject.Kind == "ServiceAccount" && env.ClusterRoleBindings[i].RoleRef.Name == "cluster-admin" {
```
edcdavid, line 257: Isn't this test case contradicting the "cluster-role-bindings"? cluster-role-bindings test would fail a pod with cluster role bindings, but this test will pass it as long as it is not "cluster-admin". Shoudn't we update the "cluster-role-bindings" test to report the "env.ClusterRoleBindings[i].RoleRef.Name" as part as the report object for non-compliant pods?

sebrandon1: Agreed this is probably the best way forward.  I forgot about the existing cluster-role-bindings test.

Let's just add more detail into the existing test report objects like David suggested.

aabughosh: yah i see we have same test

pkg/testhelper/testhelper.go
```diff
@@ -106,6 +106,7 @@ const (
 	ReasonForCompliance             = "Reason For Compliance"
 	Category                        = "Category"
 	RoleBindingName                 = "Role Binding Name"
+	ClusterRoleBindingName          = "Cluster Role Binding Name"
```
sebrandon1, line 109: Should this be `Cluster Role Name`?  I think the roleRef references the Cluster Role itself.

image

sebrandon1: https://kubernetes.io/docs/reference/access-authn-authz/rbac/#clusterrolebinding-example

sebrandon1: In this case, the name of the cluster role the binding is referencing would be 'secret-reader'.

edcdavid (approved):

pkg/testhelper/testhelper.go
```diff
@@ -106,6 +106,7 @@ const (
 	ReasonForCompliance             = "Reason For Compliance"
 	Category                        = "Category"
 	RoleBindingName                 = "Role Binding Name"
+	ClusterRoleBindingName          = "Cluster Role Reference Name"
```
sebrandon1, line 109: Should the name of the variable be changed as well?

`ClusterRoleBindingName` --> `ClusterRoleName`

aabughosh: yah sure

sebrandon1 (approved):

PR #1455, opened 2023-09-20
config: delete static container image check

jmontesi:
The field `certifiedcontainerinfo` is deleted from the `tnf_config.yml`. It was an old way of checking the certification status of container images before the autodiscovery mechanism was implemented, which we will be the one remaining. The field  `checkDiscoveredContainerCertificationStatus` is also deleted as it will always be active.

As an alternative for checking the certification status of container images we have implemented a new `tnf` command (see [PR#1450](https://github.com/test-network-function/cnf-certification-test/pull/1450))


ramperher:
I suppose CATALOG.md should be updated to remove the affiliated-certification-container-is-certified test reference, right?

jmontesi:
> I suppose CATALOG.md should be updated to remove the affiliated-certification-container-is-certified test reference, right? (Updt) Also, I've seen the DCI job, and it's failing because affiliated-certification-container-is-certified test is failing: https://www.distributed-ci.io/jobs/40a7416f-6288-4ef8-ab67-089a032ba1f3/jobStates, but this test should disappear with this change, right? Only affiliated-certification-container-is-certified-digest test should be executed, from what I understood.

Yes, after talking with the team we have decided to remove the affiliated-certification-container-is-certified test case, as it is covered by affiliated-certification-container-is-certified-digest. I will update the PR with this change.

ramperher (approved):
LGTM

edcdavid (approved):

sebrandon1 (approved):

sebrandon1:
Whoops I merged #1462 which is causing conflicts here.

PR #1461, opened 2023-09-20
Always --pull when building Dockerfile

sebrandon1:
Because the `oct` image is tagged as latest, we should always --pull our base images:

https://docs.docker.com/engine/reference/commandline/pull/

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -161,13 +161,13 @@ delete-db:
 	rm -rf ${REPO_DIR}/offline-db
 
 build-image-local:
-	docker build --no-cache \
+	docker build --pull --no-cache \
 		-t ${REGISTRY_LOCAL}/${TNF_IMAGE_NAME}:${IMAGE_TAG} \
 		-t ${REGISTRY}/${TNF_IMAGE_NAME}:${IMAGE_TAG} \
 		-f Dockerfile .
 
 build-image-tnf:
-	docker build --no-cache \
+	docker build --pull --no-cache \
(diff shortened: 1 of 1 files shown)


rdavid (approved):

jmontesi:
I wonder if the trade-off we are making here is worth it. While this solution makes the image building process more "robust" is introducing a big inefficiency due to the fact that there are three `FROM` instructions in the `Dockerfile`, meaning that those three images will have to be pulled every time the image is rebuilt. Then, I also see that the `--no-cache` option is being used, making the process even slower.

The OCT image does not change much, so before introducing this change I'd reconsider if it's really necessary.

rdavid:
All CI tests function without the use of a cache on a fresh operating system installation. I wouldn't be worried about performance on a local machine.

sebrandon1:
I first noticed this on my QE runner machines.  We were using an `oct` image that was 6 weeks old (roughly the same time I started up the qe runners for the first time).

jmontesi:
> All CI tests function without the use of a cache on a fresh operating system installation. I wouldn't be worried about performance on a local machine.

Yes, the CI workflow starts from scratch, so it will take the same amount of time because it has to pull the images anyway. I was thinking about a local development cycle where you modify one line and have to pull 3 images and go through all the stages in the Dockerfile in order to compile the code. And then you want to do it again...
But it's ok, one can always call `docker build` directly and leave those Makefile tasks for the CI.

PR #1474, opened 2023-09-26
Revert certified digest PR#1464

edcdavid:
This reverts commit aafff629adf4c2ea5945e849eb2b660b3a0cee7a.
The reverted PR only works when a pod has only one container. we still need a way to match containers  in pod spec to containerstatuses in status field. The only way I can see to do this now is to use the container name , which is unique per pod. Reverting until we find another way to match container in both fields.


sebrandon1 (approved):

edcdavid:
closing this PR in favor of https://github.com/test-network-function/cnf-certification-test/pull/1477

edcdavid:
closing this PR in favor of https://github.com/test-network-function/cnf-certification-test/pull/1477

PR #1492, opened 2023-09-29
Improve Dockerfile building speed

sebrandon1:
Compress the number of layers to improve the speed of building the Dockerfile.  Also removing the --no-cache because I'm not sure why it's needed.


sebrandon1:
Closing.  It's still equally as slow on the CI as prior.  I figured it would be a substantial speedup but I guess not.

PR #1500, opened 2023-10-05
Send gchat alert message to dev team if unstable image creation failed.

greyerof:
When a PR is submitted, pre-main.yaml workflow will create the container and push it into quay.io. This change will notify via gchat msg whenever there is an issue with one of the steps that resulted in that new unstable image not being created (hence not pushed) properly.


edcdavid (approved):

sebrandon1 (approved):
In a subsequent PR, we should add this to the tnf-image.yml for the same idea when our release runs fail.

greyerof:
@sebrandon1
> In a subsequent PR, we should add this to the tnf-image.yml for the same idea when our release runs fail.

Yep, good idea.

PR #1502, opened 2023-10-06
Fix for rhcos_versions.sh script.

greyerof:
The -u "unique" param for the sort command will eliminate the duplicated lines in the rhcos_version_map file, hence reducing both the file size and the frequency of the automatic PRs created to update that file.

~I've also uploaded the result file `rhcos_version_map` from running the script locally with this fix.~


rdavid (approved):

greyerof:
Finally decided not to include any changes in the file `rhcos_version_map` and let the workflow to fix it (reduce the duplicated lines) next time it runs. I had to force-push because I did a couple of wrong "git merge main" until I got it right.

sebrandon1 (approved):
Good idea!

PR #1513, opened 2023-10-10
Removed local certdb and use it from OCT.

greyerof:
Both online and offline certificacion status check functions had been moved to OCT's pkg folder.

See https://github.com/test-network-function/oct/pull/69


sebrandon1:
@greyerof Let's make an official `v0.0.1` release in the `oct` repo so we can use that here in the go.mod instead of pointing to a commit.

greyerof:
@sebrandon1 
> @greyerof Let's make an official `v0.0.1` release in the `oct` repo so we can use that here in the go.mod instead of pointing to a commit.
Ok, I'll do that and will update this PR accordingly.

greyerof:
@sebrandon1 , done in https://github.com/test-network-function/cnf-certification-test/pull/1513/commits/bb267abde379526aa1a3cbc5a06297f80c1b22de

sebrandon1 (approved):

PR #1517, opened 2023-10-11
Add container details to error message

rdavid:


cnf-certification-test/networking/suite.go
```diff
@@ -150,7 +150,9 @@ func testExecProbDenyAtCPUPinning(dpdkPods []*provider.Pod) {
 	for _, cpuPinnedPod := range dpdkPods {
 		for _, cut := range cpuPinnedPod.Containers {
 			if cut.HasExecProbes() {
-				ginkgo.Fail("Exec prob is not allowed")
+				ginkgo.Fail(fmt.Sprintf(
```
greyerof, line 153: This is a good moment to fix this test case: we shouldn't call ginkgo.Fail() here, as it aborts the execution of the test case at this point, so the rest containers won't be checked.

The container's name+pod+namespace should  be saved to create a nonCompliantObjects (or compliantObject) list. See the test case code in `testRestartOnRebootLabelOnPodsUsingSriov()` for reference.

PR #1523, opened 2023-10-16
config: add two more options to the CNF config generator

jmontesi:
The options allow adding a list of managed Deployments and StatefulSets.

Also, the Collector option has been hidden from the Create menu as this feature is still in development.

### DIFF
--- a/cmd/tnf/generate/config/config.go
+++ b/cmd/tnf/generate/config/config.go
@@ -86,7 +86,7 @@ func createConfiguration() {
 	createMenu := []configOption{
 		{Option: cnfResources, Help: cnfResourcesHelp},
 		{Option: exceptions, Help: exceptionsdHelp},
-		{Option: collector, Help: collectordHelp},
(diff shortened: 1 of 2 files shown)


cmd/tnf/generate/config/config.go
```diff
@@ -86,7 +86,7 @@ func createConfiguration() {
 	createMenu := []configOption{
 		{Option: cnfResources, Help: cnfResourcesHelp},
 		{Option: exceptions, Help: exceptionsdHelp},
-		{Option: collector, Help: collectordHelp},
+		// {Option: collector, Help: collectordHelp},
```
rdavid, line 89: What's the purpose of retaining commented lines? Git won't forget :-)

rdavid (approved):

cmd/tnf/generate/config/config.go
sebrandon1: They will be uncommented soon when we enable the collector.  I'm fine with keeping them here.

PR #1564, opened 2023-10-27
performance: avoid failing test case if a process cannot be found

jmontesi:
It applies to `rt-apps-no-exec-probes`.

The reason a process cannot be found after it has been properly retrieved previously could be that the process has finished. So, in this case it seems more adequate to just ignore it than to produce an error. If the error is of another kind, the container is added to the non compliant list but the processing for that container continues instead of breaking, to have more info on the processes.

The commit also changes the verification order in that test case, making the faster check of exec probes at the beginning, skipping the slower part in which the processes are list.

Finally, the compliant containers are added to its list with the reason for compliance.

### DIFF
--- a/cnf-certification-test/performance/suite.go
+++ b/cnf-certification-test/performance/suite.go
@@ -233,30 +233,6 @@ func testSchedulingPolicyInCPUPool(env *provider.TestEnvironment,
 	testhelper.AddTestResultReason(compliantContainersPids, nonCompliantContainersPids, tnf.ClaimFilePrintf, ginkgo.Fail)
 }
 
-func getExecProbesCmds(c *provider.Container) map[string]bool {
-	cmds := map[string]bool{}
-
-	if c.LivenessProbe != nil && c.LivenessProbe.Exec != nil {
-		cmd := strings.Join(c.LivenessProbe.Exec.Command, "")
-		cmd = strings.Join(strings.Fields(cmd), "")
-		cmds[cmd] = true
-	}
-
-	if c.ReadinessProbe != nil && c.ReadinessProbe.Exec != nil {
-		cmd := strings.Join(c.ReadinessProbe.Exec.Command, "")
-		cmd = strings.Join(strings.Fields(cmd), "")
-		cmds[cmd] = true
-	}
-
-	if c.StartupProbe != nil && c.StartupProbe.Exec.Command != nil {
-		cmd := strings.Join(c.StartupProbe.Exec.Command, "")
-		cmd = strings.Join(strings.Fields(cmd), "")
-		cmds[cmd] = true
-	}
-
-	return cmds
-}
-
 const noProcessFoundErrMsg = "No such process"
 
 func testRtAppsNoExecProbes(env *provider.TestEnvironment, cuts []*provider.Container) {
@@ -275,16 +251,8 @@ func testRtAppsNoExecProbes(env *provider.TestEnvironment, cuts []*provider.Cont
 			break
 		}
 
-		execProbesCmds := getExecProbesCmds(cut)
 		allProcessesCompliant := true
 		for _, p := range processes {
-			if execProbesCmds[strings.Join(strings.Fields(p.Args), "")] {
-				compliantObjects = append(compliantObjects, testhelper.NewContainerReportObject(cut.Namespace, cut.Podname, cut.Name, "Container process belongs to an exec probe (skipping verification)", true).
-					AddField(testhelper.ProcessID, strconv.Itoa(p.Pid)).
-					AddField(testhelper.ProcessCommandLine, p.Args))
-				continue
-			}
-
 			schedPolicy, _, err := scheduling.GetProcessCPUScheduling(p.Pid, cut)
 			if err != nil {
 				// If the process does not exist anymore it means that it has finished since the time the process list
--- a/cnf-certification-test/performance/suite.go
+++ b/cnf-certification-test/performance/suite.go
@@ -237,21 +237,21 @@ func getExecProbesCmds(c *provider.Container) map[string]bool {
 	cmds := map[string]bool{}
 
 	if c.LivenessProbe != nil && c.LivenessProbe.Exec != nil {
-		for _, cmd := range c.LivenessProbe.Exec.Command {
-			cmds[cmd] = true
-		}
+		cmd := strings.Join(c.LivenessProbe.Exec.Command, "")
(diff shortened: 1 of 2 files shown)


cnf-certification-test/performance/suite.go
```diff
@@ -232,30 +233,50 @@ func testSchedulingPolicyInCPUPool(env *provider.TestEnvironment,
...
 		for _, p := range processes {
 			schedPolicy, _, err := scheduling.GetProcessCPUScheduling(p.Pid, cut)
 			if err != nil {
+				// If the process does not exist anymore it means that it has finished since the time the process list
+				// was retrieved. In this case, just ignore the error and continue processing the rest of the processes.
+				if strings.Contains(err.Error(), noProcessFoundErrMsg) {
+					continue
+				}
 				tnf.ClaimFilePrintf("Could not determine the scheduling policy for container %s (pid=%v), err: %v", cut, p.Pid, err)
 				nonCompliantObjects = append(nonCompliantObjects, testhelper.NewContainerReportObject(cut.Namespace, cut.Podname, cut.Name, "Could not determine the scheduling policy for container", false).
```
edcdavid, line 264: Why not add the ProcessCommandLine as part of the failure reason? It might give more information on whether this is a valid failure case or not.

cnf-certification-test/performance/suite.go
```diff
@@ -232,30 +233,50 @@ func testSchedulingPolicyInCPUPool(env *provider.TestEnvironment,
...
 			break
 		}
+
+		allProcessesCompliant := true
 		for _, p := range processes {
 			schedPolicy, _, err := scheduling.GetProcessCPUScheduling(p.Pid, cut)
 			if err != nil {
+				// If the process does not exist anymore it means that it has finished since the time the process list
+				// was retrieved. In this case, just ignore the error and continue processing the rest of the processes.
+				if strings.Contains(err.Error(), noProcessFoundErrMsg) {
```
edcdavid, line 260: I would still like to see these in the failure reason. In the reason we can state that the process disappeared before we could do the check. This is not a usual case (as far as I understand) and could qualify for an exception in my opinion.

jmontesi: I'm not sure. I don't think we should include the container into the non-compliant category just because by chance we checked one process at the time it was about to disappear. Especially because it is unlikely that this type of processes have a RT policy. Usually RT processes run in busy-loop.

cnf-certification-test/performance/suite.go
jmontesi: Will do.

cnf-certification-test/performance/suite.go
edcdavid: "Usually RT processes run in busy-loop" -> it could be because the process repetitively fails to start or other failure scenarios. For certification, I would rather have more understanding of the issue. Then maybe we could add it in the reason for success? Since this condition will lead to a successful certification and at the same time we can know about it.

edcdavid: We could skip (by matching command line) processes spawned by exec probes from consideration in this test. Looks like they were the source of the short lived processes.

cnf-certification-test/performance/suite.go
jmontesi: Done.

cnf-certification-test/performance/suite.go
jmontesi: Done.

cnf-certification-test/performance/suite.go
```diff
@@ -232,30 +233,86 @@ func testSchedulingPolicyInCPUPool(env *provider.TestEnvironment,
 	testhelper.AddTestResultReason(compliantContainersPids, nonCompliantContainersPids, tnf.ClaimFilePrintf, ginkgo.Fail)
 }
 
+func getExecProbesCmds(c *provider.Container) map[string]bool {
+	cmds := map[string]bool{}
+
+	if c.LivenessProbe != nil && c.LivenessProbe.Exec != nil {
+		for _, cmd := range c.LivenessProbe.Exec.Command {
+			cmds[cmd] = true
```
edcdavid, line 241: Here we are storing each command component in the map, e.g. "cat test.yaml" will result in 2 entries in the map: "cat" and "test.yaml". Later, testing will never match these command bits since they are compared with the full command line from ps. 
Instead, reconstruct the full command line here minus any spaces, carriage return or semi colons and compare it with the output from ps (also stripped from semi colons, spaces, carriage return, etc)

sebrandon1 (approved):

cnf-certification-test/performance/suite.go
```diff
@@ -232,30 +233,86 @@ func testSchedulingPolicyInCPUPool(env *provider.TestEnvironment,
...
 func testRtAppsNoExecProbes(env *provider.TestEnvironment, cuts []*provider.Container) {
 	var compliantObjects []*testhelper.ReportObject
 	var nonCompliantObjects []*testhelper.ReportObject
 	for _, cut := range cuts {
+		if !cut.HasExecProbes() {
+			compliantObjects = append(compliantObjects, testhelper.NewContainerReportObject(cut.Namespace, cut.Podname, cut.Name, "Container does not define exec probes", true))
+			continue
+		}
+
 		processes, err := crclient.GetContainerProcesses(cut, env)
```
edcdavid, line 271: This function returns the full list of processes, including the ones that are spawed by a parent process. For instance in the output below:
`/ # ps -e -o pid,ppid,args
PID   PPID  COMMAND
    1     0 /bin/sh -c touch /tmp/healthy;  sleep 200000
    7     1 sleep 200000
 2779     0 sh
 3001     0 sh -c echo test && sleep 1
 3007  3001 sleep 1
 3008  2779 ps -e -o pid,ppid,args
`
The liveliness command ` 3001     0 sh -c echo test && sleep 1`  also spawns ` 3007  3001 sleep 1`. We will check both commands but the second one (`sleep 1`) will not match, the liveliness command in spec. 
I would skip any commands with a parent PID (ppid) matching a exec probe PID. We should add the ppid in the Process struture

jmontesi: I see, thanks for pointing that out. I think that the consideration of this particular scenario (which is already covered more generally in the original fix) is getting too complex to be addressed in this PR, so I prefer to merge this one to have this test fixed as soon as possible and implement that improvement in another PR.

PR #1576, opened 2023-11-01
Send message to dev team if any QE run fails

sebrandon1:
Similar to: #1503 

If there are failures in any of the QE nightly runs, send a message to the dev team.


rdavid (approved):

greyerof:
@sebrandon1 are our QE jobs really mature and stable enough? Otherwise, we might end up having too many alerts about QE failures that we wouldn't have time/people to attend/fix them quickly. There's already a pending [failing job](https://github.com/test-network-function/cnf-certification-test/actions/runs/6727056904) from the lifecycle ts (7h ago). To be honest, I'm not sure that this chat alert msg is the best idea here.

sebrandon1:
@greyerof I'll close this for now and we can reopen when there's a more stable set of runs.

PR #1590, opened 2023-11-07
Do not consider the exec probes in performance-rt-apps-no-exec-probes

edcdavid:
Reverts commit d7dc6cd1d4d482b1b8e39e4c3789429c5785fd9a and adds logic to whitelist processes belonging to the exec probes as described in comment: https://github.com/test-network-function/cnf-certification-test/pull/1564#discussion_r1376928075

### DIFF
--- a/cnf-certification-test/performance/suite.go
+++ b/cnf-certification-test/performance/suite.go
@@ -248,7 +248,7 @@ func getExecProbesCmds(c *provider.Container) map[string]bool {
 		cmds[cmd] = true
 	}
 
-	if c.StartupProbe != nil && c.StartupProbe.Exec.Command != nil {
+	if c.StartupProbe != nil && c.StartupProbe.Exec != nil {
 		cmd := strings.Join(c.StartupProbe.Exec.Command, "")
 		cmd = strings.Join(strings.Fields(cmd), "")
 		cmds[cmd] = true
--- a/cnf-certification-test/performance/suite.go
+++ b/cnf-certification-test/performance/suite.go
@@ -233,6 +233,30 @@ func testSchedulingPolicyInCPUPool(env *provider.TestEnvironment,
 	testhelper.AddTestResultReason(compliantContainersPids, nonCompliantContainersPids, tnf.ClaimFilePrintf, ginkgo.Fail)
 }
 
+func getExecProbesCmds(c *provider.Container) map[string]bool {
+	cmds := map[string]bool{}
+
+	if c.LivenessProbe != nil && c.LivenessProbe.Exec != nil {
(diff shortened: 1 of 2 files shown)


greyerof (approved):

cnf-certification-test/performance/suite.go
```diff
@@ -233,6 +233,30 @@ func testSchedulingPolicyInCPUPool(env *provider.TestEnvironment,
...
+	}
+
+	if c.StartupProbe != nil && c.StartupProbe.Exec.Command != nil {
+		cmd := strings.Join(c.StartupProbe.Exec.Command, "")
+		cmd = strings.Join(strings.Fields(cmd), "")
+		cmds[cmd] = true
+	}
+
+	return cmds
+}
```
rdavid, line 258:
```suggestion
func getExecProbesCmds(c *provider.Container) map[string]bool {
    cmds := make(map[string]bool)

    probeCommands := []func() []string{
        func() []string { return c.LivenessProbe.Exec.Command },
        func() []string { return c.ReadinessProbe.Exec.Command },
        func() []string { return c.StartupProbe.Exec.Command },
    }

    for _, cmdFunc := range probeCommands {
        if cmd := cmdFunc(); cmd != nil {
            cleanedCmd := strings.Join(cmd, "")
            cleanedCmd = strings.Join(strings.Fields(cleanedCmd), "")
            cmds[cleanedCmd] = true
        }
    }

    return cmds
}
```

rdavid (approved):

cnf-certification-test/performance/suite.go
jmontesi: [Clear is better than clever](https://www.youtube.com/watch?v=PAAkCSZUG1c&t=875s) :)

sebrandon1 (approved):

cnf-certification-test/performance/suite.go
edcdavid: @rdavid let keep it as-is this time. I adapted you idea to match the current logic and I also prefer it as it is now. See:
func getExecProbesCmds(c *provider.Container) map[string]bool {
	cmds := make(map[string]bool)
	probeCommands := []func() *v1.Probe{
		func() *v1.Probe { return c.LivenessProbe },
		func() *v1.Probe { return c.ReadinessProbe },
		func() *v1.Probe { return c.StartupProbe },
	}
	for _, cmdFunc := range probeCommands {
		if probe := cmdFunc(); probe != nil {
			if probe.Exec != nil {
				cleanedCmd := strings.Join(probe.Exec.Command, "")
				cleanedCmd = strings.Join(strings.Fields(cleanedCmd), "")
				cmds[cleanedCmd] = true
			}
		}
	}
	return cmds
}

PR #1593, opened 2023-11-08
Fix regression in  performance-rt-apps-no-exec-probes

edcdavid:
Fix regression from
https://github.com/test-network-function/cnf-certification-test/pull/1590
passing performance QE test:
https://github.com/test-network-function/cnf-certification-test/actions/runs/6802199359
https://github.com/test-network-function/cnf-certification-test/actions/runs/6802987960

### DIFF
--- a/cnf-certification-test/performance/suite.go
+++ b/cnf-certification-test/performance/suite.go
@@ -325,13 +325,18 @@ func filterProbeProcesses(allProcesses []*crclient.Process, cut *provider.Contai
 	}
 	// remove all exec probes and their children from the process list
 	for _, p := range allProcesses {
+		isExecProbeProcess := false
 		for _, parentProbePid := range execProbeProcesses {
 			if p.Pid == parentProbePid || p.PPid == parentProbePid {
-				// skip exec probe processes (child or parent)
-				continue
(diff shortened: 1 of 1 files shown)


sebrandon1 (approved):

cnf-certification-test/performance/suite.go
```diff
@@ -325,13 +325,18 @@ func filterProbeProcesses(allProcesses []*crclient.Process, cut *provider.Contai
 	}
 	// remove all exec probes and their children from the process list
 	for _, p := range allProcesses {
+		isExecProbeProcess := false
 		for _, parentProbePid := range execProbeProcesses {
```
rdavid, line 329: I'd you use Contains:
```
import (
	"fmt"
	"slices"
)

func main() {

	things := []string{"foo", "bar", "baz"}
	a := slices.Contains(things, "foo") // true
	fmt.Printf("%t", a)
}
```

rdavid (approved):

cnf-certification-test/performance/suite.go
edcdavid: nice find! simplifies a lot

PR #1614, opened 2023-11-13
Transition manageability suite to ginkgoless

sebrandon1:
Building upon the work done by @greyerof, this PR will transition the `manageability` suite to what we're calling "ginkgoless".

PR #1615, opened 2023-11-13
Dependency updates for ginkgo removal branch

sebrandon1:
#1610 and #1612


tonyskapunk:
@sebrandon1 I see the DCI check failed, but it has been merged already, do you mind if we test it again just to confirm?

sebrandon1:
Hello @tonyskapunk this wasn't our `main` branch its a proof of concept branch `ginkgo_removal` that we're playing around with.  There are two other PRs #1614 and #1616 that are also failing DCI.  Do you want to retest those?

tonyskapunk:
Ah, that's quite interesting, thanks for pointing it out, not sure if we should be testing changes not against the main branch :thinking:
At first I'd think we shouldn't but wdyt? I assume this branch will later open a PR against main and at that point it will be tested.

sebrandon1:
Correct we're going to try to get everything working in our PoC branch and then make a final PR to merge everything over to main at that time.  Essentially we're trying to get rid of ginkgo as a dependency.

tonyskapunk:
In that case, let me see if we can avoid testing PR not to main, that way we get our of your way and we spend less cycles on something that will be tested later :smile_cat:

sebrandon1:
Sounds good thanks for checking in!

PR #1622, opened 2023-11-15
Transition certification suite to ginkgoless.

greyerof:
Non-backward compatibility: the tc/check
"affiliated-certification-helm-version" will be skipped if no helm
chart releases were discovered. Also, in case we discovered some chart
releases and helm is v3, every chart release will be added as a compliant
report object.

~PR #1620 needs to be merged first, as this code uses the helper function
"testhelper.GetNoContainersUnderTestSkipFn()" created by that PR.~

Also, added the expansion of the "args..." in the ClaimFilePrintf call
to Logf.

### DIFF
--- a/generated_policy.json
+++ b/generated_policy.json
@@ -2,6 +2,11 @@
   "grades": {
     "requiredPassingTests": [
       {
+        "id": "affiliated-certification-operator-is-certified",
+        "suite": "affiliated-certification",
+        "tags": "common"
+      },
+      {
         "id": "manageability-container-port-name-format",
         "suite": "manageability",
         "tags": "extended"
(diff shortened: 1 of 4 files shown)


sebrandon1 (approved):

generated_policy.json
```diff
@@ -36,11 +41,11 @@
         "suite": "performance",
         "tags": "faredge"
       },
-      {
+            {
```
jmontesi, line 44: Block misaligned.

jmontesi (approved):

generated_policy.json
greyerof: Done.

PR #1645, opened 2023-11-20
Allow more than one skip functions per check.

greyerof:
The `check.WithSkipFn(skipFunc)` has been replaced by a variadic func version to allow more than one skip functions to be called: `check.WithSkipFn(skipFunc1, skipFunc2...)`

By default, the check will be skipped if any skip function returns true when they're called (which happens before running the CheckFn).

If we need to skip the check in case all the skip functions return true, the check.WithSkipModeAll() needs to be used.

Examples:
1. Skip check if no deployments ***or*** no statefulsets were found (`testhelper.SkipIfEmptyAny()`):
```
check := checksdb.NewCheck(id, labels).
  WithSkipCheckFn(testhelper.GetNoDeploymentsUnderTestSkipFn(&env), testhelper.GetNoStatefulSetsUnderTestSkipFn(&env)).
  WithCheckFn(func() error {
    ..
  })
```
2. We want to skip the check if no deployments ***and*** no statefulsets were found (`testhelper.SkipIfEmptyAll()`):
```
check := checksdb.NewCheck(id, labels).
  WithSkipCheckFn(testhelper.GetNoDeploymentsUnderTestSkipFn(&env), testhelper.GetNoStatefulSetsUnderTestSkipFn(&env)).
  WithSkipCheckModeAll().
  WithCheckFn(func() error {
    ..
  })
```
The skip functions are appended to a slice in the same order as they're passed to the check.WithSkipCheckFn() calls, and it can be called more than once. E.g.: in the previous examples, this call:
```
  ...
  WithSkipCheckFn(testhelper.GetNoDeploymentsUnderTestSkipFn(&env), testhelper.GetNoStatefulSetsUnderTestSkipFn(&env)).
  ...
```
is equal to:
```
  ...
  WithSkipCheckFn(testhelper.GetNoDeploymentsUnderTestSkipFn(&env)).
  WithSkipCheckFn(testhelper.GetNoStatefulSetsUnderTestSkipFn(&env)).
  ...
```
Extra changes:
- Added recovery mechanism from panic when running the skip functions, similar to the beforeX/afterX ones.
- Added helper skip func creators in the testhelper package for no CRDs, no statefulSets and no deployments found.
- Adapted the observability ts.

### DIFF
--- a/cnf-certification-test/observability/suite.go
+++ b/cnf-certification-test/observability/suite.go
@@ -27,7 +27,6 @@ import (
 	"github.com/test-network-function/cnf-certification-test/cnf-certification-test/identifiers"
 	pdbv1 "github.com/test-network-function/cnf-certification-test/cnf-certification-test/observability/pdb"
 
-	// "github.com/test-network-function/cnf-certification-test/cnf-certification-test/results"
 	"github.com/test-network-function/cnf-certification-test/internal/clientsholder"
 	"github.com/test-network-function/cnf-certification-test/pkg/checksdb"
 	"github.com/test-network-function/cnf-certification-test/pkg/provider"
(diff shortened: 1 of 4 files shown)


sebrandon1 (approved):
Excellent!

PR #1646, opened 2023-11-20
Allow cluster role bindings for pods ultimately owned by a cluster wide operator

edcdavid:
Pods' Owner references are traversed using unstructured generic objects until the top owner are reached and then recorded. If the top owner includes a CSV that is "cluster wide", we allow cluster role bindings to be used by the pods. 
A CSV is clusterwide if:
- the IsClusterWide variable in out operator object is set (checks the target namespaces) 
- the CSV installMethods are either multi-namespace or all-namespace

### DIFF
--- a/pkg/provider/pods.go
+++ b/pkg/provider/pods.go
@@ -30,6 +30,7 @@ import (
 	rbacv1 "k8s.io/api/rbac/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/dynamic"
 )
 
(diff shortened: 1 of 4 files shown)


pkg/provider/pods.go
```diff
@@ -404,3 +405,71 @@ func (p *Pod) IsRunAsUserID(uid int64) bool {
 	}
 	return *p.Pod.Spec.SecurityContext.RunAsUser == uid
 }
+
+// Get the list of top owners of pods
```
sebrandon1, line 409: This would probably be a good candidate for unit testing.

edcdavid: Will add the unit test in a separate PR

PR #1663, opened 2023-11-28
log: add new log package that uses slog as global logger

jmontesi:
The log package provides top-level funtions for creating log records in four different log levels as well as some other helper functions to set up the logging parameters and create special log handlers.

Two slog handlers are created: the `MultiHandler` that allows distributing the logs to several io.Writer objects and the `CustomHandler` to produce the log output in a specifc format.

The `checksdb.Check` type is also modified to add two new fields: a logger and a log archive where the check's log will be stored. The check's logger has the two handlers described above chained, so that the log output is written to the log file and to the check's log archive.

A log file has been created with a default name _cnf-test-log_.

Only a couple of log lines have been produces using this new logger as an example. The bulk of the work for converting the old log lines so that they use the new package will be done in a separate change.

### DIFF
--- a/internal/log/custom_handler.go
+++ b/internal/log/custom_handler.go
@@ -1,0 +1,117 @@
+package log
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"log/slog"
+	"path/filepath"
+	"runtime"
+	"sync"
+	"time"
+)
+
+type CustomHandler struct {
+	opts  Options
+	attrs []slog.Attr
+	mu    *sync.Mutex
+	out   io.Writer
+}
+
+type Options struct {
+	// Level reports the minimum level to log.
+	// Levels with lower levels are discarded.
+	// If nil, the Handler uses [slog.LevelInfo].
+	Level slog.Leveler
+}
+
+func NewCustomHandler(out io.Writer, opts *Options) *CustomHandler {
+	h := &CustomHandler{out: out, mu: &sync.Mutex{}}
+	if opts != nil {
+		h.opts = *opts
+	}
+	if h.opts.Level == nil {
+		h.opts.Level = slog.LevelInfo
+	}
+	return h
+}
+
+func (h *CustomHandler) Enabled(ctx context.Context, level slog.Level) bool {
+	return level >= h.opts.Level.Level()
+}
+
+// The Handle method will write a log line with the following format:
+// LOG_LEVEL [TIME] [SOURCE_FILE] [CUSTOM_ATTRS] MSG
+func (h *CustomHandler) Handle(ctx context.Context, r slog.Record) error {
+	buf := make([]byte, 0, 1024)
+	// Level
+	buf = h.appendAttr(buf, slog.Any(slog.LevelKey, r.Level))
+	// Time
+	if !r.Time.IsZero() {
+		buf = h.appendAttr(buf, slog.Time(slog.TimeKey, r.Time))
+	}
+	// Source
+	if r.PC != 0 {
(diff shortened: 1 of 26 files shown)


pkg/certsuite/certsuite.go
```diff
@@ -14,6 +21,16 @@ import (
...
+func LoadChecksDB() {
+	accesscontrol.LoadChecks()
+	certification.LoadChecks()
+	lifecycle.LoadChecks()
+	manageability.LoadChecks()
+	networking.LoadChecks()
+	observability.LoadChecks()
+	performance.LoadChecks()
+}
+
```
sebrandon1, line 33: @jmontesi Could the `LoadChecks` changes be in a separate PR?  I love the idea of getting rid of those Init() funcs and it doesn't look like its tied to the logging changes.

jmontesi: Sure, here's the [PR](https://github.com/test-network-function/cnf-certification-test/pull/1664).

internal/log/custom_handler.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+func (h *CustomHandler) Enabled(_ context.Context, level slog.Level) bool {
+	return level >= h.opts.Level.Level()
+}
+
+// The Handle method will write a log line with the following format:
+// LOG_LEVEL [TIME] [SOURCE_FILE] [CUSTOM_ATTRS] MSG
+//
+//nolint:gocritic
```
greyerof, line 46: Please add a comment on why this gocritic is needed.

internal/log/custom_handler.go
```diff
@@ -0,0 +1,118 @@
...
+func (h *CustomHandler) Handle(_ context.Context, r slog.Record) error {
+	buf := make([]byte, 0, 1024) //nolint:gomnd
```
greyerof, line 48: Taking into account that we shouldn't be worried about performance/speed, both `var buf []byte` or `buf := []byte{}` options should also work here and they wouldn't need the `nolint`.

main.go
```diff
@@ -50,6 +45,8 @@ const (
 	TNFReportKey                  = "cnf-certification-test"
 	extraInfoKey                  = "testsExtraInfo"
 	noLabelsExpr                  = "none"
+	logFileName                   = "cnf-test-log"
```
greyerof, line 48: I know it's just a sample file name, but I believe it should have an extension. My proposal for a default name is `cnf-certsuite.log`.

greyerof (approved):
Looks good. I just left some suggestions.

internal/log/custom_handler.go
```diff
@@ -0,0 +1,118 @@
...
+	buf = append(buf, "\n"...)
+	h.mu.Lock()
+	defer h.mu.Unlock()
+	_, err := h.out.Write(buf)
+	return err
+}
+
+// Not implemented. WithGroup just returns the same handler.
+func (h *CustomHandler) WithGroup(_ string) slog.Handler {
+	return h
```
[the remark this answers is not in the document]
jmontesi: I don't think it's a good idea to log inside a Handler. I will return nil instead.

internal/log/custom_handler.go
jmontesi: Comment added.

internal/log/custom_handler.go
jmontesi: Done.

main.go
jmontesi: Done.

PR #1665, opened 2023-11-28
Fix for test case operator-install-status-no-privileges.

greyerof:
The original motivation for this test case was checking whether the
operator wasn't creating any SA with RBAC rules related to SCC.

The implementation was wrong as it was (only) checking the resourceNames
list to be empty to pass. But empty list means that any resourceName is
allowed (as per k8s docs).

Since it's impossible to predict all the "privileged" SCCs that might
have been created in the cluster apart from the default ones, this new
implementation is another "canary in the mine" check so we can flag
operators that try to use/create cluster's SCCs.

Operators will need to justify the SCC accesses in their SAs.


cnf-certification-test/operator/suite.go
```diff
@@ -79,11 +79,18 @@ func testOperatorInstallationWithoutPrivileges(env *provider.TestEnvironment) {
...
 		clusterPermissions := csv.Spec.InstallStrategy.StrategySpec.ClusterPermissions
 		if len(clusterPermissions) == 0 {
-			logrus.Debugf("No clusterPermissions found in %s", env.Operators[i])
-			compliantObjects = append(compliantObjects, testhelper.NewOperatorReportObject(env.Operators[i].Namespace, env.Operators[i].Name, "Operator has no privileges on cluster resources", true))
+			logrus.Debugf("No clusterPermissions found in %s", operator)
+			compliantObjects = append(compliantObjects, testhelper.NewOperatorReportObject(operator.Namespace, operator.Name, "Operator has no privileges on cluster resources", true))
+			continue
+		}
+
+		if operator.IsClusterWide {
```
edcdavid, line 91: Shouldn't we add also a check on the all namepace install mode here? For example, with kiali-ossm operator, isClusterWide is false, but installMode only includes All namespaces:
image

greyerof: I've tested it with the kiali-ossm operator and the isClusterWide flag was correctly set to true. ¿Have you included in the tnf_config.yaml the name of the namespace where the subscription of the kiali-ossm operator was created to install it (openshift-operators)?

Apetree100122 (approved):

Apetree100122 (approved):

cnf-certification-test/operator/suite.go
```diff
@@ -84,42 +82,61 @@ func testOperatorInstallationPhaseSucceeded(check *checksdb.Check, env *provider
...
+	for i := range env.Operators {
+		operator := env.Operators[i]
+		csv := operator.Csv
+		check.LogDebug("Checking operator %s", operator)
+		clusterPermissions := csv.Spec.InstallStrategy.StrategySpec.ClusterPermissions
 		if len(clusterPermissions) == 0 {
-			check.LogInfo("Operator %q has no privileged on cluster resources. No clusterPermissions found.", op)
-			compliantObjects = append(compliantObjects, testhelper.NewOperatorReportObject(op.Namespace, op.Name, "Operator has no privileges on cluster resources", true))
+			check.LogInfo("No clusterPermissions found in %s's CSV", operator)
+			compliantObjects = append(compliantObjects, testhelper.NewOperatorReportObject(operator.Namespace, operator.Name, "No RBAC rules for Security Context Constraints found in CSV", true))
```
edcdavid, line 95: Add more details here to distinguish the reasons for success, for instance:
"No RBAC rules for Security Context Constraints found in CSV" -> "No RBAC rules for Security Context Constraints found in CSV (No clusterPermissions found)"

greyerof: Done in https://github.com/test-network-function/cnf-certification-test/pull/1665/commits/2d5c9de23196b0ef7a89fc01ff37cbd209dce1df

sebrandon1 (approved):

PR #1670, opened 2023-11-29
Transition preflight suite to ginkgoless.

greyerof:
The conditions to avoid pre-running the preflight lib's checks have been moved to a helper public function ShouldRun() that is called from certsuite.LoadChecksDB().

### DIFF
--- a/cnf-certification-test/preflight/suite.go
+++ b/cnf-certification-test/preflight/suite.go
@@ -19,44 +19,26 @@ package preflight
 import (
 	"strings"
 
-	"github.com/onsi/ginkgo/v2"
 	plibRuntime "github.com/redhat-openshift-ecosystem/openshift-preflight/certification"
 	"github.com/sirupsen/logrus"
 	"github.com/test-network-function/cnf-certification-test/cnf-certification-test/common"
 	"github.com/test-network-function/cnf-certification-test/cnf-certification-test/identifiers"
+	"github.com/test-network-function/cnf-certification-test/pkg/checksdb"
+	"github.com/test-network-function/cnf-certification-test/pkg/configuration"
 	"github.com/test-network-function/cnf-certification-test/pkg/provider"
 	"github.com/test-network-function/cnf-certification-test/pkg/testhelper"
 	"github.com/test-network-function/cnf-certification-test/pkg/tnf"
 )
 
-var _ = ginkgo.Describe(common.PreflightTestKey, func() {
-	logrus.Debugf("Entering %s suite", common.PreflightTestKey)
-	env := provider.GetTestEnvironment()
-	ginkgo.BeforeEach(func() {
+var (
+	env provider.TestEnvironment
+
+	beforeEachFn = func(check *checksdb.Check) error {
+		logrus.Infof("Check %s: getting test environment.", check.ID)
 		env = provider.GetTestEnvironment()
-	})
-
-	// Add safeguard against running the tests if the docker config does not exist.
-	if env.GetDockerConfigFile() == "" || env.GetDockerConfigFile() == "NA" {
-		logrus.Debug("Skipping the preflight suite because the Docker Config file is not provided.")
-		return
+		return nil
 	}
-
-	// Safeguard against running the preflight tests if the label filter is set but does not include the preflight label
-	ginkgoConfig, _ := ginkgo.GinkgoConfiguration()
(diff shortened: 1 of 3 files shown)


cnf-certification-test/preflight/suite.go
```diff
@@ -19,55 +19,79 @@ package preflight
...
-	})
+		return nil
+	}
+)
 
-	// Add safeguard against running the tests if the docker config does not exist.
-	if env.GetDockerConfigFile() == "" || env.GetDockerConfigFile() == "NA" {
-		logrus.Debug("Skipping the preflight suite because the Docker Config file is not provided.")
-		return
+func labelsAllowTestRun(labelFilter string, allowedLabels []string) bool {
```
sebrandon1, line 43: This could be stringhelper.StringInSlice.

sebrandon1 (approved):
Looks good.  Just one minor comment.

pkg/certsuite/certsuite.go
```diff
@@ -21,14 +22,19 @@ import (
...
 	accesscontrol.LoadChecks()
 	certification.LoadChecks()
 	lifecycle.LoadChecks()
 	manageability.LoadChecks()
 	networking.LoadChecks()
 	observability.LoadChecks()
 	performance.LoadChecks()
+
+	if preflight.ShouldRun(labelsExpr) {
+		logrus.Infof("Loading preflight's lib checks.")
```
jmontesi, line 35: This log looks redundant with the one already present inside `preflight.LoadChecks()`. Also, I'm changing those logs from "Entering [suite] suite" to "Loading [suite] checks", which I think describes better the task.

jmontesi (approved):

cnf-certification-test/preflight/suite.go
sebrandon1: 🤔 They look pretty similar to me unless I'm missing something? 

https://github.com/test-network-function/cnf-certification-test/blob/main/pkg/stringhelper/stringhelper.go#L24C66-L24C66

You can pass the contains boolean that does the `strings.Contains` as well.

cnf-certification-test/preflight/suite.go
```diff
@@ -123,38 +147,42 @@ func generatePreflightContainerGinkgoTest(testName, description, suggestion stri
...
-		// Collect all of the failed and errored containers
-		var failedContainers []string
-		var erroredContainers []string
-		for _, cut := range containers {
-			for _, r := range cut.PreflightResults.Passed {
-				if r.Name() == testName {
-					logrus.Infof("%s has passed preflight test: %s", cut.String(), testName)
+	testID, tags := identifiers.GetGinkgoTestIDAndLabels(aID)
+	check := checksdb.NewCheck(testID, tags).
+		WithSkipCheckFn(testhelper.GetNoContainersUnderTestSkipFn(&env)).
```
edcdavid, line 153: Why not add the docker check here as a check function? This way the skip reason will appear in the claim file for each tests scoped by the label

edcdavid (approved):

cnf-certification-test/preflight/suite.go
greyerof: That'd be great, but It's not that simple with the current code. That was my first approach, but when I tested it, running the preflight ts without the PFTL_DOCKERCONFIG env var set, preflight would fail to download containers of some of the discovered operators (e.g. openshift's ones), [returning error](https://github.com/test-network-function/cnf-certification-test/blob/ae2981eb3e58c0012abe32636796f4de5d4433d8/pkg/provider/containers.go#L127) on that preflight's container check. But that is happening [before](https://github.com/test-network-function/cnf-certification-test/blob/ae2981eb3e58c0012abe32636796f4de5d4433d8/cnf-certification-test/preflight/suite.go#L102) we can create our own check for it.

I'm not even sure that it can be refactored to add that container to the nonCompliant list when that happens, because when preflight's [check.Run()](https://github.com/test-network-function/cnf-certification-test/blob/ae2981eb3e58c0012abe32636796f4de5d4433d8/pkg/provider/containers.go#L123) function returns error, you shouldn't expect the preflight's Result struct to be properly [initialized](https://github.com/redhat-openshift-ecosystem/openshift-preflight/blob/31f7e815872037c910b57e16a50489cb1e566174/container/check_container.go#L42). But we need that struct's fields to create our internal test identifier that is used afterwards to generate our checksdb.Check.

Anyways, I prefer to limit the scope of this PR to the minimum set of required changes to migrate this test suite as it is right now. We can try addressing that refactor in following PRs.

PR #1679, opened 2023-12-01
Remove ginkgo.AbortSuite

sebrandon1:
Created an `abortChan` that the check itself is able to trigger if we need to abort the entire suite.

### DIFF
--- a/cnf-certification-test/lifecycle/podrecreation/podrecreation.go
+++ b/cnf-certification-test/lifecycle/podrecreation/podrecreation.go
@@ -144,7 +144,7 @@ func CordonCleanup(node string, check *checksdb.Check) {
 	err := CordonHelper(node, Uncordon)
 	if err != nil {
 		logrus.Errorf("cleanup: error uncordoning the node: %s, err=%s", node, err)
-		check.SetResultAborted(fmt.Sprintf("cleanup: error uncordoning the node: %s, err=%s", node, err))
+		check.Abort()
 	}
 }
 
--- a/cnf-certification-test/lifecycle/podrecreation/podrecreation.go
+++ b/cnf-certification-test/lifecycle/podrecreation/podrecreation.go
@@ -22,9 +22,9 @@ import (
 	"sync"
 	"time"
 
-	"github.com/onsi/ginkgo/v2"
 	"github.com/sirupsen/logrus"
 	"github.com/test-network-function/cnf-certification-test/internal/clientsholder"
+	"github.com/test-network-function/cnf-certification-test/pkg/checksdb"
 	"github.com/test-network-function/cnf-certification-test/pkg/provider"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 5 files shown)


cnf-certification-test/lifecycle/podrecreation/podrecreation.go
```diff
@@ -140,11 +140,11 @@ func deletePod(pod *corev1.Pod, mode string, wg *sync.WaitGroup) error {
 	return nil
 }
 
-func CordonCleanup(node string) {
+func CordonCleanup(node string, check *checksdb.Check) {
 	err := CordonHelper(node, Uncordon)
 	if err != nil {
 		logrus.Errorf("cleanup: error uncordoning the node: %s, err=%s", node, err)
-		ginkgo.AbortSuite(fmt.Sprintf("cleanup: error uncordoning the node: %s, err=%s", node, err))
+		check.SetResultAborted(fmt.Sprintf("cleanup: error uncordoning the node: %s, err=%s", node, err))
```
edcdavid, line 147: SetResultAborted does not seem to abort skip the subsequent tests. maybe we could propagate the aborted test info to the RunChecks function and use "OnAbort" to skip remaining tests, similar to the timeout and sigint

sebrandon1: In the latest commit, I added an `abortChan` to the `check` struct itself where the struct can abort itself if needed.  In Go channels are by reference by default similar to slices so we should be good creating the abort chan and passing it in to the check.

edcdavid (approved):

cnf-certification-test/lifecycle/suite.go
```diff
@@ -634,7 +634,7 @@ func testPodsRecreation(check *checksdb.Check, env *provider.TestEnvironment) {
 	}
 
 	for nodeName := range podsets.GetAllNodesForAllPodSets(env.Pods) {
-		defer podrecreation.CordonCleanup(nodeName) //nolint:gocritic // The defer in loop is intentional, calling the cleanup function once per node
+		defer podrecreation.CordonCleanup(nodeName, check) //nolint:gocritic // The defer in loop is intentional, calling the cleanup function once per node
```
greyerof, line 637: I'm not sure this is a good idea. The code in this checkFn runs in a separate goroutine. Sending the abort signal won't stop the execution of the rest of deferred funcs, which might take some (unknown time). In parallel, the reader will run the OnAbort() on the remaining test suites and will exit the program execution. If I'm not wrong, this means that the main goroutine might make the program to finish before all the deferred funcs here have finished.

sebrandon1: I merged #1687 before this.

I've been playing around with just adding a `panic()` in the execution and it seems to give a full blown panic when I run the `pods-recreation` test:

image


image

But it seems to gracefully finish out the suite in the logs:

image

What do you think about just adding the panic?  Could be a simple solution?

greyerof: Yeah, panic is (part of) the solution, but we need to detect it and recover from it to avoid showing the stacktrace. I found a way to implement that on top of this change, so I believe you can merge this as it is right now.

greyerof (approved):

PR #1688, opened 2023-12-04
Fix WaitGroup done call

bnshr:


sebrandon1 (approved):

cnf-certification-test/lifecycle/podrecreation/podrecreation.go
```diff
@@ -134,8 +134,8 @@ func deletePod(pod *corev1.Pod, mode string, wg *sync.WaitGroup) error {
 	podName := pod.Name
 	namespace := pod.Namespace
 	go func() {
+		defer wg.Done()
```
edcdavid, line 137: Add reason for the change in description:
Using defer for wg.Done() ensures that Done is called even if the goroutine encounters a panic or if there's an early return.

PR #1691, opened 2023-12-04
Add JUnit XML output

sebrandon1:
Creates a file named `cnf-certification-tests_junit.xml` after the run to match what we created before.

I still have to feature flag this out so it isn't created by default similar to the way preflight handles the XML creation.


greyerof (approved):

script/run-container.sh
```diff
@@ -31,6 +31,7 @@ TNF_OMIT_ARTIFACTS_ZIP_FILE_DEFAULT=false
 TNF_INCLUDE_WEB_FILES_IN_OUTPUT_FOLDER_DEFAULT=false
 ON_DEMAND_DEBUG_PODS_DEFAULT=false
 TNF_ENABLE_DATA_COLLECTION_DEFAULT=false
+TNF_ENABLE_XML_CREATION=false
```
edcdavid, line 34: Shouldn't this be TNF_ENABLE_XML_CREATION_DEFAULT? with a default of true, since the old file was always generated before for DCI

edcdavid (approved):

PR #1684, opened 2023-12-01
Propagate errors from preflight in the claim file 

edcdavid:
Propagates errors from preflight in the claim file  even if preflight fails to run due to connectivity issues, missing docker config, etc...
~~Also contains temporary workaround for leftover ginkgo code in lifecycle, see https://github.com/test-network-function/cnf-certification-test/pull/1679~~

PR #1692, opened 2023-12-05
Avoid loading preflight checks in web server mode.

greyerof:
This is a temporary workaround to allow running the app in web server mode when no kubeconfig file/env var has been provided.

For preflight, the clientsholder needs to be fully and correctly initialized in order to run the checks, otherwise it won't have any TestEnvironment to get the containers/operators under test.

For web server mode, there's no need to have a valid clientsholder yet, as the kubeconfig file is provided via web form, and it can be a different one on each run. The clientsholder is then created with the clientsholder.GetNewClientsHolder().

This workaround should probably be removed/refactored when PR #1684 is merged.

### DIFF
--- a/pkg/certsuite/certsuite.go
+++ b/pkg/certsuite/certsuite.go
@@ -24,6 +24,14 @@ import (
 	"github.com/test-network-function/cnf-certification-test/pkg/provider"
 )
 
+var (
+	webServerMode bool
+)
+
+func SetWebServerMode(serverMode bool) {
+	webServerMode = serverMode
+}
+
 func LoadChecksDB(labelsExpr string) {
 	accesscontrol.LoadChecks()
 	certification.LoadChecks()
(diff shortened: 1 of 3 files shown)


pkg/certsuite/certsuite.go
```diff
@@ -35,7 +43,9 @@ func LoadChecksDB(labelsExpr string) {
 	platform.LoadChecks()
 	operator.LoadChecks()
 
-	if preflight.ShouldRun(labelsExpr) {
+	// Avoid loading preflight checks in webserver mode, since the app may start
+	// without any kubeconfig file provided.
+	if !webServerMode && preflight.ShouldRun(labelsExpr) {
```
edcdavid, line 48: proposed long term solution at: https://github.com/test-network-function/cnf-certification-test/pull/1696

greyerof:
Closing PR in favour of https://github.com/test-network-function/cnf-certification-test/pull/1696

PR #1713, opened 2023-12-11
Non graceful abort for unrecoverable/critical failures inside checkFn.

greyerof:
The check.Abort() has been modified to panic with a message wrapped in a checksdb type. The deferred function in runCheck() will do nothing in case that especial (own) panic is detected, and no other check of that group will run.

This way we stop the goroutine for group.RunChecks() as soon as possible to avoid race conditions with the main goroutine, which will call group.OnAbort() and will register all the remaining checks as skipped, using the reason from the msg channel.

### DIFF
--- a/pkg/checksdb/check.go
+++ b/pkg/checksdb/check.go
@@ -18,10 +18,6 @@ const (
 	CheckResultFailed  = "failed"
 	CheckResultError   = "error"
 	CheckResultAborted = "aborted"
-
-	CheckResultTagPass = "PASS"
-	CheckResultTagSkip = "SKIP"
-	CheckResultTagFail = "FAIL"
 )
 
 type skipMode int
@@ -58,7 +54,7 @@ type Check struct {
 	StartTime, EndTime time.Time
 	Timeout            time.Duration
 	Error              error
-	abortChan          chan bool
+	abortChan          chan string
 }
 
 func NewCheck(id string, labels []string) *Check {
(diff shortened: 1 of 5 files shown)


pkg/checksdb/check.go
```diff
@@ -74,14 +70,17 @@ func NewCheck(id string, labels []string) *Check {
...
-func (check *Check) Abort() {
+func (check *Check) Abort(reason string) {
 	check.mutex.Lock()
 	defer check.mutex.Unlock()
 
-	check.abortChan <- true
+	abortMsg := check.ID + " issued non-graceful abort: " + reason
+
+	check.abortChan <- abortMsg
+	panic(AbortPanicMsg(abortMsg))
```
edcdavid, line 80: Why do we need to panic here?

edcdavid (approved):

pkg/checksdb/check.go
sebrandon1: I believe this is to trigger the [recover()](https://gobyexample.com/recover) func.

edcdavid: right, but what will the recover do in this case?
`			// Don't do anything in case the check was manually aborted by check.Abort().
			if msg, ok := r.(AbortPanicMsg); ok {
				log.Warn("Check was manually aborted, msg: %v", msg)
				err = fmt.Errorf("%v", msg)
				return
			}`
It seems that we are just going back to normal exec mode for a panic we created.
Won't just using the signal work the same?

greyerof: Normal panics from check functions, like index out-of-range runtime panics will make this check to be flagged as errored and the remaining checks of this group will be skipped, but the other group's checks will run normally: that was the original behaviour in ginkgo.

This is just a simple/quick replacement/alternate mechanism for ginkgo's AbortSuite(), so this specific panic recovery avoids running any other check of this group, while relying on the signal receiver to not run any other group's checks. Also, it relies on the signal receiver to call the OnAbort() to flag the current check as aborted and the remaining as skipped.

Anyways, I'll try to improve and simplify the whole mechanism in further PRs.

sebrandon1 (approved):

PR #1725, opened 2023-12-13
re-add logs in claim file

edcdavid:
Depends-On: https://github.com/test-network-function/test-network-function-claim/pull/110

### DIFF
--- a/pkg/checksdb/check.go
+++ b/pkg/checksdb/check.go
@@ -45,7 +45,8 @@ type Check struct {
 
 	Result         CheckResult
 	CapturedOutput string
-	FailureReason  string
+	Details        string
+	SkipReason     string
 
 	logger     *log.Logger
 	logArchive *strings.Builder
@@ -189,7 +190,7 @@ func (check *Check) SetResult(compliantObjects, nonCompliantObjects []*testhelpe
 		check.LogError("Failed to get result objects string for check %s: %v", check.ID, err)
 	}
 
-	check.CapturedOutput = resultObjectsStr
+	check.Details = resultObjectsStr
 
 	// If an error/panic happened before, do not change the result.
 	if check.Result == CheckResultError {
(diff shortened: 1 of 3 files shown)


greyerof (approved):

pkg/checksdb/check.go
```diff
@@ -45,7 +45,8 @@ type Check struct {
 
 	Result         CheckResult
 	CapturedOutput string
-	FailureReason  string
+	Details        string
```
jmontesi, line 48: If not required outside the `checksdb` package the scope of this field should be restricted to its package. The same applies to the `SkipReason` field.

pkg/checksdb/check.go
```diff
@@ -198,11 +199,11 @@ func (check *Check) SetResult(compliantObjects, nonCompliantObjects []*testhelpe
 
 	if len(nonCompliantObjects) > 0 {
 		check.Result = CheckResultFailed
-		check.FailureReason = resultObjectsStr
+		check.SkipReason = resultObjectsStr
```
jmontesi, line 202: It might be better if the `SkipReason` field only contains a simple string explaining why the check has been skipped. The full context for the result objects is already available in the `Details` field.

If it the check has not been skipped then it should be left empty.

pkg/checksdb/check.go
```diff
@@ -216,7 +217,7 @@ func (check *Check) SetResultFailed(reason string) {
 	}
 
 	check.Result = CheckResultFailed
-	check.FailureReason = reason
+	check.SkipReason = reason
```
jmontesi, line 220: If the check failed it means that it has not been skipped, so the `SkipReason` should be empty. The reason for the failure is that there are non-compliant objects that can be found in `Details`.

pkg/checksdb/check.go
edcdavid: fixed

pkg/checksdb/check.go
edcdavid: fixed

sebrandon1 (approved):

pkg/checksdb/check.go
edcdavid: fixed. Removed SetResultFailed function as it is not used (setResult already handles the failed case)

cmd/tnf/claim/show/failures/failures.go
```diff
@@ -253,12 +253,12 @@ func getFailedTestCasesByTestSuite(claimResultsByTestSuite map[string][]*claim.T
 				TestCaseDescription: tc.CatalogInfo.Description,
 			}
 
-			nonCompliantObjects, err := getNonCompliantObjectsFromFailureReason(tc.FailureReason)
+			nonCompliantObjects, err := getNonCompliantObjectsFromFailureReason(tc.SkipReason)
```
jmontesi, line 256: Shouldn't this code use the new field `CheckDetails` instead?

edcdavid: fixed. good catch!

edcdavid: fixed. Good catch!

PR #1720, opened 2023-12-13
to add a msg about the k8s bug

aabughosh:

PR #1727, opened 2023-12-14
add wanning msg for a bug on k8s on ginkgo removal brunch

aabughosh:


sebrandon1:
I don't think its entirely necessary because we will bring in the change from #1720 when we do the git merge eventually.

aabughosh:
merged on main

PR #1745, opened 2023-12-18
CLI output improvements: check's elapsed time and last log trace.

greyerof:
The aim of this change is adding a bit of visual responsiveness to the current cli output of the ginkoless branch when the checks run.

For each running check it will show the elapsed time and the last log line emitted using check.Log() func. Once the check finishes, the cli line will also be rewritten but only the result will be left.

E.g.: while the check is running:
```
[ RUNNING ] lifecycle-pod-recreation (4s)   INFO  [Dec 18 07:50:14.778] [suite.go: 642] [lifecycle-pod-recreation] Draining and Cordoning node kind-worker3:
```
And when it's finished:
```
[ PASS ] lifecycle-pod-recreation
```

Also, if the check was skipped, aborted/errored, the reason will also appear so the user can quickly have an idea of what happened to that check.

E.g:
```
[ SKIP ] performance-max-resources-exec-probes  (Not matching labels)
```

This per-running check's extra info is shown only in case the program runs in a tty/ptty mode (manual run through `cnf-run-suites.sh`).

### DIFF
--- a/internal/cli/cli.go
+++ b/internal/cli/cli.go
@@ -2,6 +2,11 @@ package cli
 
 import (
 	"fmt"
+	"os"
+	"strings"
+	"time"
+
+	"golang.org/x/term"
 )
 
 const (
@@ -14,6 +19,8 @@ const (
 	Cyan   = "\033[36m"
 	Gray   = "\033[37m"
 	White  = "\033[97m"
+
+	ClearLineCode = "\r\033[2K\r"
 )
 
 // ASCII art generated on http://www.patorjk.com/software/taag/ with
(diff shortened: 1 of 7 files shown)


internal/cli/cli.go
```diff
@@ -41,10 +48,100 @@ const (
...
+var CliCheckLogSniffer = &cliCheckLogSniffer{}
+
+var (
+	isTTY bool
+
+	checkLoggerChan chan string
+	stopChan        chan bool
+)
+
+func init() {
```
sebrandon1, line 60: Does this call need to be done as part of an init()?

sebrandon1 (approved):
One comment about init().  Looks good!

edcdavid (approved):

internal/cli/cli.go
greyerof: The alternatives don't look quite good, either, but I'll try to change it in a further PR, since I want to improve the non-tty output.

PR #1751, opened 2023-12-19
Properly use TNF_ENABLE_XML_CREATION

ramperher:
build-depends: https://github.com/redhatci/ansible-collection-redhatci-ocp/pull/100

In `run-container.sh`, it's used `TNF_ENABLE_XML_CREATION` instead of `ENABLE_XML_CREATION` var. Fixing that


greyerof (approved):
Both versions work, but I also prefer the "TNF_"-prefixed one.

ramperher:
> Both versions work, but I also prefer the "TNF_"-prefixed one.

I agree, I've included TNF_ prefix

edcdavid (approved):

sebrandon1 (approved):

PR #1753, opened 2023-12-19
Adjust location of the XML output file

sebrandon1:
We should reference the `flags.OutputDir` here to place the file in the correct place.

PR #1759, opened 2023-12-20
Adjust XML output

sebrandon1:
Follow up to #1753 

Changes:
- Adds `omitempty` to all XML attr's.
- Parse the startTime and endTime stamps correctly and place the duration in the XML.
- Set `FailureMessage` and `SkippedMessage` structs to pointers so they can be nil if not used.

Additional:
- Sort XML output alphabetically.

### DIFF
--- a/pkg/claimhelper/claimhelper.go
+++ b/pkg/claimhelper/claimhelper.go
@@ -21,7 +21,9 @@ import (
 	"encoding/xml"
 	"fmt"
 	"path/filepath"
+	"sort"
 	"strconv"
+	"strings"
 
 	"os"
 	"time"
@@ -41,7 +43,7 @@ const (
 	CNFFeatureValidationJunitXMLFileName = "validation_junit.xml"
 	CNFFeatureValidationReportKey        = "cnf-feature-validation"
 	// dateTimeFormatDirective is the directive used to format date/time according to ISO 8601.
-	DateTimeFormatDirective = "2006-01-02T15:04:05+00:00"
+	DateTimeFormatDirective = "2006-01-02 15:04:05 -0700 MST"
 
 	// States for test cases
 	TestStateFailed  = "failed"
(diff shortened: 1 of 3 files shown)


ramperher (changes requested):
I can see the XML file is now omitting the `skipped` and `failure` labels correctly, but now, all the tnf tests are reported by passed in the DCI UI :sweat_smile: https://www.distributed-ci.io/jobs/f848c42c-50a5-4f5d-8a8c-ce77080cb20e/tests/bb0b560a-7edd-4733-945d-6846c461fc4a

I've been taking a look to the Excel file in https://www.distributed-ci.io/jobs/f848c42c-50a5-4f5d-8a8c-ce77080cb20e/files > cnf-certification-tests_junit.xml, and I could see it's used `Skipped` and `Failure` in upper case. I think they should be in lower case, so `skipped` and `failure`.

DCI is not parsing these elements in fact. I've checked, e.g., a test that is failing, such as `affiliated-certification-container-is-certified-digest`, and the message provided under the `Failure` element is not displayed. I think that, by just changing that to lower case, and also for `Skipped`, it should be fine.

Also, I just wanted to emphasize what I commented yesterday regarding other aspects to check, that are still present:

- Try to sort the tests in a specific way (alphabetically, in the order they're executed, etc.). Before, they were sorted based on the execution order, but now, they appear randomly, and it's a bit difficult to check the results from the same test suite. Having that in alphabetically order could be fine.
- The execution time still looks like artificial, they're mostly 0.00 in all cases, but I believe the real execution time should be there.
- The messages displayed on the skipped tests, they still seem to be less intuitive than the previous format. Is it feasible to return to the previous format?

sebrandon1:
The messages displayed for skipped cases is different than the previous branch and would have to be addressed seperately.  @greyerof maybe we need to add more context there?

greyerof:
@sebrandon1 I'll take a look.

sebrandon1:
@ramperher Please check my latest commit.

Changes:
- Corrected the timestamping (hopefully).  😄 
- Sorted the output alphabetically.
- Adjusted the <Skipped> and <Failure> to be lowercase now.

ramperher (approved):
Now it looks much better, thanks @sebrandon1 ! From my side, it's fine to merge it, the remaining aspect is to check the skipped message, but I think this can be done in a separate change.

greyerof (approved):

PR #1820, opened 2024-01-18
Add CRC testing against Ubuntu

sebrandon1:
Thanks to larger free-tier Github runners, we might be able to fully test our suite using CRC locally installed on these hosted runners.

See: https://github.blog/2024-01-17-github-hosted-runners-double-the-power-for-open-source/ 

CRC requires 9GB of RAM for full-blown OpenShift and the new runners have 14GB of RAM instead of the previous 7GB.


sebrandon1:
Looks like I'm at a standstill with this because of a `libvirt` group problem.  See the following posts:
- https://github.com/actions/runner-images/issues/7670#issuecomment-1900711711
- https://github.com/actions/runner-images/discussions/7191#discussioncomment-8183574

If you try to add the `runner` user to the `libvirt` group in Github Actions, it seems you are forced to reload the terminal for the group changes to take effect which you can't do in Github Actions.  It seems like the only way for this to be "fixed" would be for the original Github Actions image to have the `libvirt` group added.

jiridanek:
@sebrandon1 I've got a workaround for this. It is one of the endless variations of starting a new shell process after adding your user to the new group

```
sudo -s -u $USER bash ./crc.sh
```

after doing that, I got

```
level=info msg="Checking if systemd-networkd is running"
Network configuration with systemd-networkd is not supported. Perhaps you can try this new network mode: https://github.com/crc-org/crc/wiki/VPN-support--with-an--userland-network-stack
```

so I enabled all the configs the help page mentioned, and I got a successful startup

https://github.com/jiridanek/crc-okd-gha/actions/runs/7596756463/job/20690835580

code for that run at https://github.com/jiridanek/crc-okd-gha/tree/492dd1ce36cdaa94cefd84187df6f255469ae08d.

I think this is enough to demonstrate that crc in GHA is workable.

PR #1821, opened 2024-01-18
Remove free disk space cleanup

sebrandon1:
Now that the size of the virtual disks of the free tier is doubled, I'm testing whether or not we need these disk space cleanups anymore to speed up our runs.

See: https://github.blog/2024-01-17-github-hosted-runners-double-the-power-for-open-source/

### DIFF
--- a/.github/workflows/pre-main.yaml
+++ b/.github/workflows/pre-main.yaml
@@ -29,6 +29,7 @@ env:
   CM_BIN: /usr/local/bin/checkmake
   CM_URL_LINUX: https://github.com/mrtazz/checkmake/releases/download/0.2.2/checkmake-0.2.2.linux.amd64 # yamllint disable-line
   SMOKE_TESTS_LABELS_FILTER: '!affiliated-certification-container-is-certified-digest && !access-control-security-context'
+  SKIP_PRELOAD_IMAGES: true
 
 jobs:
   lint:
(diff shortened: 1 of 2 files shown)


.github/workflows/pre-main.yaml
```diff
@@ -158,17 +158,17 @@ jobs:
...
-        with:
-          tool-cache: false
-          large-packages: true
-          android: true
-          dotnet: true
-          haskell: true
-          docker-images: true
-          swap-storage: true
-        continue-on-error: true
+      # - name: Free Disk Space (Ubuntu)
```
rdavid, line 161: We have a Git history; there is no need for commented-out code.

rdavid (approved):

.github/workflows/pre-main.yaml
sebrandon1: Yeah I know I was mostly just testing.  Seems like we're still on the old runners yet:
`/var/lib/containerd/io.containerd.content.v1.content/ingest/e7f5ce25d5889a54f1a5d43ee7f77067a6bdd2544472124edcbca557ac10f40f/data: no space left on device: unknown`

Going to mark WIP and try again tomorrow.

sebrandon1:
Seems to work if I add our env variable: `SKIP_PRELOAD_IMAGES` to skip that section of the `make rebuild-cluster`.  Removing the WIP tag.

PR #1829, opened 2024-01-22
adjust collector's ci test to send claim file to prod collector

shirmoran:


cnf-certification-test/tnf_config.yml
```diff
@@ -36,6 +36,6 @@ servicesignorelist:
   - "new-pro-controller-manager-metrics-service"
   - "mysql"
 collectorAppEndPoint: ""
-executedBy: "default-executed-by"
+executedBy: ""
```
bnshr, line 39: Are these fields necessary to add with empty string?

bnshr (changes requested):

cnf-certification-test/tnf_config.yml
shirmoran: The partners are supposed to fill them.

bnshr: Yes, but this is the config for CI, so these fields can be empty. ExecutedBy is the mandatory, but the others not. So, we do not need those fields to be present in the config for CI test.

shirmoran: Though the tnf_config.yml file is not only for CI use, so won't it be more convenient for them to be there - so the partners just have to fill them in if they want.

shirmoran: and we should probably also add the ec2 endpoint as default for the `collectorAppEndPoint` in the tnf-config

bnshr: Partners dont need to fill ec2 endpoint. The code that parses this tnf_config should fill the collector_url and make request to there based on executedBy.

shirmoran: why based on executedBy? we run the CI testin rds too.. So I guess it's always ec2's endpoint. 
Though I agree it shouldn't be set in the tnf_config .

sebrandon1 (approved):

cnf-certification-test/tnf_config.yml
bnshr: IIRC, this field was mandatory for claim upload, the rest is not.

bnshr: Ah, you are replacing these values. Looks good then!

bnshr (approved):

PR #1837, opened 2024-01-25
Fix cluster bootstrapping step.

greyerof:
That "make make" was making the second step to fail silently, since the commands were separated by ";" instead of "&&."

### DIFF
--- a/.github/workflows/pre-main.yaml
+++ b/.github/workflows/pre-main.yaml
@@ -215,21 +215,21 @@ jobs:
         with:
           timeout_minutes: 90
           max_attempts: 3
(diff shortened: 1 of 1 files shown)


.github/workflows/pre-main.yaml
```diff
@@ -215,7 +215,7 @@ jobs:
         with:
           timeout_minutes: 90
           max_attempts: 3
-          command: cd ${GITHUB_WORKSPACE}/cnf-certification-test-partner; make bootstrap-cluster; make make bootstrap-docker-ubuntu-local; make bootstrap-python-ubuntu-local
+          command: cd ${GITHUB_WORKSPACE}/cnf-certification-test-partner; make bootstrap-cluster && make bootstrap-docker-ubuntu-local && make bootstrap-python-ubuntu-local
```
rdavid, line 218: Always check the return value, including when changing directories.

rdavid (approved):

sebrandon1 (approved):
Good catch!

PR #1843, opened 2024-01-26
Add functionality for running against local CRC instances

sebrandon1:
https://developers.redhat.com/products/openshift-local/overview

If you are running an instance of OpenShift Local (formerly CodeReady Containers) on your local machine and want to run the containerized version of the test suite against it, you have to use `--add-host` for the container to be able reach the apiserver correctly.

### DIFF
--- a/script/run-container.sh
+++ b/script/run-container.sh
@@ -160,6 +160,10 @@ if [ -n "${DNS_ARG}" ]; then
 	DNS_ARG="--dns $DNS_ARG"
 fi
 
+if [ -n "${TNF_ENABLE_CRC_TESTING}" ]; then
+	ADD_HOST_ARG="--add-host api.crc.testing:host-gateway"
+fi
+
 set -x
 # shellcheck disable=SC2068,SC2086 # Double quote array expansions.
 ${TNF_CONTAINER_CLIENT} run --rm $DNS_ARG \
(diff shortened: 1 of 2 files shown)


script/run-container.sh
```diff
@@ -160,6 +160,10 @@ if [ -n "${DNS_ARG}" ]; then
 	DNS_ARG="--dns $DNS_ARG"
 fi
 
+if [ -n "${TNF_ENABLE_CRC_TESTING}" ]; then
+	ADD_HOST_ARG="--add-host api.crc.testing:host-gateway"
+fi
+
```
rdavid, line 166:
```suggestion
if [ -n "${TNF_ENABLE_CRC_TESTING}" ]; then
	ADD_HOST_ARG="--add-host api.crc.testing:host-gateway"
else
	ADD_HOST_ARG=''
fi

```

Otherwise, `$ADD_HOST_ARG` fails when executed with `set -o nounset`, which is recommended. This can be addressed in a separate pull request for all variables.

rdavid, line 166:
```suggestion
ADD_HOST_ARG="${TNF_ENABLE_CRC_TESTING:+--add-host api.crc.testing:host-gateway}"

```

`ADD_HOST_ARG` will be assigned `--add...` value if `TNF_ENABLE_CRC_TESTING` is set, otherwise `''`. With current state `$ADD_HOST_ARG` fails when executed with `set -o nounset`, which is recommended. This can be addressed in a separate pull request for all variables.

rdavid (approved):

PR #1847, opened 2024-01-30
Add filter for tainted node test

sebrandon1:
Add a simple loop to check and make sure we are only testing nodes that have CNF workload deployed on them for the tainted node test.

A scenario this applies to would be if a user was purposefully tainting some nodes.  The debug daemonset would be spawned on those nodes but the CNF workload would not.

### DIFF
--- a/cnf-certification-test/platform/suite.go
+++ b/cnf-certification-test/platform/suite.go
@@ -272,8 +272,22 @@ func testTainted(check *checksdb.Check, env *provider.TestEnvironment) {
 	// Loop through the debug pods that are tied to each node.
 	for _, dp := range env.DebugPods {
 		nodeName := dp.Spec.NodeName
-
 		check.LogInfo("Testing node %q", nodeName)
+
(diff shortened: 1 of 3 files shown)


cnf-certification-test/platform/suite.go
```diff
@@ -272,9 +272,23 @@ func testTainted(check *checksdb.Check, env *provider.TestEnvironment) {
 	// Loop through the debug pods that are tied to each node.
 	for _, dp := range env.DebugPods {
 		nodeName := dp.Spec.NodeName
-
 		check.LogInfo("Testing node %q", nodeName)
 
+		// Ensure we are only testing nodes that have CNF workload deployed on them.
+		nodeHasWorkload := false
```
edcdavid, line 278: Proposal: Move this code to provider->nodes.go. Instead of looping through the debug pods here, loop through the nodes and check if workload is deployed with new function. Then get the debug pod with the env.DebugPods[cut.NodeName] map

edcdavid (approved):

jmontesi (approved):

PR #1849, opened 2024-01-31
Simplify operations involving variables

rdavid:


script/run-container.sh
```diff
@@ -142,27 +142,13 @@ if [[ $LOCAL_DOCKERCFG != NA ]]; then
 	container_tnf_dockercfg_volumes_cmd_args=$(printf -- "-v %s " "${container_tnf_dockercfg_volume_bindings[@]}")
 fi
 
-# Safeguard against an empty variable
-if [ -z "$CONTAINER_TNF_DOCKERCFG" ]; then
-	CONTAINER_TNF_DOCKERCFG=NA
-fi
-
-if [ -n "${LOCAL_TNF_CONFIG}" ]; then
```
edcdavid, line 150: What about this one: LOCAL_TNF_CONFIG

ramperher:
The XML JUnit file is not being generated here, and it should. This is managed with a variable that, somehow, is not taken into account: `TNF_ENABLE_XML_CREATION: true`. Please take care of this and check why the XML JUnit file is not being created

edcdavid (approved):

sebrandon1 (approved):

PR #1853, opened 2024-02-01
Specify network interface for ICMP tests

sebrandon1:
Changes include:
- Changing `MultusIPs: map[string][]string` to `MultusNetworkInterfaces: map[string][]provider.CniNetworkInterface`.
- Making `provider.CniNetworkInterface` struct public.
- Modifying the TestPing() func to accept interface name.

### DIFF
--- a/cnf-certification-test/networking/icmp/icmp.go
+++ b/cnf-certification-test/networking/icmp/icmp.go
@@ -63,10 +63,7 @@ func BuildNetTestContext(pods []*provider.Pod, aIPVersion netcommons.IPVersion, 
 			}
 			for netKey, multusNetworkInterface := range put.MultusNetworkInterfaces {
 				// The first container is used to get the network namespace
-				// Loop through all the network interfaces and their IPs
-				for _, networkInterface := range multusNetworkInterface {
-					processContainerIpsPerNet(put.Containers[0], netKey, networkInterface.IPs, networkInterface.Interface, netsUnderTest, aIPVersion, logger)
-				}
+				processContainerIpsPerNet(put.Containers[0], netKey, multusNetworkInterface.IPs, multusNetworkInterface.Interface, netsUnderTest, aIPVersion, logger)
 			}
 			continue
 		}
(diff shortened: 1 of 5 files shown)


pkg/provider/provider.go
```diff
@@ -462,14 +462,14 @@ func GetPodIPsPerNet(annotation string) (ips map[string][]string, err error) {
 	// Otherwise add all non default interfaces
 	for _, cniInterface := range cniInfo {
 		if !cniInterface.Default {
-			ips[cniInterface.Name] = cniInterface.IPs
+			ips[cniInterface.Name] = append(ips[cniInterface.Name], cniInterface)
```
edcdavid, line 465: Why have a list of CniNetworkInterface here, It shoudl be a CniNetworkInterface. The List of IPs is inside the CniNetworkInterface.

cnf-certification-test/networking/icmp/icmp.go
```diff
@@ -61,17 +61,20 @@ func BuildNetTestContext(pods []*provider.Pod, aIPVersion netcommons.IPVersion,
 				logger.Info("Skipping pod %q because it is excluded from %q connectivity tests only", put.Name, aType)
 				continue
 			}
-			for netKey, multusIPAddress := range put.MultusIPs {
+			for netKey, multusNetworkInterface := range put.MultusNetworkInterfaces {
 				// The first container is used to get the network namespace
-				processContainerIpsPerNet(put.Containers[0], netKey, multusIPAddress, netsUnderTest, aIPVersion, logger)
+				// Loop through all the network interfaces and their IPs
+				for _, networkInterface := range multusNetworkInterface {
+					processContainerIpsPerNet(put.Containers[0], netKey, networkInterface.IPs, networkInterface.Interface, netsUnderTest, aIPVersion, logger)
```
edcdavid, line 68: Same here, we don't need a double loop just get the list of IPs from the single CniNetworkInterface per Net (see previous comment)

pkg/provider/provider.go
sebrandon1: Good call I'll switch it.

cnf-certification-test/networking/icmp/icmp.go
sebrandon1: Done.

pkg/provider/provider.go
sebrandon1: Done.

edcdavid (approved):

PR #1005, opened 2023-04-10
platform-alteration-base-image test case workaround.

greyerof:
The previous implementation of this test cases fails in OCP 4.12.7+ due to podman installed on those OCP worker nodes (v4.2.0) using an older version of the containers/storage lib than the one cri-o uses on those nodes. As a result, "podman diff" won't work because it tries to read a file "containers.json" that has been renamed to
"volatile-containers.json" in latest OCP builds.

This workaround is based on a (compatible) precompiled podman binary existing in a /root/podman folder of the debug pods. Currently, v4.4.0 is the version that can be built with UBI 8 that can read the new "volatile-containers.json" inside worker nodes and fix the "podman diff" function.

In order to avoid copying the binary into the node, I prefer mounting the containing folder into node's /tmp/tnf-podman so the program can be run with /tmp/tnf-podman/podman.

The new implementation of the test case does the following:
1. Creates the temp folder /host/tmp/tnf-podman
2. Mounts debug pod's /root/podman into /host/tmp/tnf-podman
3. Calls /chroot /host /tmp/tnf-podman diff <container-id>
4. Unmounts /host/tmp/tnf-podman
5. Removes /host/tmp/tnf-podman

PR #1861, opened 2024-02-07
Fix for platform-alteration-base-image test case.

greyerof:
The [original workaround](https://github.com/test-network-function/cnf-certification-test/pull/1005) to use a custom podman for this test case stopped working when we switched our debug pod's container image [from ubi8 to ubi9](https://github.com/test-network-function/cnf-certification-test-partner/pull/390) due to libc missing in RHEL8.x based nodes (OCP 4.12.z and lower). This pre-built podman binary was used in this tc no matter the OCP version of the cluster.

This change will make the test case to use our ubi8-based podman only if the test suite is checking CNFs on OCPs 4.12.z or lower. For newer versions of OCP, the podman that's presinstalled in the nodes will be used, as it seems to work well again. I've tested it with SNOs 4.12, 4.13 and 4.14.

We should remove this ugly workaround when 4.12 reaches EOL.

The fix to build podman with ubi8 again is already merged in the partner repo here:
https://github.com/test-network-function/cnf-certification-test-partner/pull/394

Also, I changed some logs to use the check's logger.


ramperher:
check dallas ocp-4.12-vanilla tnf-test-cnf tnf-test-cnf:ansible_extravars=test_network_function_version=HEAD

edcdavid:
Regarding this:
"The original workaround to use a custom podman for this test case stopped working when we switched our debug pod's container image from ubi8 to ubi9 due to libc missing in RHEL8.x based nodes (OCP 4.12.z and lower"
Could we just build podman with ubi9 with a static library to resolve this ? See: https://github.com/containers/podman/blob/6b592bd4e46bc08126ed23b1a09077073eaa80a7/Makefile#L122
Should be able to run in all versions.

greyerof:
> Regarding this: "The original workaround to use a custom podman for this test case stopped working when we switched our debug pod's container image from ubi8 to ubi9 due to libc missing in RHEL8.x based nodes (OCP 4.12.z and lower" Could we just build podman with ubi9 with a static library to resolve this ? See: https://github.com/containers/podman/blob/6b592bd4e46bc08126ed23b1a09077073eaa80a7/Makefile#L122 Should be able to run in all versions.

@edcdavid that might solve the issue of podman not finding the right libc version, but libc versions are usually quite linked to the running kernel/shared libs versions. Running podman using a RHEL9.x (statically linked) libc in a RHEL8.x's kernel and libs which were compiled with a different libc version might cause some weird [compatiblity issues](https://unix.stackexchange.com/a/505877). It doesn't look like a safe move to me, it's not worth the risk. Apart from that, the ideal scenario is not needing any other podman but the one that's already preinstalled in the nodes, which will happen when 4.12 is EOL.

edcdavid (approved):

ramperher (approved):
Test is passing now: https://www.distributed-ci.io/jobs/d44597cb-9694-447e-a6d8-a2faa3e12e2f/tests/b80be7a3-a76f-4858-a547-8ff9666a0599

@greyerof feel free to merge this, thanks!

PR #1872, opened 2024-02-14
workflows: use the tnf check command instead of the gradetool

jmontesi:
The "tnf check results" command verifies that the test results are equal to the expected results found in a template file.

It provides a simpler and more integrated way of verification than the external gradetool.

### DIFF
--- a/.github/workflows/pre-main.yaml
+++ b/.github/workflows/pre-main.yaml
@@ -13,8 +13,6 @@ env:
   TNF_IMAGE_TAG: unstable
   OCT_IMAGE_NAME: testnetworkfunction/oct
   OCT_IMAGE_TAG: latest
-  GRADETOOL_IMAGE_NAME: testnetworkfunction/gradetool
-  GRADETOOL_IMAGE_TAG: latest
   TNF_CONTAINER_CLIENT: docker
   TNF_NON_INTRUSIVE_ONLY: false
   TNF_ALLOW_PREFLIGHT_INSECURE: false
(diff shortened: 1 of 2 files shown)


.github/workflows/pre-main.yaml
```diff
@@ -275,28 +273,11 @@ jobs:
...
-            --volume "${GITHUB_WORKSPACE}"/generated_policy.json:/policy.json \
-            --volume "${GITHUB_WORKSPACE}"/cnf-certification-test/claim.json:/claim.json \
-            ${REGISTRY}/${GRADETOOL_IMAGE_NAME}:${GRADETOOL_IMAGE_TAG} \
-              --OutputPath results.txt \
-              --policy /policy.json \
-              --results /claim.json \
-              >"${GITHUB_WORKSPACE}"/results.txt
-          docker system prune --volumes -f
+      - name: Build the TNF tool
+        run: make build-tnf-tool
```
rdavid, line 277: It can be done in one step: `make build-tnf-tool check-results`

rdavid (approved):

greyerof (approved):

sebrandon1 (approved):

.github/workflows/pre-main.yaml
jmontesi: Thanks for the suggestion but I prefer to leave it as two logical steps in the workflow. This way if it fails it's easier to identify if it was the tool compilation or the results verification. Also, it mirrors better the container job, which does not use the `make` to run the tool.

PR #1896, opened 2024-03-04
autodiscover: modify Istio Service Mesh detection

jmontesi:
The Istio Service Mesh detection required a two step process:
  1) Check for the existence of the "istio-system" namespace.
  2) Check for the presence of a CR from the Istio Operator called
     "installed-state".

But it has been reported that when Istio is installed using the Helm chart method this CR is not created. So, the step 2 above has been replaced for checking the presence of the "istiod" Deployment, which is in charge of running the Istio discovery service.


bnshr:
Looks good. Make sure to create a PR in QE repo to reflect this change, if the QE test gets failed / skipped.

ramperher (approved):
It's behaving as expected, thanks!

PR #1901, opened 2024-03-06
Only return error when an actual error occurs during a CNF run

jmontesi:
Currently, the return code of the `cnf-certification-test` binary is an error code if there is an issue preventing a correct execution or if there are failed test cases. But having failed test cases is just an indication of a workload that does not meet some requirements, not a problem with the CNF Certification itself, so the return code should show that the execution was successful. The results of the execution should always be checked using the information provided in the CLI or in the claim or log file but not using the binary's return code.

This change modifies the return code of the `cnf-certification-test` binary to be successful even if there are some test cases that fail, leaving the error codes for actual runtime errors.

### DIFF
--- a/pkg/certsuite/certsuite.go
+++ b/pkg/certsuite/certsuite.go
@@ -1,7 +1,6 @@
 package certsuite
 
 import (
-	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


sebrandon1 (approved):

sebrandon1:
This change and https://github.com/test-network-function/cnfcert-tests-verification/pull/703 can merge at the same time whenever you are ready @jmontesi

PR #1912, opened 2024-03-12
Add configuration field for collector's endpoint

shirmoran:
Configuration field for collector's endpoint was added to `tnf_config.yaml`.
In addition, prod collector's DNS was set as the default endpoint.

### DIFF
--- a/pkg/certsuite/certsuite.go
+++ b/pkg/certsuite/certsuite.go
@@ -48,7 +48,7 @@ func LoadChecksDB(labelsExpr string) {
 
 const (
 	junitXMLOutputFileName = "cnf-certification-tests_junit.xml"
-	collectorAppEndPoint   = "http://44.195.143.94"
+	collectorAppDNS        = "http://claims-collector.cnf-certifications.sysdeseng.com"
 )
 
 func getK8sClientsConfigFileNames() []string {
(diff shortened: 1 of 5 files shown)


pkg/certsuite/certsuite.go
```diff
@@ -48,7 +48,7 @@ func LoadChecksDB(labelsExpr string) {
 
 const (
 	junitXMLOutputFileName = "cnf-certification-tests_junit.xml"
-	collectorAppEndPoint   = "http://44.195.143.94"
+	collectorAppDNS        = "http://claims-collector.cnf-certifications.sysdeseng.com"
```
rdavid, line 51: DNS stands for Domain Name Service. Did you mean `collectorAppName`?

shirmoran: I meant the DNS, as this URL is using the DNS and not the IP

rdavid: @shirmoran DNS is a service that is run by a server. Perhaps you meant URL?

bnshr: Yeah, `collectorAppUrl` is a better name.

bnshr (approved):

sebrandon1 (approved):

PR #1919, opened 2024-03-14
Avoid unnecessary rebuild of the test environment

jmontesi:
The logs show that the test environment is built twice before starting running the test cases, which should not be necessary. However, disabling this rebuild of the test environment lead to an error while marshaling the `TestConfiguration` to JSON format. This PR refactors the way the Preflight test results are processed in order to produce an output of the test environment adequate for JSON marshaling without the need to rebuild it.

The problem came with the direct addition of the struct `plibRuntime.Results` into our `Container` struct. This former struct contains an interface that prevents the struct to be properly transformed to JSON. To solve this issue, another struct called `PreflightResultsDB` that only contains the information needed has been used instead. 

The test environment rebuild hid this problem because the Preflight results were cleared in the process, allowing the JSON marshaling to work correctly. The results could be retrieved later (even after being cleared) due to the fact that the old results were stored in the closure that contains the test function (defined in `WithCheckFn()`).

### DIFF
--- a/cnf-certification-test/preflight/suite.go
+++ b/cnf-certification-test/preflight/suite.go
@@ -193,7 +193,7 @@ func generatePreflightOperatorCnfCertTest(checksGroup *checksdb.ChecksGroup, tes
 	}, identifiers.TagPreflight)
 
 	checksGroup.Add(checksdb.NewCheck(identifiers.GetTestIDAndLabels(aID)).
-		WithSkipCheckFn(testhelper.GetNoContainersUnderTestSkipFn(&env)).
+		WithSkipCheckFn(testhelper.GetNoOperatorsSkipFn(&env)).
 		WithCheckFn(func(check *checksdb.Check) error {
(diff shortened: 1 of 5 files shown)


ramperher:
> from change #1919:
> 
> * FAILURE https://www.distributed-ci.io/jobs/f4066011-5b95-45e6-9759-92c2aa22ac84/jobStates

The job is failing in the execution of tnf, so I believe this needs to be checked

sebrandon1 (approved):

ramperher (changes requested):
tnf cannot be launched in our deployments, in this job: https://www.distributed-ci.io/jobs/0f0456cb-2996-4615-960f-f237cc09a5cd/files, in the Files section, you can take a look to dci-tnf-execution.log and cnf-certsuite.log to see the errors it is throwing. With this failure, this cannot be merged, else DCI jobs will all fail

edcdavid (approved):

ramperher (approved):

PR #1920, opened 2024-03-14
feat : Check for semantic versioning of the installed application operators 

bnshr:
Refers [CNF Operator Issue 881](https://issues.redhat.com/browse/CNFCERT-881)


cnf-certification-test/operator/util/util.go
```diff
@@ -0,0 +1,11 @@
+package operator
+
+import "regexp"
+
+const semvarRegex = "^(0|[1-9]\\d*)\\.(0|[1-9]\\d*)\\.(0|[1-9]\\d*)(?:-((?:0|[1-9]\\d*|\\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\\.(?:0|[1-9]\\d*|\\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\\+([0-9a-zA-Z-]+(?:\\.[0-9a-zA-Z-]+)*))?$"
+
+func isValidSemanticVersion(version string) bool {
```
rdavid, line 7: Is it better to use a library, e.g.:
 https://github.com/Masterminds/semver

sebrandon1: I agree, if we can _outsource_ it to an outside library lets do it.  We already have masterminds/semver included in this repo already.

bnshr: That's cool. Thanks! I did not know about this library earlier. I will update the code and check if the unit tests pass.

sebrandon1 (approved):

cnf-certification-test/operator/util/util.go
```diff
@@ -0,0 +1,10 @@
+package operator
+
+import (
+	"github.com/Masterminds/semver/v3"
+)
+
+func isValidSemanticVersion(version string) bool {
```
jmontesi, line 7: There is already a function [IsValid()](https://pkg.go.dev/golang.org/x/mod/semver#IsValid) in the `semver` library to check if a version is valid. Also, I don't think we should do unit test of functions that belong to another library, so I suggest closing this PR and using this function wherever is needed.

bnshr: The function that you mentioned is part of golang.org/x/mod/semver package, not in the imported package I used. @sebrandon1 Do you think we shall use the package that Juan mentioned or use the already used one?

bnshr: @jmontesi I ran my unit test against your suggested package, but many test cases failed - this means that some valid semantic versions are not supported by the golang semver. So, in my opinion, using the package from Masterminds is better to use.

cnf-certification-test/operator/suite.go
```diff
@@ -63,6 +63,38 @@ func LoadChecks() {
...
+
+	allOperators := env.AllOperators
+	for _, operator := range allOperators {
+		operatorVersion := operator.Version
+		check.LogInfo("Testing Operator %q for version %s", operator, operatorVersion)
+
+		if provider.IsValidSemanticVersion(operatorVersion) {
+			check.LogInfo("Operator %q has semantic version %s", operator, operatorVersion)
+			compliantObjects = append(compliantObjects, testhelper.NewOperatorReportObject(operator.Namespace, operator.Name,
+				"Operator has semantic version ", true).AddField("version", operatorVersion))
```
greyerof, line 89: `AddField("version")`: Instead of this literal "version" here, maybe you want to reuse the existing field type [Version](https://github.com/test-network-function/cnf-certification-test/blob/d63fb4f2c8da66079f80992ffaac3094312f02b6/pkg/testhelper/testhelper.go#L132).

cnf-certification-test/operator/suite.go
```diff
@@ -63,6 +63,38 @@ func LoadChecks() {
...
+		check.LogInfo("Testing Operator %q for version %s", operator, operatorVersion)
+
+		if provider.IsValidSemanticVersion(operatorVersion) {
+			check.LogInfo("Operator %q has semantic version %s", operator, operatorVersion)
+			compliantObjects = append(compliantObjects, testhelper.NewOperatorReportObject(operator.Namespace, operator.Name,
+				"Operator has semantic version ", true).AddField(testhelper.Version, operatorVersion))
+		} else {
+			check.LogError("Operator %q has invalid semantic versioning %s", operator, operatorVersion)
+			nonCompliantObjects = append(nonCompliantObjects, testhelper.NewOperatorReportObject(operator.Namespace, operator.Name,
+				"Operator has semantic version ", false).AddField(testhelper.Version, operatorVersion))
```
greyerof, line 93: Shouldn't it be "Operator has an **invalid** semantic version"? as the reason for this operator to be non-compliant?

greyerof (approved):

ramperher (changes requested):
If this includes a new test, shouldn't be included in CATALOG.md?

cnf-certification-test/operator/operator.go
```diff
@@ -0,0 +1 @@
+package operator
```
jmontesi, line 1: Should this file be deleted?

bnshr: Yeah, looks like new file. I think I can add some functions inside provider/operators.go

cnf-certification-test/identifiers/identifiers.go
```diff
@@ -908,6 +909,22 @@ tag. (2) It does not have any of the following prefixes: default, openshift-, is
 		},
 		TagCommon)
 
+	TestOperatorHasSemanticVersioningIdentifier = AddCatalogEntry(
+		"installed-operator-semantic-versioning",
```
jmontesi, line 913: I'd just call it "semantic-versioning". This way when prefixed with the suite name "operator" the test case name would simply be "operator-semantic-versioning" instead of "operator-installed-operator-semantic-versioning".

cnf-certification-test/identifiers/identifiers.go
```diff
@@ -908,6 +909,22 @@ tag. (2) It does not have any of the following prefixes: default, openshift-, is
...
+		common.OperatorTestKey,
+		`Tests whether an application Operator has semantic versioning.`,
```
jmontesi, line 915:
```suggestion
		`Tests whether an application Operator's semantic versioning is correct`,
```

cnf-certification-test/identifiers/remediation.go
```diff
@@ -79,6 +79,8 @@ const (
 
 	OperatorIsInstalledViaOLMRemediation = `Ensure that your Operator is installed via OLM.`
 
+	OperatorHasSemanticVersioningRemediation = `Ensure that the installed Operator has semantic versioning.`
```
jmontesi, line 82:
```suggestion
	OperatorHasSemanticVersioningRemediation = `Ensure that the installed Operator's  semantic versioning is correct.`
```

cnf-certification-test/operator/suite.go
```diff
@@ -63,6 +63,38 @@ func LoadChecks() {
...
+	var compliantObjects []*testhelper.ReportObject
+	var nonCompliantObjects []*testhelper.ReportObject
+
+	allOperators := env.AllOperators
+	for _, operator := range allOperators {
+		operatorVersion := operator.Version
+		check.LogInfo("Testing Operator %q for version %s", operator, operatorVersion)
+
+		if provider.IsValidSemanticVersion(operatorVersion) {
+			check.LogInfo("Operator %q has semantic version %s", operator, operatorVersion)
```
jmontesi, line 87:
```suggestion
			check.LogInfo("Operator %q semantic version %q is valid", operator, operatorVersion)
```

cnf-certification-test/operator/suite.go
```diff
@@ -63,6 +63,38 @@ func LoadChecks() {
...
+			compliantObjects = append(compliantObjects, testhelper.NewOperatorReportObject(operator.Namespace, operator.Name,
+				"Operator has semantic version ", true).AddField(testhelper.Version, operatorVersion))
```
jmontesi, line 89:
```suggestion
				"Operator semantic version is valid", true).AddField(testhelper.Version, operatorVersion))
```

pkg/provider/operators.go
```diff
@@ -318,3 +319,8 @@ func getOperatorTargetNamespaces(namespace string) ([]string, error) {
 
 	return list.Items[0].Spec.TargetNamespaces, nil
 }
+
+func IsValidSemanticVersion(version string) bool {
```
jmontesi, line 323: I think this function would fit better into some generic "utils" package as the semantic versioning check applies to any software version, not just to Operators.

bnshr: I would consider that as part of refactoring once we have some more operator related issues implemented.  It's too early to refactor now.

cnf-certification-test/identifiers/identifiers.go
jmontesi: The Operators need to have a label or belong to a namespace defined in the `tnf_config.yml` to be tested, so I think there should be no confusion.

bnshr: No, we are considering the operators here irrespective of labels or namespace.

cnf-certification-test/identifiers/identifiers.go
bnshr: Following this, I propose something else : `Tests whether the operator's version used semantic versioning` https://sdk.operatorframework.io/docs/best-practices/best-practices/

ramperher (approved):
Thanks for addressing my comment!

cnf-certification-test/identifiers/remediation.go
bnshr: Following this, I propose something else : `Ensure that the operator's version used semantic versioning.` 
https://sdk.operatorframework.io/docs/best-practices/best-practices/

cnf-certification-test/identifiers/identifiers.go
jmontesi: But before you said this only applies to non-platform operators. If you don't take into account the labels or the namespace you are testing all operators in the cluster.

PR #1929, opened 2024-03-19
Changed alert msg to be sent to a slack webhook.

greyerof:
Created an action that sends a customizable message along with some static information related to the workflow's job it is running.

It's currently used in conjunction with `if: ${{ failure() }}` to send a message to configured webhook as an alert for a failing job, but can be reused for any other purpose, as long as the webhook endpoint accepts the json object that will be sent to.

The webhook is free to format the final message accordingly but these are the fields it will receive:
```
{
  "attempt": "1",
  "message": "Alert or custom message!",
  "job_url": "https:/github.com/path/to/workflow/job",
  "commit_url": "https://github.com/path/to/commit",
  "repo_url": "https://github.com/path/to/repo"
}
```

Where job_url, commit_url and attempt point to the job where the workflow is/was running, the commit_url of the commit that triggered the workflow and the job attempt number respectively.

Only the message is customizable by means of the step's `with.message` field. The other fields are internally set by the action.

The webhook URL should be stored in a secret and passed to the `with.slack_webook` field of the step.

### DIFF
--- a/.github/actions/slack-webhook-sender/action.yml
+++ b/.github/actions/slack-webhook-sender/action.yml
@@ -1,0 +1,34 @@
+name: slack-webhook-sender
+description: 'Sends a slack message plus some more workflow job related fields to a slack webhook endpoint.'
+inputs:
+  message:
+    description: 'Text that will be send in the json .message field.'
+    required: true
+    default: 'Hello, world!'
+  slack_webhook:
+    description: 'Slack webhook where the data will be posted.'
+    required: true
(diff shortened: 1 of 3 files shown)


sebrandon1 (approved):

rdavid (approved):

rdavid:
@sebrandon1 We have YAML files with both `.yml` and `.yaml` extensions. Shouldn't we standardize to one type? I vote for `.yml`.

PR #1940, opened 2024-03-25
Add --platform flag to Dockerfile

sebrandon1:
https://docs.docker.com/build/building/multi-platform/

For now, we only really need to build the image with the platform `linux/amd64`.  The --platform flag allows the build to pass on M-series Mac's.

EDIT: I had to add hadolint ignore's to the Dockerfile because the linter does not like the --platform flag.


sebrandon1:
So this _works_ but it is extremely slow on my M-series CPU.  I'm going to close this and see if I can build natively and see if that breaks anything.

PR #1942, opened 2024-03-27
Enable multi-platform image builds

sebrandon1:
Modifies the Dockerfile and the workflows to build correctly for each platform, arm64 and amd64.

There are two automation workflows:
- `.github/workflows/pre-main.yaml` (creates the 'unstable' image)
- `.github/workflows/tnf-image.yaml` (creates the tagged releases)

These are both now modified to build images for both arm64 and amd64 images, package them into a manifest, and then push to our registries.

When I tested this locally, you can see I created an image named `bapalm/test-suite:v0.0.1`:
image

It says "See Child Manifests".

If you expand on the child manifests you can see:
image

### DIFF
--- a/.github/workflows/pre-main.yaml
+++ b/.github/workflows/pre-main.yaml
@@ -483,9 +483,31 @@ jobs:
           username: ${{ secrets.QUAY_ROBOT_USERNAME }}
           password: ${{ secrets.QUAY_ROBOT_TOKEN }}
 
-      - name: (if on main and upstream) Push the newly built image to Quay.io
+      - name: Build the `cnf-certification-test` images for multi-platform (x86)
         if: ${{ github.ref == 'refs/heads/main' && github.repository_owner == 'test-network-function' }}
-        run: docker push --all-tags ${REGISTRY}/${TNF_IMAGE_NAME}
+        uses: nick-fields/retry@v3
+        with:
+          timeout_minutes: 90
+          max_attempts: 3
+          command: make build-image-local-x86
+        env:
+          IMAGE_TAG: ${TNF_IMAGE_TAG}
+
+      - name: Build the `cnf-certification-test` images for multi-platform (arm64)
+        if: ${{ github.ref == 'refs/heads/main' && github.repository_owner == 'test-network-function' }}
+        uses: nick-fields/retry@v3
+        with:
+          timeout_minutes: 90
+          max_attempts: 3
+          command: make build-image-local-arm
+        env:
+          IMAGE_TAG: ${TNF_IMAGE_TAG}
+
+      - name: Create and push the manifest list for the `cnf-certification-test` image
+        if: ${{ github.ref == 'refs/heads/main' && github.repository_owner == 'test-network-function' }}
+        run: |
+          make create-manifest-local
+          docker manifest push ${REGISTRY}/${TNF_IMAGE_NAME}:${TNF_IMAGE_TAG}
 
   check-all-dependencies-are-merged:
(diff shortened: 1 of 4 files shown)


sebrandon1:
@ramperher Can you take a look at this DCI job [result](https://www.distributed-ci.io/jobs/cf4c3ce0-a640-4bdc-b0cc-e72b732e0629/jobStates?sort=date&task=73ae6dea-722f-4530-9ed6-22826d9bc0d5)?

```
task path: /usr/share/ansible/collections/ansible_collections/redhatci/ocp/roles/verify_tests/tasks/parse_junit_file.yml:37
fatal: [jumphost]: FAILED! => {"changed": false, "msg": "The following expectations failed: [{'failed_expectation': {'testcase': '^((?!networking-reserved-partner-ports|access-control-security-context|platform-alteration-tainted-node-kernel|affiliated-certification-container-is-certified-digest|preflight-HasUniqueTag|preflight-ScorecardBasicSpecCheck|preflight-RequiredAnnotations|preflight-ScorecardOlmSuiteCheck).)*$', 'passed': True}, 'actual_result': {'testcase': 'affiliated-certification-helmchart-is-certified', 'passed': False}}]"}
```

.github/workflows/pre-main.yaml
```diff
@@ -483,9 +483,31 @@ jobs:
...
+
+      - name: Build the `cnf-certification-test` images for multi-platform (arm64)
+        if: ${{ github.ref == 'refs/heads/main' && github.repository_owner == 'test-network-function' }}
+        uses: nick-fields/retry@v3
+        with:
+          timeout_minutes: 90
+          max_attempts: 3
+          command: make build-image-local-arm
+        env:
+          IMAGE_TAG: ${TNF_IMAGE_TAG}
```
rdavid, line 504: You can combine two commands into one step.

Dockerfile
```diff
@@ -19,17 +19,28 @@ RUN \
...
+	OPERATOR_SDK_ARM_FILENAME=operator-sdk_linux_arm64
+
 # Install Go binary and set the PATH
 ENV \
 	GO_DL_URL=https://golang.org/dl \
-	GO_BIN_TAR=go1.22.1.linux-amd64.tar.gz \
 	GOPATH=/root/go
-ENV GO_BIN_URL_x86_64=${GO_DL_URL}/${GO_BIN_TAR}
+ENV GO_BIN_URL_x86_64=${GO_DL_URL}/go1.22.1.linux-amd64.tar.gz
+ENV GO_BIN_URL_aarch64=${GO_DL_URL}/go1.22.1.linux-arm64.tar.gz
```
rdavid, line 32: Is it possible to install Go from a package manager?

edcdavid (approved):

.github/workflows/pre-main.yaml
sebrandon1: I thought about that but I figured I wanted the granularity to be able to see which step fails if there are any problems.

Dockerfile
sebrandon1: Good question.  If you go the package manager route they _may_ not have the desired package version available whereas we can just download right from the source with the curl.

rdavid: Indeed, do we truly require the desired package version? Using a package manager would result in cleaner code.

rdavid (approved):

Dockerfile
sebrandon1: I would say yes, better to be explicit about versioning and reproducible builds.

PR #1947, opened 2024-04-01
WIP: security requirements of the container-native operators

shimritproj:
According to the **operator best practice requirements** document, we check the recommended security requirements of the container operators:

- USER id should not be 0

- readOnlyRootFilesystem = true

- runAsNonRoot = true

- automount service account token = false{}


shimritproj:
It draft . 
you can see the PR here : https://github.com/test-network-function/cnf-certification-test/pull/1967

PR #1951, opened 2024-04-03
operator: add new test case operator-single-crd-owner

jmontesi:
The test case verifies that every CRD is owned by only one operator.

### DIFF
--- a/CATALOG.md
+++ b/CATALOG.md
@@ -7,7 +7,7 @@ Depending on the CNF type, not all tests are required to pass to satisfy best pr
 
 ## Test cases summary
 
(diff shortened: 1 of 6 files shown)


greyerof (approved):
I wonder whether we should report the conflicting operators list (NewOperatorReportObject) instead.

jmontesi:
> I wonder whether we should report the conflicting operators list (NewOperatorReportObject) instead.

I thought about using an OperatorReportObject instead but as there could potentially be more than one operator owing a CRD but only one CRD I used the CrdReportObject (adding a field with the operator list).

PR #1967, opened 2024-04-08
security requirements of the container-native operators 

shimritproj:
**Issue Link** 
https://issues.redhat.com/browse/CNFCERT-882?filter=-1

This PR includes 4 operator test cases.

- **testOperatorPodsRunAsUserID():** This test verifies that no pods managed by operators run with the root user ID (UID) of 0, which could introduce security vulnerabilities. 

- **testOperatorPodsRunAsNonRoot():** This test ensures that pods managed by operators adhere to security best practices by running as non-root users.

- **testOperatorPodsAutomountTokens():** This test evaluates the configuration of automount service tokens in pods managed by operators.

- **testOperatorContainersReadOnlyFilesystem():** This test verifies whether containers within pods managed by operators have a read-only root filesystem, enhancing security by preventing unauthorized modifications.

We get pods of all operators and then test the above conditions.

In addition, the 'rbac' folder moved into the 'common' folder because it is used by functions in many places in the code, and we want to avoid duplicating code.

build-depends:  https://github.com/dci-labs/dallas-pipelines/pull/1159


cnf-certification-test/operator/suite.go
```diff
@@ -303,3 +312,54 @@ func testOperatorSingleCrdOwner(check *checksdb.Check, env *provider.TestEnviron
...
 	check.SetResult(compliantObjects, nonCompliantObjects)
 }
+
+func testOperatorSecurityRequiremnents(check *checksdb.Check, env *provider.TestEnvironment) {
+	var compliantObjects []*testhelper.ReportObject
+	var nonCompliantObjects []*testhelper.ReportObject
+	for i := range env.Operators {
+		operator := env.Operators[i]
+		check.LogInfo("Testing Operator %q", operator)
+		for _, put := range env.Pods {
```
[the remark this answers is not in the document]
edcdavid: @shimritproj could you update the description with more details on the goal and implementation of this PR?
Based on what @bnshr shared, looks like we need 2 steps:
- recursively identify pods belonging to the operator using owner references and annotations
- check the security context in the pod and in the container for offending configuration. Note that configuration in the container overrides the configuration in the pod if it exists.

ramperher:
@shimritproj could you add this new test to the CATALOG.md? Thanks

pkg/provider/containers.go
```diff
@@ -185,3 +185,8 @@ func (c *Container) HasExecProbes() bool {
 func (c *Container) IsTagEmpty() bool {
 	return c.ContainerImageIdentifier.Tag == ""
 }
+
+func (c *Container) IsReadOnlyRootFilesystem(logger *log.Logger) bool {
```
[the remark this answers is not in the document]
shimritproj: sure, I will add it @sebrandon1

cnf-certification-test/operator/suite.go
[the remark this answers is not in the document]
bnshr: Good point David on SC. Adding a link for Security Context topic. https://medium.com/pareture/k8s-podsecuritycontext-vs-containersecuritycontext-4ca682670f7b

cnf-certification-test/operator/suite.go
```diff
@@ -303,3 +312,54 @@ func testOperatorSingleCrdOwner(check *checksdb.Check, env *provider.TestEnviron
 
 	check.SetResult(compliantObjects, nonCompliantObjects)
 }
+
+func testOperatorSecurityRequiremnents(check *checksdb.Check, env *provider.TestEnvironment) {
```
sebrandon1, line 316: I guess this is more of an organization question, but what do we think about splitting this into 4 separate tests?  Since we are setting compliant/non-compliant objects depending on multiple factors, wouldn't it be more organizational to have completely separate tests for these different test cases?

sebrandon1: For example, instead of `testOperatorSecurityRequiremnents()`, we would have:
- testOperatorRunAsUserID()
- testOperatorRunAsNonRoot()
- testOperatorAutomountTokens()
- testOperatorReadOnlyFilesystem()

shimritproj: I think this is a good idea and it can help in the future. I will split it into 4 functions. @sebrandon1

cnf-certification-test/operator/suite.go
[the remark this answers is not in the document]
shimritproj: I changed the loop to use **env.AllPods**, which contains all the pods I had. From there, I filtered and checked if the kind and name of the operator matched the OwnerReference.

bnshr:
The current PR description is undesirably verbose. 
Please change the PR description to something relevant that describes the changes you made in this PR and optionally how.

cnf-certification-test/operator/suite.go
```diff
@@ -318,3 +349,109 @@ func testOperatorSingleCrdOwner(check *checksdb.Check, env *provider.TestEnviron
...
+					check.LogError("Pod %q container %q is read not only root file system.", put.Name, cut.Name)
+					nonCompliantObjects = append(nonCompliantObjects, testhelper.NewPodReportObject(put.Namespace, put.Name, "Pod has been found read not only root file system", false))
+				}
+			}
+		}
+	}
+	check.SetResult(compliantObjects, nonCompliantObjects)
+}
+
+func findAllPods(env *provider.TestEnvironment, operator *provider.Operator) []*provider.Pod {
```
aabughosh, line 443: i think its better to create a new file for function that are helped function

ramperher:
> from change #1967:
> 
> * SUCCESS https://www.distributed-ci.io/jobs/54276bb8-3f43-4c76-ae9b-108f7d651f53/jobStates

Don't know if it's expected, but the four new tests added in this change are all skipped in this DCI job.

bnshr:
@shimritproj Could you please move the description of the functions as code comments?

cnf-certification-test/operator/suite.go
bnshr, line 443: You can get all Operator Pods at one time, instead of this function. That way you avoid the outer for loop. Is that possible?

cnf-certification-test/operator/suite.go
```diff
@@ -318,3 +349,109 @@ func testOperatorSingleCrdOwner(check *checksdb.Check, env *provider.TestEnviron
...
+		check.LogInfo("Testing Operator %q", operator)
+		AllPods := findAllPods(env, operator)
+		for _, put := range AllPods {
+			check.LogInfo("Testing Pod %q", put)
+			// Evaluate the pod's automount service tokens and any attached service accounts
+			client := clientsholder.GetClientsHolder()
+			podPassed, newMsg := rbac.EvaluateAutomountTokens(client.K8sClient.CoreV1(), put.Pod)
+			if !podPassed {
+				check.LogInfo("Pod %q have automount service tokens set to false", put)
+				compliantObjects = append(compliantObjects, testhelper.NewPodReportObject(put.Namespace, put.Name, "Pod %q have automount service tokens set to false", true))
```
bnshr, line 409: We do not get the information which operator pod could not meet the security requirement. Please consider to add the name of the operator in the result objects. That can help for debugging.

bnshr (changes requested):

cnf-certification-test/operator/suite.go
bnshr: @sebrandon1 I was thinking to have a list of security requirements for any pod (the listed ones in the JIRA) named as getAllPodsWithSecurityRequirements(), instead of having a new test case for each of the requirement. Apply those requirements to any pods (operator, specific application, namespace pods) and then get the test case passed and say that these pods are satisfying the requirements. If we list 5 items (which can change later also if we want to add more restrictive requirements) as security requirements, then any pod that passed all of them are called to satisfy the requirement. Otherwise not. IMO, the requirements for security comes as a block, not separately.

shimritproj:
> > from change #1967:
> > 
> > * SUCCESS https://www.distributed-ci.io/jobs/54276bb8-3f43-4c76-ae9b-108f7d651f53/jobStates
> 
> Don't know if it's expected, but the four new tests added in this change are all skipped in this DCI job. Are they behaving properly?

I am aware of the issue and actively working on it. This is why I indicated that the PR is still a work in progress and not yet ready for review. @ramperher

cnf-certification-test/operator/suite.go
```diff
@@ -318,3 +349,110 @@ func testOperatorSingleCrdOwner(check *checksdb.Check, env *provider.TestEnviron
...
+	}
+	check.SetResult(compliantObjects, nonCompliantObjects)
+}
+
+func findAllPods(env *provider.TestEnvironment, operator *provider.Operator) []*provider.Pod {
+	var AllPods []*provider.Pod
+	for _, put := range env.AllPods {
+		fmt.Printf("pod: %v", put)
+		for _, ownerRef := range put.ObjectMeta.GetOwnerReferences() {
+			for i := range operator.Csv.Spec.CustomResourceDefinitions.Owned {
```
bnshr, line 448: This does not look right to me. Is it tested against some operator? 
In [my PR](https://github.com/test-network-function/cnf-certification-test/pull/2130), we get all custom resource pods owned by the operator. So you do not need this logic. My suggestion is to remove this part and test. You can get all Pods from the field mentioned in the PR description.

bnshr (changes requested):

cnf-certification-test/operator/suite.go
bnshr: Resolved

bnshr: Resolved.

cnf-certification-test/operator/suite.go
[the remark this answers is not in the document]
bnshr: We used `env.AllOperatorPods`.

cnf-certification-test/operator/suite.go
bnshr: Resolved!

expected_results.yaml
```diff
@@ -54,11 +54,15 @@ testCases:
     - observability-crd-status
     - observability-pod-disruption-budget
     - observability-termination-policy
+    - operator-automount-tokens
     - operator-crd-openapi-schema
     - operator-crd-versioning
     - operator-install-source
     - operator-install-status-no-privileges
     - operator-install-status-succeeded
+    - operator-read-only-file-system
```
sebrandon1, line 63: I think this one will have to be moved to the `fail` section but let's see what the smoke tests say.

pkg/provider/pods.go
```diff
@@ -404,6 +404,19 @@ func (p *Pod) IsRunAsUserID(uid int64) bool {
 	return *p.Pod.Spec.SecurityContext.RunAsUser == uid
 }
 
+func (p *Pod) IsRunAsNonRoot() bool {
+	for i := range p.Pod.Spec.Containers {
+		securityContext := p.Pod.Spec.Containers[i].SecurityContext
```
sebrandon1, line 409: This looks better.  Thanks!

manurodriguez:
>  testOperatorContainersReadOnlyFilesystem(): This test verifies whether containers within pods managed by operators have a read-only root filesystem, enhancing security by preventing unauthorized modifications.

I see last [job](https://www.distributed-ci.io/jobs/601a13cb-d032-4311-83e2-fd4454f4c8a4/jobStates?sort=date&task=ad7ccb30-5b0a-453f-8e77-862261fcc93a) failed because the condition above made match:
```
TASK [redhatci.ocp.verify_tests : Fail if not all test results are as expected for cnf-certification-tests_junit.xml]
... 'actual_result': {'testcase': 'operator-read-only-file-system', 'passed': False}}]"}
```
And from the [test](https://www.distributed-ci.io/jobs/601a13cb-d032-4311-83e2-fd4454f4c8a4/tests/fc1d9c99-3300-4f7b-9862-1d978cbab805) details:
```
{"CompliantObjectsOut":null,"NonCompliantObjectsOut":[{"ObjectType":"Pod","ObjectFieldsKeys":["Reason For Non Compliance","Namespace","Pod Name"],"ObjectFieldsValues":["Pod has been found read not only root file system","test-cnf","mongodb-enterprise-operator-65cfdf-72phr"]}]}
```
I guess we need to start taking this into account in our pipeline right? Thanks!

sebrandon1:
@manurodriguez Correct.  We seem to be hitting the same-ish problem with our operator we install as part of our smoke tests where it doesn't pass the tests now.  So it's possible that the mongodb operator used in your tests is also failing for the same reasons.

manurodriguez:
> @manurodriguez Correct. We seem to be hitting the same-ish problem with our operator we install as part of our smoke tests where it doesn't pass the tests now. So it's possible that the mongodb operator used in your tests is also failing for the same reasons.

Thanks for the clarification @sebrandon1, then because we do not control the mongodb operator definitions, I think we can create a PR to ignore the  `operator-read-only-file-system` test failure in the pipeline and only use it as a dependency to test this PR cnf-certification-test#1967

Once the test `operator-read-only-file-system` be part of an official TNF release, then we can decide if ignoring the test can be part of the official fix in our pipeline or think of another solution, wdyt?

sebrandon1:
That's where we're at with the `hazelcast` operator in the smoke tests.  We don't control the definitions so we will have to just handle it as an expected error.

pkg/provider/pods.go
```diff
@@ -404,6 +404,19 @@ func (p *Pod) IsRunAsUserID(uid int64) bool {
...
+		if securityContext == nil || securityContext.RunAsNonRoot == nil {
+			continue
+		}
+		if *securityContext.RunAsNonRoot {
+			return true
+		}
+	}
+	return false
```
greyerof, line 417: I'm not sure if this is ok. As per k8s docs:

image

So if I'm not wrong, this means that some fields can be set in pod's SecurityContext and/or container's SecurityContext, where container's ones take precedence. So if "RunAsNonRoot" was not found in the container, we should also check whether that field exists in pod's SecurityContext.

pkg/provider/pods.go
```diff
@@ -404,6 +404,19 @@ func (p *Pod) IsRunAsUserID(uid int64) bool {
...
 }
 
+func (p *Pod) IsRunAsNonRoot() bool {
+	for i := range p.Pod.Spec.Containers {
+		securityContext := p.Pod.Spec.Containers[i].SecurityContext
+		if securityContext == nil || securityContext.RunAsNonRoot == nil {
+			continue
+		}
+		if *securityContext.RunAsNonRoot {
+			return true
```
greyerof, line 414: You shouldn't return true here, as there might be other containers in the pod that need to be checked...

shimritproj: good point! I will change it.

pkg/provider/pods.go
sebrandon1: Yes good call.  What we were seeing in the hazelcast operator was set on the container but yes we should check both properties at the pod-level and the container-level.

shimritproj: good point! i will add it.

pkg/provider/pods.go
```diff
@@ -404,6 +404,29 @@ func (p *Pod) IsRunAsUserID(uid int64) bool {
...
+		if p.Pod.Spec.Containers[i].SecurityContext != nil || p.Pod.Spec.Containers[i].SecurityContext.RunAsNonRoot != nil {
+			if !*p.Pod.Spec.Containers[i].SecurityContext.RunAsNonRoot {
+				flag = false
+				break
+			}
+		}
+	}
+	if len(p.Pod.Spec.Containers) != 0 && flag {
+		return true
+	}
```
sebrandon1, line 420:
```suggestion
	for i := range p.Pod.Spec.Containers {
		if p.Pod.Spec.Containers[i].SecurityContext != nil || p.Pod.Spec.Containers[i].SecurityContext.RunAsNonRoot != nil {
			if !*p.Pod.Spec.Containers[i].SecurityContext.RunAsNonRoot {
				return false
			}
		}
	}
```

sebrandon1 (commented):
Comment about the loop

ramperher:
@shimritproj , if the change is WIP, could you add the `dci-disable` label to avoid launching DCI jobs? (which are mainly failing right now)

pkg/provider/pods.go
```diff
@@ -404,6 +405,16 @@ func (p *Pod) IsRunAsUserID(uid int64) bool {
 	return *p.Pod.Spec.SecurityContext.RunAsUser == uid
 }
 
+func (p *Pod) IsRunAsNonRoot() bool {
+	// Check pod-level security context
+	if p.Pod.Spec.SecurityContext != nil || p.Pod.Spec.SecurityContext.RunAsNonRoot != nil {
```
aabughosh, line 410: this cpndtion need to be with and ant not or

cnf-certification-test/operator/suite.go
```diff
@@ -317,3 +347,100 @@ func testOperatorSingleCrdOwner(check *checksdb.Check, env *provider.TestEnviron
 
 	check.SetResult(compliantObjects, nonCompliantObjects)
 }
+
+func testOperatorPodsRunAsUserID(check *checksdb.Check, env *provider.TestEnvironment) {
+	var compliantObjects []*testhelper.ReportObject
+	var nonCompliantObjects []*testhelper.ReportObject
+
+	for csv, pods := range env.CSVToPodListMap {
+		check.LogInfo("Csv: %q", csv)
```
bnshr, line 356: csv is not the csv name. Rather https://github.com/test-network-function/cnf-certification-test/blob/7d772bb7c4e96795c85a50998e41a148a4c33b9e/pkg/autodiscover/autodiscover.go#L315

Print the name and namespace of the CSV by string split operation. There is one constant created for this.

cnf-certification-test/operator/suite.go
```diff
@@ -317,3 +347,100 @@ func testOperatorSingleCrdOwner(check *checksdb.Check, env *provider.TestEnviron
...
+		for _, pod := range pods {
+			check.LogInfo("Testing Pod %q", pod)
+			if pod.IsRunAsUserID(0) {
+				check.LogError("Pod %q UserID of the pods created by the operator is 0", pod.Name)
```
bnshr, line 360: Better alternative is : 
`check.LogError("Pod %s in namespace %s is running with UserID = 0", pod.Name, pod.Namespace)`

bnshr (changes requested):
Make changes in all logs wherever applicable. I did not copy paste the same suggestions everywhere.

cnf-certification-test/operator/suite.go
shimritproj: i agree , i will fix all the logs.

cnf-certification-test/operator/suite.go
```diff
@@ -317,3 +347,100 @@ func testOperatorSingleCrdOwner(check *checksdb.Check, env *provider.TestEnviron
...
+		for _, pod := range pods {
+			// We are looking through both the containers and the pods separately to make compliant and non-compliant objects.
+
+			for _, c := range pod.Containers {
+				if c.IsContainerRunAsNonRoot() {
+					check.LogInfo("Container %q created by the operator is run as not root", c.Name)
+					compliantObjects = append(compliantObjects, testhelper.NewPodReportObject(c.Namespace, c.Name, "Container has been found is run as not root", true))
+				} else {
+					check.LogError("Pod %q created by the operator is run as root", pod.Name)
+					nonCompliantObjects = append(nonCompliantObjects, testhelper.NewPodReportObject(pod.Namespace, pod.Name, "Container has been found is run as root", false))
```
sebrandon1, line 386:
```suggestion
					check.LogError("Container %q created by the operator is run as root", pod.Name)
					nonCompliantObjects = append(nonCompliantObjects, testhelper.NewPodReportObject(pod.Namespace, pod.Name, "Container has been found is run as root", false))
```

shimritproj: I changed all the messages, it is an old version.

ramperher (approved):
Reviewed the issue with DCI job, it's because of the workload we're deploying, tests look fine.

sebrandon1:
@ramperher Does this mean that if we rekick the DCI job the error is handled now as an expected failure?

ramperher:
> @ramperher Does this mean that if we rekick the DCI job the error is handled now as an expected failure?

Exactly, we're running an operator that is not under our control, so I assume the tests that were failing before must be considered as exceptions.

sebrandon1 (approved):

PR #2002, opened 2024-04-23
Update oct to v0.0.9

sebrandon1:
`oct` had a bug that was not populating the offline db correctly and was not allowing for online database checks either.

See: https://github.com/test-network-function/oct/pull/104

v0.0.9 contains the fix to allow for the onlinecheck to actually work again.


sebrandon1:
Closing this in favor of v0.0.10

PR #2009, opened 2024-04-25
Update preflight to v1.9.2

sebrandon1:
https://github.com/redhat-openshift-ecosystem/openshift-preflight/releases/tag/1.9.2


sebrandon1:
Doesn't seem like this is possible at this point in time.  There is an underlying dependency that isn't allowing the upgrade.

PR #2033, opened 2024-05-06
scripts: use the test suite binary directly in "run-tnf-container.sh"

jmontesi:
This avoids the dependency of the script "run-cnf-suites.sh".


run-tnf-container.sh
```diff
@@ -25,9 +25,9 @@ export TNF_IMAGE_NAME=cnf-certification-test
 export TNF_IMAGE_TAG=latest
 export TNF_OFFICIAL_ORG=quay.io/testnetworkfunction/
 export TNF_OFFICIAL_IMAGE="${TNF_OFFICIAL_ORG}${TNF_IMAGE_NAME}:${TNF_IMAGE_TAG}"
-export TNF_CMD="./run-cnf-suites.sh"
+export TNF_BIN_DIR="cnf-certification-test"
+export TNF_CMD="cnf-certification-test"
```
rdavid, line 29: There is no need for quotation marks.

script/run-container.sh
```diff
@@ -32,6 +32,13 @@ TNF_INCLUDE_WEB_FILES_IN_OUTPUT_FOLDER_DEFAULT=false
 ON_DEMAND_DEBUG_PODS_DEFAULT=false
 TNF_ENABLE_DATA_COLLECTION_DEFAULT=false
 TNF_ENABLE_XML_CREATION_DEFAULT=false
+TIMEOUT_ARG=--timeout=${TIMEOUT:-24h0m0s}
+OUTPUT_LOC_ARG=--output-dir=$CONTAINER_TNF_DIR
+
+if [[ $TNF_LABEL == "all" ]]; then
```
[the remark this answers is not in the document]
jmontesi: It would be very strange but it could. In that case no test would run.

run-tnf-container.sh
jmontesi: Done.

script/run-container.sh
```diff
@@ -32,6 +32,13 @@ TNF_INCLUDE_WEB_FILES_IN_OUTPUT_FOLDER_DEFAULT=false
 ON_DEMAND_DEBUG_PODS_DEFAULT=false
 TNF_ENABLE_DATA_COLLECTION_DEFAULT=false
 TNF_ENABLE_XML_CREATION_DEFAULT=false
+TIMEOUT_ARG=--timeout=${TIMEOUT:-24h0m0s}
+OUTPUT_LOC_ARG=--output-dir=$CONTAINER_TNF_DIR
```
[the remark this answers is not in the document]
jmontesi: Done.

script/run-container.sh
```diff
@@ -32,6 +32,13 @@ TNF_INCLUDE_WEB_FILES_IN_OUTPUT_FOLDER_DEFAULT=false
 ON_DEMAND_DEBUG_PODS_DEFAULT=false
 TNF_ENABLE_DATA_COLLECTION_DEFAULT=false
 TNF_ENABLE_XML_CREATION_DEFAULT=false
+TIMEOUT_ARG=--timeout=${TIMEOUT:-24h0m0s}
```
edcdavid, line 35: Ideally I believe that we should not need any scripts here. We should move support for offline db inside the CNFCERT container itself and just provide an example on how to run the CNFCERT container in the documentation using podman.

jmontesi: I agree. I'm proceeding step by step here though. Trying first to decouple the scrips from one another, then prepare the binary to run in a standalone way efficiently and then deleting the scripts.

edcdavid (approved):

PR #2065, opened 2024-05-14
cmd/tnf: add subcommand to run the test suite

jmontesi:
Use `./tnf run <flags>` to run the test suite. See `./tnf run -h `for more info.

The original way of running the test suite using the "cnf-certification-test" binary and the bash scripts is still available and used in the GitHub CI. This change just add the possibility of running the test suite with the "tnf" command.

### DIFF
--- a/pkg/certsuite/certsuite.go
+++ b/pkg/certsuite/certsuite.go
@@ -101,8 +101,15 @@ func processFlags() {
 	}
 }
 
+func isTnfSubcommand() bool {
+	cmd := strings.Split(os.Args[0], "/")
+	return cmd[len(cmd)-1] == "tnf"
+}
+
 func Startup() {
-	flags.InitFlags()
+	if !isTnfSubcommand() {
+		flags.InitFlags()
+	}
 
 	err := configuration.LoadEnvironmentVariables()
 	if err != nil {
--- a/pkg/certsuite/certsuite.go
+++ b/pkg/certsuite/certsuite.go
@@ -101,13 +101,8 @@ func processFlags() {
 	}
 }
 
-func isTnfSubcommand() bool {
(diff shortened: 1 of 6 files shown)


pkg/certsuite/certsuite.go
```diff
@@ -101,8 +101,15 @@ func processFlags() {
 	}
 }
 
+func isTnfSubcommand() bool {
+	cmd := strings.Split(os.Args[0], "/")
+	return cmd[len(cmd)-1] == "tnf"
+}
+
 func Startup() {
```
edcdavid, line 109: Why not a parameter "initFlags bool" to startup to trigger initialization of flags? This way we don't need the isTnfSubcommand function

pkg/provider/provider.go
```diff
@@ -224,6 +225,9 @@ func buildTestEnvironment() { //nolint:funlen
 	env = TestEnvironment{}
 
 	env.variables = *configuration.GetTestParameters()
+	if flags.ConfigurationFile != "" {
```
edcdavid, line 228: If the goal is to give precedence to flags, why not load environement variables first, then we won't need this code.

jmontesi: We could do that but in that case the new "cobra" flags would have to be loaded inside the "certsuite" package or the code to load the env vars moved outside of it. As we agreed to have only flags and no env vars consider this code temporary until the env vars are not used. At the moment it is required for compatibility between both modes of running the test suite.

pkg/certsuite/certsuite.go
jmontesi: Agree.

edcdavid (approved):

PR #2067, opened 2024-05-14
Add MacOS arm64 smoke test

sebrandon1:
Adding smoke test for arm64 images.


sebrandon1:
Closed because MacOS virtualization isn't supported on M1 processors yet.

PR #2092, opened 2024-05-21
Parse flags after env vars to override them if required

jmontesi:
To support the transition from the standard test suite binary to the test suite run as a "certsuite run" subcommand the precedence of flags needs to be higher than that of the env vars, as flags will be the main mechanism for passing options to the test suite.

As no new flags will be added to the old flag library, it's necessary to leave the env vars but override them when the test suite it run with the "certsuite" command.

### DIFF
--- a/cnf-certification-test/results/html/results.html
+++ b/cnf-certification-test/results/html/results.html
@@ -1,1 +1,230 @@
-This is a placeholder for the results.html file that will be downloaded from https://github.com/test-network-function/parser when this project is built
+<!DOCTYPE html>
+<html lang="en">
+
+<head>
+  <title>CNF Certification Test Parser</title>
+  <meta charset="UTF-8">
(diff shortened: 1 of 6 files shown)


cnf-certification-test/results/html/results.html
```diff
@@ -1 +1,230 @@
-This is a placeholder for the results.html file that will be downloaded from https://github.com/test-network-function/parser when this project is built
+<!DOCTYPE html>
```
sebrandon1, line 1: I think this was mistakenly committed.

sebrandon1 (commented):
One comment about the results.html

cnf-certification-test/results/html/results.html
jmontesi: Right. It's fixed now.

greyerof (approved):

PR #2094, opened 2024-05-21
Test 'use_binary' branch on macos runner

sebrandon1:


sebrandon1:
```
level=info msg="Checking minimum RAM requirements"
crc requires at least 11.27GB to run
```

PR #2095, opened 2024-05-21
Test 'use_binary' branch

sebrandon1:


sebrandon1:
Closing this, as it was a good experiment but we don't necessarily need automated testing of the binary just yet.  Maybe when arm-based runners are available we can re-open.

PR #2098, opened 2024-05-23
cmd/certsuite: add new flags

jmontesi:


sebrandon1 (approved):
Does this change require updates to the DCI team as well since they use some of the flags like the XML creation?

ramperher:
> Does this change require updates to the DCI team as well since they use some of the flags like the XML creation?

This change only affect to the cmd, is it correct? Then, I believe nothing needs to be changed in our side because we always use the tnf container instead of the cmd.

jmontesi:
> > Does this change require updates to the DCI team as well since they use some of the flags like the XML creation?
> 
> This change only affect to the cmd, is it correct? Then, I believe nothing needs to be changed in our side because we always use the tnf container instead of the cmd.

Correct. The changes I've done so far do not require current users to update anything.

greyerof (approved):

PR #2107, opened 2024-05-28
Batch operators test catalog source automation

shirmoran:
Instead of getting as an input a catalog name, the script will get a catalog index and will create a "CatalogSource" resource based on it.  At the end of the run the "CatalogSource" will be deleted.

* In addition the "status" variable was removed - the exit code is checked after every command and an error will appear if necessary.

### DIFF
--- a/run-basic-batch-operators-test.sh
+++ b/run-basic-batch-operators-test.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-set -o errexit -o nounset -o pipefail
+set -o nounset -o pipefail
 
 # Test run timestamp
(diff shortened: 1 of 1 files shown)


tonyskapunk:
check workload

tonyskapunk:
check workload tnf-test-cnf-green

jmontesi:
In my view, this bash script is reaching the point where I would consider translating it to another language, such as Python o even Go. What do you think?

rdavid (approved):

sebrandon1 (approved):

edcdavid (approved):
Great work Thx!

PR #2109, opened 2024-05-28
Bug fix: adjust OCP beta slice

sebrandon1:
The `ocpBetaVersions` slice was only containing entries for 4.13 and 4.14.  I added more entries for 4.15, 4.16, and 4.17 and I also added a unit test for the version that @ramperher reported as failing.

Bonus, I changed the strings that report "master" nodes to "control plane" nodes.


ramperher (approved):
Looks nice! We will be able to test it probably next weekend where we'll launch OCP 4.16 rc release, but in the meantime I think it's fine to merge this.

PR #2112, opened 2024-05-29
Test: Revert changes to ubuntu 24.04 for cancelled jobs

sebrandon1:


sebrandon1:
Seems like others are reporting early cancellations in here: https://github.com/actions/runner-images/issues/9848

rdavid (approved):

PR #2130, opened 2024-06-03
feat : Add helper function to get all operator pods

bnshr:
**Issue** [JIRA Link](https://issues.redhat.com/browse/CNFCERT-885)

**Includes**
- Logic to get all pods managed by the operators under test from the namespaces mentioned through `targetNamespaces` in CSV. The client code gets this from `env.AllOperatorPods` 
- Refactoring of the retrieval of owner references code (moved from `provider` to `podhelper`

**NOTE** 
To be used in [this PR](https://github.com/test-network-function/cnf-certification-test/pull/1967)

### DIFF
--- a/pkg/autodiscover/autodiscover.go
+++ b/pkg/autodiscover/autodiscover.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"regexp"
+	"strings"
 	"time"
 
 	configv1 "github.com/openshift/api/config/v1"
@@ -56,6 +57,7 @@ type DiscoveredTestData struct {
 	Pods                   []corev1.Pod
 	AllPods                []corev1.Pod
 	DebugPods              []corev1.Pod
+	AllOperatorPods        []corev1.Pod
 	ResourceQuotaItems     []corev1.ResourceQuota
 	PodDisruptionBudgets   []policyv1.PodDisruptionBudget
 	NetworkPolicies        []networkingv1.NetworkPolicy
@@ -182,6 +184,14 @@ func DoAutoDiscover(config *configuration.TestConfiguration) DiscoveredTestData 
 	data.Subscriptions = findSubscriptions(oc.OlmClient, data.Namespaces)
 	data.HelmChartReleases = getHelmList(oc.RestConfig, data.Namespaces)
 
+	// Get all operator pods
+	var allOperatorPods []corev1.Pod
+	for _, csv := range data.Csvs {
+		pods := getOperatorCsvPods(csv)
+		allOperatorPods = append(allOperatorPods, pods...)
+	}
+	data.AllOperatorPods = allOperatorPods
+
 	openshiftVersion, err := getOpenshiftVersion(oc.OcpClient)
 	if err != nil {
 		log.Fatal("Failed to get the OpenShift version, err: %v", err)
@@ -279,4 +289,44 @@ func getOpenshiftVersion(oClient clientconfigv1.ConfigV1Interface) (ver string, 
 	}
 
 	return "", errors.New("could not get openshift version from clusterOperator")
+}
+
+// Get the list of pods mananged by the operator - retrieve from the CSV and then check the operator group to find the target namespaces to retrieve the managed pods
+func getOperatorCsvPods(csv *olmv1Alpha.ClusterServiceVersion) []corev1.Pod {
+	annotations := csv.Annotations
+
+	targetNamespaces := annotations["olm.targetNamespaces"]          // This is a comma-separated string, example : a,b,c where a, b and c are target namespaces
+	operatorTargetNamespaces := strings.Split(targetNamespaces, ",") // For cluster installed operator olm.targetNamespaces: "", so handle differently
+
+	var podList []corev1.Pod
+
+	for _, targetNamespace := range operatorTargetNamespaces {
+		pods, err := getAllPodsInNamespaceManagedByOlmOperator(strings.TrimSpace(targetNamespace))
+		if err != nil {
+			continue
+		}
+		podList = append(podList, pods...)
+	}
+
+	return podList
+}
+
+func getAllPodsInNamespaceManagedByOlmOperator(targetNamespace string) (managedPods []corev1.Pod, err error) {
+	client := clientsholder.GetClientsHolder()
+
+	podsList, err := client.K8sClient.CoreV1().Pods(targetNamespace).List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	for index := range podsList.Items {
+		labels := podsList.Items[index].Labels
+
+		if _, ok := labels["olm.managed"]; !ok { // Ignore non-OLM pods in the namespace
+			continue
+		}
+
+		managedPods = append(managedPods, podsList.Items[index])
+	}
+	return managedPods, nil
 }
--- a/pkg/autodiscover/autodiscover.go
+++ b/pkg/autodiscover/autodiscover.go
@@ -31,7 +31,7 @@ import (
 	"github.com/test-network-function/cnf-certification-test/internal/log"
 	"github.com/test-network-function/cnf-certification-test/pkg/compatibility"
 	"github.com/test-network-function/cnf-certification-test/pkg/configuration"
-	"github.com/test-network-function/cnf-certification-test/pkg/provider"
+	"github.com/test-network-function/cnf-certification-test/pkg/podhelper"
 	"helm.sh/helm/v3/pkg/release"
 	appsv1 "k8s.io/api/apps/v1"
 	scalingv1 "k8s.io/api/autoscaling/v1"
@@ -189,7 +189,10 @@ func DoAutoDiscover(config *configuration.TestConfiguration) DiscoveredTestData 
 	// Get all operator pods
 	var allOperatorPods []corev1.Pod
 	for _, csv := range data.Csvs {
-		pods := getOperatorCsvPods(csv)
+		pods, podError := getOperatorCsvPods(csv)
+		if podError != nil {
+			log.Fatal("Cannot get pods from CSV %s, err: %v", csv.Name, podError)
+		}
 		allOperatorPods = append(allOperatorPods, pods...)
 	}
 	data.AllOperatorPods = allOperatorPods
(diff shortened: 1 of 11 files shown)


cnf-certification-test/operator/util.go
```diff
@@ -0,0 +1,80 @@
+package operator
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/test-network-function/cnf-certification-test/internal/clientsholder"
+	v1 "k8s.io/api/core/v1"
```
sebrandon1, line 8:
```suggestion
	corev1 "k8s.io/api/core/v1"
```

sebrandon1 (commented):
One comment about the import.

Also we probably should have some simple unit tests wrapping these.

cnf-certification-test/operator/util.go
```diff
@@ -0,0 +1,80 @@
...
+		podList = append(podList, pods...)
+	}
+
+	return podList, nil
+}
+
+func getOperatorGroupTargetNamespaceByName(operatorGroupName string) ([]string, error) {
+	client := clientsholder.GetClientsHolder()
+
+	operatorGroup, err := client.OlmClient.OperatorsV1().OperatorGroups(operatorGroupName).Get(context.TODO(), operatorGroupName, metav1.GetOptions{})
```
edcdavid, line 44: OperatorGroups(operatorGroupName) should be namespace (operatorGroupNamespace) ?

cnf-certification-test/operator/util.go
```diff
@@ -0,0 +1,95 @@
...
+		annotations := podsList.Items[index].Annotations
+		if annotations["olm.operatorGroup"] == operatorGroupName {
+			managedPods = append(managedPods, &podsList.Items[index])
+		}
+	}
+	return managedPods, nil
+}
+
+// Filter cluster operators
+func getNonClusterOperatorCsvs(allCsvs []*olmv1Alpha.ClusterServiceVersion) (nonClusterOperatorCsvs []*olmv1Alpha.ClusterServiceVersion) {
```
edcdavid, line 85: Why not support operators install with all namespaces mode?

cnf-certification-test/operator/util.go
bnshr: Yeah, right. I think the code got messed up a bit during refactoring. I moved stuffs from here to there. Otherwise, the get happened based on operator namespace and operator group name. Thanks!

cnf-certification-test/operator/util.go
bnshr: The Operator Best Practice docs only is considering application operators. So, we assume that infrastructure operators are those which has InstallMode = AllNamespace. This is a strange assumption, but we do not know how to different infra from app operators. Hence, this logic.

cnf-certification-test/operator/util.go
```diff
@@ -0,0 +1,96 @@
...
+	// Get the operator group to get the targetNamespace
+	operatorTargetNamespaces, err := getOperatorGroupTargetNamespaceByName(operatorNamespace, operatorGroupName)
+	if err != nil {
+		return nil, fmt.Errorf("operator Group for found for %s", csvName)
+	}
+
+	var podList []*corev1.Pod
+
+	for _, targetNamespace := range operatorTargetNamespaces {
+		pods, err := getAllPodsManagedByOperatorGroup(targetNamespace, operatorGroupName)
```
edcdavid, line 47: Don't we need more checks to make sure that the pods belong to the operators and namespaces under test only? The operator group can cover multiple operators, not only those under test, same for the namespaces(although maybe that would be a test config error).

bnshr: Namespaces under test - N/A. We consider all operators (excluding cluster operators).

The logic for the client code should consider only non-cluster operators. And that is done this way
* Get all CSVs from env.csv
* Remove non-cluster operator CSVs

In this code, for each non-cluster operator CSV, get operator group to list the target namespaces. And operator group is the linking between all these (CSV, target namespaces and pods). 

I will take a look once again and add some labels check like ""app.kubernetes.io/created-by": "my-operator" in the resources. Then we need a mapping between operator name and CSV name.

edcdavid: Regarding namespace under test, I was thinking of a scenario where the target namespaces in operator group are ns1 and ns1 but the suite is running with a namespace under test of n1 only. Then wouldn't pods belonging to ns2 returned by this code?

cnf-certification-test/operator/util_test.go
```diff
@@ -0,0 +1,147 @@
...
+// with this program; if not, write to the Free Software Foundation, Inc.,
+// 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+package operator
+
+import (
+	"testing"
+
+	olmv1Alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
+	"gotest.tools/v3/assert"
```
sebrandon1, line 23: Use `stretchr/testify`.

cnf-certification-test/operator/util.go
```diff
@@ -0,0 +1,96 @@
...
+	client := clientsholder.GetClientsHolder()
+
+	podsList, err := client.K8sClient.CoreV1().Pods(targetNamespace).List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	for index := range podsList.Items {
+		annotations := podsList.Items[index].Annotations
+		if annotations["olm.operatorGroup"] == operatorGroupName {
```
sebrandon1, line 78: Does case need to be taken into account here?  Should we use like `ToLower` before comparing the names?

cnf-certification-test/operator/util.go
```diff
@@ -0,0 +1,96 @@
...
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// Get the list of pods mananged by the operator - retrieve from the CSV and then check the operator group to find the target namespaces to retrieve the managed pods
+func GetNonClusterOperatorCsvPods(csv *olmv1Alpha.ClusterServiceVersion) ([]*corev1.Pod, error) {
+	csvName := csv.Name
+	annotations := csv.Annotations
+	operatorGroupName := annotations["olm.operatorGroup"]
+	operatorNamespace := annotations["olm.operatorNamespace"]
```
sebrandon1, line 36: Should we have a pre-check here that early exits if these don't exist?

sebrandon1 (commented):
Some more comments

cnf-certification-test/operator/util.go
bnshr: If the namespace under test which is defined by tnf config file is mentioned, the newly added tests for operator best practices do not regard that. **These test cases consider all the operators in the test environment.**

cnf-certification-test/operator/util.go
bnshr: It's a map of annotations. So we need to get the value using the same key. I would add exists check here.

cnf-certification-test/operator/util.go
bnshr: This is generated by OLM. So, they would be present always. Still, I can add some checks.

sebrandon1 (dismissed):

sebrandon1:
@bnshr One more thing we could add here to make everything easier to consume, is to add `env.AllOperatorPods` as part of our existing pod groupings:

https://github.com/test-network-function/cnf-certification-test/blob/main/pkg/provider/provider.go#L74-L77

Just add:
```
// Pod Groupings
Pods      []*Pod                 `json:"testPods"`
DebugPods map[string]*corev1.Pod // map from nodename to debugPod
AllPods   []*Pod                 `json:"AllPods"`
AllOperatorPods []*Pod `json:"AllOperatorPods"`           <---------------
```

We could just easily call the funcs you have created here to collect all of the pods during the autodiscover process.

bnshr:
> @bnshr One more thing we could add here to make everything easier to consume, is to add `env.AllOperatorPods` as part of our existing pod groupings:
> 
> https://github.com/test-network-function/cnf-certification-test/blob/main/pkg/provider/provider.go#L74-L77
> 
> Just add:
> 
> ```
> // Pod Groupings
> Pods      []*Pod                 `json:"testPods"`
> DebugPods map[string]*corev1.Pod // map from nodename to debugPod
> AllPods   []*Pod                 `json:"AllPods"`
> AllOperatorPods []*Pod `json:"AllOperatorPods"`           <---------------
> ```
> 
> We could just easily call the funcs you have created here to collect all of the pods during the autodiscover process.

Makes sense, as I can see that 4 new test cases are created by Shimrit.

sebrandon1 (dismissed):

pkg/autodiscover/autodiscover.go
```diff
@@ -280,3 +290,43 @@ func getOpenshiftVersion(oClient clientconfigv1.ConfigV1Interface) (ver string,
 
 	return "", errors.New("could not get openshift version from clusterOperator")
 }
+
+// Get the list of pods mananged by the operator - retrieve from the CSV and then check the operator group to find the target namespaces to retrieve the managed pods
+func getOperatorCsvPods(csv *olmv1Alpha.ClusterServiceVersion) []corev1.Pod {
+	annotations := csv.Annotations
+
+	targetNamespaces := annotations["olm.targetNamespaces"]          // This is a comma-separated string, example : a,b,c where a, b and c are target namespaces
```
edcdavid, line 298: The targetNamespaces is an attribute of the operatorGroup, which can be shared by multiple operators. So if 2 operators op1 and op2 share the same operator group, and only op1 is selected via label, this code is going to catch pods from op1 and op2, whereas we should only catch pods from op1.

bnshr: True, then we do not have a way to get only op1 related pods. Because there is no mapping between these pods created by the controller of the operator op1.

bnshr: Thanks @edcdavid 👍  I updated the code as discussed.

pkg/podhelper/podhelper_test.go
```diff
@@ -0,0 +1,92 @@
...
+
+import (
+	"reflect"
+	"testing"
+
+	olmv1Alpha "github.com/operator-framework/api/pkg/operators/v1alpha1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+
+	v1app "k8s.io/api/apps/v1"
```
sebrandon1, line 11:
```suggestion
	appsv1 "k8s.io/api/apps/v1"
```

pkg/autodiscover/autodiscover.go
```diff
@@ -56,6 +59,7 @@ type DiscoveredTestData struct {
 	Pods                   []corev1.Pod
 	AllPods                []corev1.Pod
 	DebugPods              []corev1.Pod
+	AllOperatorPods        map[string][]*corev1.Pod
```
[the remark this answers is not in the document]
bnshr: This is the name and namespace of the csv. The mapping is between csv and its pods in multiple namespaces if mentioned in the `targetNamespaces`.

pkg/podhelper/podhelper_test.go
bnshr: Changed in other unrelated files also!

sebrandon1: Thanks!

pkg/autodiscover/autodiscover.go
```diff
@@ -280,3 +290,81 @@ func getOpenshiftVersion(oClient clientconfigv1.ConfigV1Interface) (ver string,
...
+
+		for _, targetNamespace := range operatorTargetNamespaces {
+			pods, err := getCsvPodsFromTargetNamespace(csv, strings.TrimSpace(targetNamespace), client)
+			if err != nil {
+				continue
+			}
+			podList = append(podList, pods...)
+		}
+
+		csvToPodsMapping[fmt.Sprintf("%s, namespace=%s", csv.Name, csv.Namespace)] = podList
```
edcdavid, line 315: create a constant with "%s, ns=%s", since the user of this map will need to use it too

pkg/provider/provider.go
```diff
@@ -72,9 +72,10 @@ type TestEnvironment struct { // rename this with testTarget
 	AbnormalEvents []*Event
 
 	// Pod Groupings
-	Pods      []*Pod                 `json:"testPods"`
-	DebugPods map[string]*corev1.Pod // map from nodename to debugPod
-	AllPods   []*Pod                 `json:"AllPods"`
+	Pods            []*Pod                 `json:"testPods"`
+	DebugPods       map[string]*corev1.Pod // map from nodename to debugPod
+	AllPods         []*Pod                 `json:"AllPods"`
+	AllOperatorPods map[string][]*Pod      `json:"AllOperatorPods"`
```
edcdavid, line 78: suggesting renaming AllOperatorPods to CSVToPodListMap

pkg/autodiscover/autodiscover.go
```diff
@@ -280,3 +290,81 @@ func getOpenshiftVersion(oClient clientconfigv1.ConfigV1Interface) (ver string,
...
+	// Get all pods from the target namespace
+	podsList, err := client.K8sClient.CoreV1().Pods(targetNamespace).List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	for index := range podsList.Items {
+		labels := podsList.Items[index].Labels
+
+		if _, ok := labels["olm.managed"]; !ok { // Ignore non-OLM pods in the namespace
```
edcdavid, line 331: not all pods are labeled, remove this line to capture all operator pods.

pkg/autodiscover/autodiscover.go
bnshr: Sure, but we do not need this as @sebrandon1 suggested to add csv info in provider.Pod.

pkg/autodiscover/autodiscover.go
bnshr: I think this condition is necessary. I checked that OLM adds this annotation in the controller created pods. This filter condition is added to remove some other pods deployed in the same namespace.

edcdavid: I checked with the hazelcast operator controller pod that we use and it didn't have the label, yet the pod traces back to the csv with the owner references, so would populate the CSVToPodListMap.

bnshr: Wait .. we are not considering the operator controller pod as the workload. Rather the pods created by the controller code. I checked with CNF operator and its created pods by the CR and we get this label.

bnshr:
> from change #2130:
> 
> * ERROR no DCI job found

There is no change related to DCI job. Any idea why DCI job is missing?

sebrandon1:
@bnshr Sometimes that happens if you aren't fully rebased on `main`.

bnshr:
> @bnshr Sometimes that happens if you aren't fully rebased on `main`.

I see. Yes, the branch was not rebased until today. Hope this time it would work.

sebrandon1 (approved):

PR #2135, opened 2024-06-05
CI workflows: delete legacy tasks using the old binary

jmontesi:
The same tasks using the new "certsuite" binary are already working as expected, so there is no need to duplicate the verification.

Also, in line with the recent renaming some "tnf" references are changed to "certsuite".

### DIFF
--- a/.github/workflows/pre-main.yaml
+++ b/.github/workflows/pre-main.yaml
@@ -218,7 +218,7 @@ jobs:
 
       # Perform smoke tests.
       - name: 'Test: Run test suites'
-        run: ./certsuite run --label-filter="${SMOKE_TESTS_LABELS_FILTER}" --output-dir="certsuite-out" --log-level=${SMOKE_TESTS_LOG_LEVEL}
+        run: ./certsuite run --label-filter="${SMOKE_TESTS_LABELS_FILTER}" --output-dir=certsuite-out --log-level="${SMOKE_TESTS_LOG_LEVEL}"
 
       - name: Upload smoke test results as an artifact
         uses: actions/upload-artifact@65462800fd760344b1a7b4382951275a0abb4808 # v4.3.3
@@ -235,7 +235,7 @@ jobs:
         run: ./certsuite check results --log-file="certsuite-out/certsuite.log"
 
       - name: 'Test: Run preflight specific test suite'
-        run: ./certsuite run --label-filter="preflight" --log-level=${SMOKE_TESTS_LOG_LEVEL}
+        run: ./certsuite run --label-filter=preflight --log-level="${SMOKE_TESTS_LOG_LEVEL}"
 
       - name: Upload preflight smoke test results as an artifact
         uses: actions/upload-artifact@65462800fd760344b1a7b4382951275a0abb4808 # v4.3.3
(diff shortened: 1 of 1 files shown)


.github/workflows/pre-main.yaml
```diff
@@ -13,26 +13,17 @@ permissions:
...
-  ON_DEMAND_DEBUG_PODS: false
+  CERTSUITE_CONFIG_DIR: /tmp/certsuite/config
+  CERTSUITE_OUTPUT_DIR: /tmp/certsuite/output
+  SMOKE_TESTS_LOG_LEVEL: debug
+  SMOKE_TESTS_LABELS_FILTER: all
   TERM: xterm-color
   CM_BIN: /usr/local/bin/checkmake
   CM_URL_LINUX: https://github.com/mrtazz/checkmake/releases/download/0.2.2/checkmake-0.2.2.linux.amd64 # yamllint disable-line
-  SMOKE_TESTS_LABELS_FILTER: all
-  SKIP_PRELOAD_IMAGES: true
```
sebrandon1, line 35: I think we need to keep this flag `SKIP_PRELOAD_IMAGES`.

sebrandon1 (commented):
One comment

.github/workflows/pre-main.yaml
jmontesi: Done.

sebrandon1 (approved):

.github/workflows/pre-main.yaml
```diff
@@ -226,47 +218,35 @@ jobs:
 
       # Perform smoke tests.
       - name: 'Test: Run test suites'
-        run: TNF_LOG_LEVEL=${TNF_SMOKE_TESTS_LOG_LEVEL} ./run-cnf-suites.sh -l "${SMOKE_TESTS_LABELS_FILTER}"
+        run: ./certsuite run --label-filter="${SMOKE_TESTS_LABELS_FILTER}" --output-dir="certsuite-out" --log-level=${SMOKE_TESTS_LOG_LEVEL}
```
rdavid, line 221: No need quotes for `certsuite-out` and need for `${SMOKE_TESTS_LOG_LEVEL}`.

.github/workflows/pre-main.yaml
```diff
@@ -226,47 +218,35 @@ jobs:
...
-        run: make build-certsuite-tool
+        run: rm -f certsuite-out/*.tar.gz
 
       - name: Check the smoke test results against the expected results template
-        run: make check-results
+        run: ./certsuite check results --log-file="certsuite-out/certsuite.log"
 
       - name: 'Test: Run preflight specific test suite'
-        run: TNF_LOG_LEVEL=${TNF_SMOKE_TESTS_LOG_LEVEL} ./run-cnf-suites.sh -l "preflight"
+        run: ./certsuite run --label-filter="preflight" --log-level=${SMOKE_TESTS_LOG_LEVEL}
```
rdavid, line 238: same comment on quates.

rdavid (approved):

.github/workflows/pre-main.yaml
jmontesi: Done.

.github/workflows/pre-main.yaml
jmontesi: Done.

PR #2166, opened 2024-06-14
Add CRC testing against Ubuntu

sebrandon1:


sebrandon1:
After a bunch of testing, once again the Ubuntu runners are just not powerful enough to be able to run the CRC deployment and be able to run the tests at the same time.  There are some places where it passes the QE test fully, but there are other tests that fail out hard and sometimes the runs do not complete all together.

sebrandon1:
If we had larger runners, we would be able to complete this easily I am guessing.

PR #2167, opened 2024-06-16
Fix batch script to enable run on an entire catalog

shirmoran:
- Added a helper function to ensure all packages are reachable before writing them in the operator list.
- Delete catalog only after all the operators in operator list were tested.

### DIFF
--- a/script/run-basic-batch-operators-test.sh
+++ b/script/run-basic-batch-operators-test.sh
@@ -275,9 +275,9 @@ wait_all_packages_ok() {
 
 get_packeges() {
 	oc get packagemanifest \
-		-n ${OPERATOR_CATALOG_NAMESPACE} -o json \
-		| jq -r '.items[] | select(.status.catalogSource == "'${OPERATOR_CATALOG_NAME}'") | .metadata.name' \
-		| wc -w
+		-n ${OPERATOR_CATALOG_NAMESPACE} -o json |
+		jq -r '.items[] | select(.status.catalogSource == "'${OPERATOR_CATALOG_NAME}'") | .metadata.name' |
+		wc -w
 }
 
 wait_for_csv_to_appear_and_label() {
--- a/script/run-basic-batch-operators-test.sh
+++ b/script/run-basic-batch-operators-test.sh
@@ -236,6 +236,44 @@ wait_package_ok() {
 	done
 }
 
+wait_all_packages_ok() {
+	local \
+		start_time \
+		timeout_seconds=600
+
+	prev_count=$(get_packeges)
+	start_time=$(date +%s 2>&1) || {
+		echo "date failed with error $?: $start_time" >>"$LOG_FILE_PATH"
+		return 0
+	}
+
+	# wait until package number is stable
+	while true; do
+		curr_count=$(get_packeges)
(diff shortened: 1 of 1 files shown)


run-basic-batch-operators-test.sh
```diff
@@ -232,6 +232,44 @@ wait_package_ok() {
 	done
 }
 
+wait_all_packages_ok() {
+	local \
+		start_time \
+		timeout_seconds=600
+
+	prev_count=$(get_packeges)
```
rdavid, line 240: Should all variables be local?

run-basic-batch-operators-test.sh
```diff
@@ -232,6 +232,44 @@ wait_package_ok() {
...
+	start_time=$(date +%s 2>&1) || {
+		echo "date failed with error $?: $start_time" >>"$LOG_FILE_PATH"
+		return 0
+	}
+
+	# wait until package number is stable
+	while true; do
+		curr_count=$(get_packeges)
+		if [ "${curr_count}" -ne "${prev_count}" ] || [ "${curr_count}" -eq 0 ]; then
+			prev_count=${curr_count}
```
rdavid, line 250:
```suggestion
			prev_count="${curr_count}"
```

run-basic-batch-operators-test.sh
```diff
@@ -232,6 +232,44 @@ wait_package_ok() {
...
+		else
+			return 0
+		fi
+
+		curr_time=$(date +%s)
+		elapsed_time=$((curr_time - start_time))
```
rdavid, line 256:
```suggestion
		elapsed_time="$((curr_time - start_time))"
```

run-basic-batch-operators-test.sh
```diff
@@ -232,6 +232,44 @@ wait_package_ok() {
...
+		fi
+
+		# Otherwise wait a bit
+		echo_color "$BLUE" "Waiting for packages to be reachable..."
+		sleep 5
+	done
+}
+
+get_packeges() {
+	oc get packagemanifest -n ${OPERATOR_CATALOG_NAMESPACE} -o json | jq -r '.items[] | select(.status.catalogSource == "'${OPERATOR_CATALOG_NAME}'") | .metadata.name' | wc -w
```
rdavid, line 270: Add error handling, e.g.:
command || {
  local err=$?
  echo >&2 "Command failed with $err."
  exit $err
}

rdavid (approved):

run-basic-batch-operators-test.sh
shirmoran: Could be. I followed other function structure, so I didn't create all of them a local, but it could be changed.
What do you think?

rdavid: Yes, a new function should be of the best quality. We will fix other functions in a separate PR.

rdavid:
@shirmoran Yes, a new function should be of the best quality. We will fix other functions in a separate PR.

run-basic-batch-operators-test.sh
shirmoran: we want to keep on checking the packages, if there will be an error the elapsed time will reach the limit and exit (line 258)

rdavid:
@shirmoran Good to go!

PR #2181, opened 2024-06-20
Dockerfile: do not build the legacy binary

jmontesi:
build-depends: https://github.com/redhatci/ansible-collection-redhatci-ocp/pull/356

The container image is updated to avoid building the old binary and copying stuff that it is not required anymore.

The "certsuite" binary is installed in "/usr/local/bin" so that it can be directly called.

The GitHub workflows have been updated accordingly.

An update in DCI and QE will be necessary as well.

### DIFF
--- a/.github/workflows/pre-main.yaml
+++ b/.github/workflows/pre-main.yaml
@@ -344,31 +344,31 @@ jobs:
       - name: 'Test: Run without any TS, just get diagnostic information'
         run: |
           docker run --rm --network host \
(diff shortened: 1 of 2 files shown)


sebrandon1 (approved):

rdavid (approved):

ramperher:
Please **do not merge** this change until this change in the collections is validated: https://github.com/redhatci/ansible-collection-redhatci-ocp/pull/356, else DCI automation will be broken

ramperher (approved):
confirmed it's working, thanks!

PR #2196, opened 2024-06-25
Add check pkg unit tests

sebrandon1:
Includes some changes from: #2195

### DIFF
--- a/internal/log/log.go
+++ b/internal/log/log.go
@@ -114,7 +114,11 @@ func GetMultiLogger(writers ...io.Writer) *Logger {
 		},
 	}
 
-	handlers := []slog.Handler{globalLogger.l.Handler()}
+	var handlers []slog.Handler
+	if globalLogger != nil {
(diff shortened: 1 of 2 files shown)


internal/log/log.go
```diff
@@ -114,7 +114,11 @@ func GetMultiLogger(writers ...io.Writer) *Logger {
 		},
 	}
 
-	handlers := []slog.Handler{globalLogger.l.Handler()}
+	var handlers []slog.Handler
+	if globalLogger != nil {
```
jmontesi, line 118: This check can be avoided by setting up a global logger inside the test cases that need it.
If the global logs (not the check's logs) are not used, the global logger can be set up as follows:

`log.SetupLogger(os.Stdout, "INFO")`

If access to the global logs is required then use:
```
var logArchive strings.Builder
log.SetupLogger(&logArchive, "INFO")
```

sebrandon1: Going to leave the nil check in this to make sure globalLogger isn't nil and we don't really need the logger for UTs.

PR #2208, opened 2024-06-28
workflows: add new workflow to upload release assets

jmontesi:
It builds and uploads the Certsuite binaries compressed in tar.gz for x86_64 and ARM 64.

build-depends: 31882


greyerof (approved):

ramperher:
DCI failure is not related to your change, it's an issue in our side, adding `build-depends: 31882` to the PR description to include the fix in our side while it's not merged.

PR #2213, opened 2024-07-01
configure : Test operator labels in tnf_config yaml

bnshr:
https://issues.redhat.com/browse/CNFCERT-990

This PR makes the lifecycle-cr-scaling test cases as PASSED from skipped as the tnf_config is configured to use the the deployed test operator (cr-scale-operator) and run this test case.

### DIFF
--- a/cnf-certification-test/tnf_config.yml
+++ b/cnf-certification-test/tnf_config.yml
@@ -5,17 +5,18 @@ podsUnderTestLabels:
 operatorsUnderTestLabels:
   - "test-network-function.com/operator:target"
   - "test-network-function.com/operator1:new"
+  - "cnf/test:cr-scale-operator"
 targetCrdFilters:
   - nameSuffix: "group1.test.com"
     scalable: false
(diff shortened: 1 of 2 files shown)


aabughosh (approved):

cnf-certification-test/tnf_config.yml
```diff
@@ -1,21 +1,23 @@
 targetNameSpaces:
   - name: tnf
+  - name: default # Added for testing cr-scale-operator, as it is installed in default namespace
```
aabughosh, line 3: defulat is not a good ns 
"description": "Tests that all CNF's resources (PUTs and CRs) belong to valid namespaces. A valid namespace meets\nthe following conditions: (1) It was declared in the yaml config file under the targetNameSpaces\ntag. (2) It does not have any of the following prefixes: default, openshift-, istio- and aspenmesh-"
maybe change the installation ns?

bnshr: Installed in tnf

greyerof (approved):

PR #2228, opened 2024-07-08
cmd/certsuite: add new "certsuite info" command to display Catalog info

jmontesi:
At the moment only the "--test-case/-t" flag is supported to show the information available in the Catalog for a particular test case.

### DIFF
--- a/cmd/certsuite/info/info.go
+++ b/cmd/certsuite/info/info.go
@@ -9,6 +9,7 @@ import (
 	"github.com/test-network-function/cnf-certification-test/cnf-certification-test/identifiers"
 	"github.com/test-network-function/cnf-certification-test/internal/cli"
 	"github.com/test-network-function/test-network-function-claim/pkg/claim"
+	"golang.org/x/term"
 )
 
 var (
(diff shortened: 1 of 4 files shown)


sebrandon1 (approved):

cmd/certsuite/info/info.go
```diff
@@ -0,0 +1,89 @@
...
+			testCase = identifiers.Catalog[id]
+			break
+		}
+	}
+
+	if testCase.Identifier.Id == "" {
+		return fmt.Errorf("no test case found matching name %q", testCaseFlag)
+	}
+
+	const lineMaxWidth = 120
```
edcdavid, line 37: Optional: get the width from terminal to properly wrap lines when the terminal is too small
```
package main

import "golang.org/x/term"

func main() {
    if term.IsTerminal(0) {
        println("in a term")
    } else {
        println("not in a term")
    }
    width, height, err := term.GetSize(0)
    if err != nil {
        return
    }
    println("width:", width, "height:", height)
}
```

edcdavid (approved):

cmd/certsuite/info/info.go
jmontesi: Done.

PR #2240, opened 2024-07-10
Add warning for invalid namespaces during autodiscover phase

sebrandon1:
`invalidNamespacePrefixes` was copied from the [accesscontrol/suite.go](https://github.com/test-network-function/cnf-certification-test/blob/df9d7703b3ffefc55f56a4a562604a63a0094a1b/cnf-certification-test/accesscontrol/suite.go#L50) so it exists in two places at once.  We can maybe correct this in the future if needed?

Right now you can potentially use `openshift-logging` or any of the invalid namespaces as part of your `targetNameSpaces` configuration values but it ends up failing the `access-control-namespace` tests.

### DIFF
--- a/pkg/autodiscover/autodiscover.go
+++ b/pkg/autodiscover/autodiscover.go
@@ -32,6 +32,7 @@ import (
 	"github.com/test-network-function/cnf-certification-test/pkg/compatibility"
 	"github.com/test-network-function/cnf-certification-test/pkg/configuration"
 	"github.com/test-network-function/cnf-certification-test/pkg/podhelper"
+	"github.com/test-network-function/cnf-certification-test/pkg/stringhelper"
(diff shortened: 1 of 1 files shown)


pkg/autodiscover/autodiscover.go
```diff
@@ -54,6 +55,15 @@ const (
 	labelTemplate              = "%s/%s"
 )
 
+var (
+	invalidNamespacePrefixes = []string{
+		"default",
+		"openshift-",
+		"istio-",
+		"aspenmesh-",
```
bnshr, line 63: Why `aspenmesh-`?

bnshr (approved):

sebrandon1:
Closing this PR as we will go another direction.

PR #2244, opened 2024-07-12
Remove operator read-only filesystem test

sebrandon1:
The `operator-sdk` does not set this value by default and if it is turned on, it causes a loop in operator controller pods.  Thanks to @ramperher for the analysis of the test case.


sebrandon1:
Marking this WIP until we decide if we want to delete it or not.

jmontesi (approved):
Do forget to update the `Catalog.md`.

sebrandon1:
Closing this as we are planning on keeping this test (for now).

PR #2250, opened 2024-07-12
Switch for-loops to range

sebrandon1:
I just saw some for-loops that could be switched over to `range`.


sebrandon1:
I forgot about the `gocritic` linter not liking these loops.

PR #2263, opened 2024-07-17
Remove some dependencies from the image

sebrandon1:
Seeing if things break if I remove these dependencies.

Follow up, it looks we can safely remove the `jq` and `gcc` deps.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -82,9 +82,9 @@ RUN make build-certsuite-tool \
 # Switch contexts back to the root CERTSUITE directory
 WORKDIR ${CERTSUITE_DIR}
 
-# Remove most of the build artefacts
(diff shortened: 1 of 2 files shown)


rdavid (approved):

Dockerfile
```diff
@@ -10,9 +10,7 @@ RUN \
 	mkdir ${CERTSUITE_DIR} \
 	&& dnf update --assumeyes --disableplugin=subscription-manager \
 	&& dnf install --assumeyes --disableplugin=subscription-manager \
-		gcc \
```
jmontesi, line 13: Remove `gcc` from line 89 as well.

jmontesi (approved):

PR #2264, opened 2024-07-18
cmd/certsuite: move the "list" flag from "run" to "info"

jmontesi:
As the "info" command already gets detailed Catalog information, it looks more convenient to integrate the functionality for listing test cases in it.

Also adds a new unit test for the "list" flag as part of the "info" command.


sebrandon1 (approved):

ramperher:
Sorry for the noise, just testing some stuff with some changes we're introducing for the refactor of the cnf_cert role

edcdavid (approved):

PR #2282, opened 2024-07-23
Change debug-partner to k8s-best-practices-debug

sebrandon1:
https://quay.io/repository/testnetworkfunction/k8s-best-practices-debug?tab=tags

https://github.com/test-network-function/k8s-best-practices-debug 

Instead of building and pushing the `debug` image from the `cnf-certification-test-partner` repo, we decided to de-couple the `debug` image from the complexity of the partner repo and just build it separately.

Changes to the partner repo (which is primarily used for the CI-related tasks) shouldn't affect the debug image (and versioning scheme) going forward.

### DIFF
--- a/version.json
+++ b/version.json
@@ -1,5 +1,5 @@
 {
-  "partner_tag": "v5.2.1",
+  "debug_tag": "v0.0.1",
   "claimFormat": "v0.4.0",
   "parserTag": "v0.4.7"
 }
--- a/version.json
+++ b/version.json
@@ -1,5 +1,5 @@
 {
-  "debug_tag": "v0.0.1",
+  "debugTag": "v0.0.1",
   "claimFormat": "v0.4.0",
   "parserTag": "v0.4.7"
 }
--- a/version.json
+++ b/version.json
@@ -1,5 +1,5 @@
 {
-  "debugTag": "v0.0.1",
+  "debugTag": "v0.0.2",
   "claimFormat": "v0.4.0",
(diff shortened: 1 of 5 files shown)


version.json
```diff
@@ -1,5 +1,5 @@
 {
-  "partner_tag": "v5.2.1",
+  "debug_tag": "v0.0.1",
```
sebrandon1, line 2: Do we care that this is not camel case like the other two variables?

sebrandon1:
@ramperher I know we just talked about the `partner_tag` variable in this file, so this change might affect your check.

version.json
edcdavid: This is used at build time only and only in this project. Should be safe to make it camel case

edcdavid (approved):
Optional: looking at it again wouldn't k8s-best-practices-debug-image,  DEBUG_IMAGE_VERSION, DEBUG_IMAGE_REPO, ... etc be clearer?

ramperher:
> @ramperher I know we just talked about the `partner_tag` variable in this file, so this change might affect your check.

I can see, I would not need to do any modification till you create a new release.
But then, my question is: does this mean that debug-partner image is no longer maintained, and we need to move to this new k8s-best-practices-debug image?

sebrandon1:
@ramperher We will maintain both images for a period of time/overlap.  No rush to remove the build/push from the old repo just yet.

rdavid (approved):

PR #2285, opened 2024-07-24
tests/observability: fix observability-pod-disruption-budged test case

jmontesi:
Use all the labels defined in the PDB's selector "MatchLabels" fields to match workloads, as the Kubernetes documentation specifies that those labels should be ANDed.

Also, take into account the "MatchExpressions" field of the selector when it can be properly converted to a map.

### DIFF
--- a/tests/observability/suite.go
+++ b/tests/observability/suite.go
@@ -27,13 +27,13 @@ import (
 	pdbv1 "github.com/test-network-function/cnf-certification-test/tests/observability/pdb"
 
 	"github.com/test-network-function/cnf-certification-test/internal/clientsholder"
-	"github.com/test-network-function/cnf-certification-test/internal/datautil"
 	"github.com/test-network-function/cnf-certification-test/internal/log"
 	"github.com/test-network-function/cnf-certification-test/pkg/checksdb"
 	"github.com/test-network-function/cnf-certification-test/pkg/provider"
 	"github.com/test-network-function/cnf-certification-test/pkg/testhelper"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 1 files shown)


tests/observability/suite.go
```diff
@@ -193,22 +195,29 @@ func testPodDisruptionBudgets(check *checksdb.Check, env *provider.TestEnvironme
...
-						compliantObjects = append(compliantObjects, testhelper.NewReportObject("Deployment: references PodDisruptionBudget", testhelper.DeploymentType, true).
-							AddField(testhelper.DeploymentName, d.Name).
-							AddField(testhelper.Namespace, d.Namespace).
-							AddField(testhelper.PodDisruptionBudgetReference, env.PodDisruptionBudgets[pdbIndex].Name))
-					}
+			pdb := &env.PodDisruptionBudgets[pdbIndex]
+			if pdb.Namespace != d.Namespace {
+				continue
+			}
+			pdbSelectorMap, err := metav1.LabelSelectorAsMap(pdb.Spec.Selector)
```
sebrandon1, line 202: `LabelSelectorAsMap` never used that one before.  Nice!

edcdavid: @jmontesi by "Also, take into account the "MatchExpressions" field of the selector when it can be properly converted to a map." do you mean to leave out some MatchExpressions that are not covered by LabelSelectorAsMap ( e.g. 
"Operators that cannot be converted into plain labels (Exists, DoesNotExist, NotIn, and In with more than one value") ? Shouldn't we support all these cases?

jmontesi: Yes, I left those cases out as I considered them quite unlikely to be used in practice to match a PDB with a workload. But reviewing again the Kubernetes API I found a way to nicely cover all cases.

edcdavid (approved):

sebrandon1 (approved):

PR #2299, opened 2024-07-29
Add option to sanitize claim file result output

sebrandon1:
Adds a flag `--sanitize-claim` which defaults to false.  This is mostly used for QE.

Modifies the `NewLabelsExprEvaluator` and related functions to its own package named `labels`.

### DIFF
--- a/pkg/claimhelper/claimhelper.go
+++ b/pkg/claimhelper/claimhelper.go
@@ -148,6 +148,11 @@ func (c *ClaimBuilder) Build(outputFile string) {
 	log.Info("Claim file created at %s", outputFile)
 }
 
+func (c *ClaimBuilder) GetClaimResults() *claim.Root {
+	c.claimRoot.Claim.Results = checksdb.GetReconciledResults()
+	return c.claimRoot
+}
+
 //nolint:funlen
 func populateXMLFromClaim(c claim.Claim, startTime, endTime time.Time) TestSuitesXML {
 	const (
@@ -302,11 +307,7 @@ func ReadClaimFile(claimFileName string) (data []byte, err error) {
 	if err != nil {
 		log.Error("ReadFile failed with err: %v", err)
 	}
-	path, err := os.Getwd()
-	if err != nil {
-		log.Error("Getwd failed with err: %v", err)
-	}
-	log.Info("Reading claim file at path: %s", path)
+	log.Info("Reading claim file at path: %s", claimFileName)
 	return data, nil
 }
 
@@ -340,6 +341,7 @@ func MarshalClaimOutput(claimRoot *claim.Root) []byte {
 
 // WriteClaimOutput writes the output payload to the claim file.  In the event of an error, this method fatally fails.
 func WriteClaimOutput(claimOutputFile string, payload []byte) {
+	log.Info("Writing claim data to %s", claimOutputFile)
 	err := os.WriteFile(claimOutputFile, payload, claimFilePermissions)
 	if err != nil {
 		log.Fatal("Error writing claim data:\n%s", string(payload))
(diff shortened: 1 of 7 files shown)


pkg/claimhelper/claimhelper.go
```diff
@@ -374,3 +376,30 @@ func CreateClaimRoot() *claim.Root {
...
+	if err != nil {
+		log.Error("ReadClaimFile failed with err: %v", err)
+		return "", err
+	}
+	var aRoot claim.Root
+	UnmarshalClaim(data, &aRoot)
+
+	// Remove the results that do not match the labels filter
+	for testID := range aRoot.Claim.Results {
+		if aRoot.Claim.Results[testID].TestID.Id != labelsFilter {
```
jmontesi, line 392: Does this work for a `labelsFilter` that is not the name of a test case? For instance for a test suite name or a label such as `telco` or `all`.

pkg/claimhelper/claimhelper.go
```diff
@@ -374,3 +376,30 @@ func CreateClaimRoot() *claim.Root {
...
+			log.Info("Removing test ID: %s from the claim", testID)
+			delete(aRoot.Claim.Results, testID)
+		}
+	}
+
+	log.Info("Number of results after sanitizing: %d", len(aRoot.Claim.Results))
```
jmontesi, line 398: These log messages seem irrelevant.

pkg/claimhelper/claimhelper.go
```diff
@@ -148,6 +148,11 @@ func (c *ClaimBuilder) Build(outputFile string) {
 	log.Info("Claim file created at %s", outputFile)
 }
 
+func (c *ClaimBuilder) GetClaimResults() *claim.Root {
```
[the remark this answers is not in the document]
sebrandon1: Ah yes, I was using this for experimenting.  I'll remove.

pkg/claimhelper/claimhelper.go
jmontesi: You could make it work for any `labelsFilter` by using the `labelsExprEvaluator` (and making it available outside the `checksdb` package) like this:

`if !labelsExprEvaluator.Eval(aRoot.Claim.Results[testID].TestID.Id) {
// delete test
}`

Otherwise I think that the functionality is too restricted to make a flag out of it.

sebrandon1: Done, this works.

pkg/claimhelper/claimhelper.go
sebrandon1: Removed.

edcdavid (commented):
when using the --sanitize-claim, the parser output is broken. It does not display any tests. Investigating ...
I used a suite label "access-control" and there is not tests under "results"

pkg/claimhelper/claimhelper.go
edcdavid: I couldn't make this work, looks like for each test ID read from the claim file, we need to generate metadata with identifiers.GetTestIDAndLabels(identifiers.TestSysAdminIdentifier), including the suite name and tag

sebrandon1: Ah I see, yes I wonder if I change:
`if !evaluator.Eval([]string{aRoot.Claim.Results[testID].TestID.Id})`
to 
`if !evaluator.Eval([]string{aRoot.Claim.Results[testID].TestID.Id, aRoot.Claim.Results[testID].TestID.Suite})`

Since the Eval() func accepts a slice of strings to evaluate.  I'll try that out.

sebrandon1: Okay yes this worked.  I was able to run:
`./certsuite run --sanitize-claim -l access-control`

And it keeps all of the access-control related tests in the claim output.

edcdavid: What about just calling the function to return the labels to match on with something like:
```
		_,labels:=identifiers.GetTestIDAndLabels(*aRoot.Claim.Results[testID].TestID)
		if !evaluator.Eval( labels) {
```
That would also cover tags

sebrandon1: This works as well.  Updated!

edcdavid (approved):

PR #2308, opened 2024-07-31
Various CI Updates

sebrandon1:
This PR needs to be merged in tandem with: https://github.com/test-network-function/cnfcert-tests-verification/pull/857/

TODO: Remove the `tnf_rename` ref's in the code after merge.


.github/workflows/qe-hosted.yml
```diff
@@ -13,6 +13,10 @@ permissions:
 
 env:
   QE_REPO: test-network-function/cnfcert-tests-verification
+  TEST_TNF_IMAGE_NAME: quay.io/testnetworkfunction/cnf-certification-test
```
jmontesi, line 16: What about using `CERTSUITE_IMAGE_NAME` instead of `TEST_TNF_IMAGE_NAME ` .

The same applies to the names below.

sebrandon1: Good call I'll update.

sebrandon1:
The QE tests will fail because they are expecting certain variables.  I'm doing more testing.

PR #2320, opened 2024-08-06
Comment should end in a period

rdavid:


.github/workflows/pre-main.yaml
```diff
@@ -74,7 +74,7 @@ jobs:
         uses: golangci/golangci-lint-action@aaa42aa0628b4ae2578232a66b541047968fac86 # v6.1.0
         with:
           version: v1.59
-          args: --timeout 10m0s
+          args: --enable godot --timeout 10m0s
```
sebrandon1, line 77: Can we just move this to the `.golangci.yml`?

jmontesi:
I would not enforce this linting rule. I think this [recommendation](https://go.dev/wiki/CodeReviewComments#comment-sentences) has been misinterpreted here as it refers to "comments documenting declarations", not to all comments.

sebrandon1:
I agree with @jmontesi after looking at it I think this isn't needed for this repo.

rdavid:
@jmontesi What do you mean by “comments documenting declarations”? Are you referring to [Go Doc Comments](https://tip.golang.org/doc/comment)? The article actually discusses using full sentences. I added periods to some comments inside functions, which isn’t necessary, but the linter checks specifically for Go Doc Comments.

jmontesi:
> @jmontesi What do you mean by “comments documenting declarations”? Are you referring to [Go Doc Comments](https://tip.golang.org/doc/comment)? The article actually discusses using full sentences. I added periods to some comments inside functions, which isn’t necessary, but the linter checks specifically for Go Doc Comments.

Basically I'm referring to the periods being added to all type comments regardless of if they document a declaration or not. For instance, this comment in the middle of a function:

`// flushes buffer to screen.`

As you modified many of those comments I thought the linter was enforcing that. But even if it only acts on comments documenting declarations I doubt if it's very relevant to our project at the moment because there's probably some previous work to do to have an organized and coherent GoDoc before focusing in those details.

rdavid:
@jmontesi Comments inside functions aren’t required to be full sentences by the linter or Golang guidelines. I can remove the periods from these comments in the PR, although I personally prefer to make every comment a full sentence. How is the Golang article's requirement not relevant to any Golang code? What about other controversial design decisions, such as tabs vs. spaces, camelCase vs. snake_case, and so on?

jmontesi:
@rdavid If you are referring to this [recommendation](https://go.dev/wiki/CodeReviewComments#comment-sentences), it clearly says that the beginning that "**Comments documenting declarations** should be full sentences", so it applies only to that particular type of comments not to any Golang code. The two examples provided are comments for a type and a function declaration. And that makes sense because those comments will be extracted into the godoc documentation.

Broadening the scope of this recommendation to all comments is questionable as those comments won't be exported. I'm not against writing full sentences for those other type of comments but I didn't want a linter rule to enforce that either (I know now that it's not the case). If you prefer making all comments a full sentence then the first letter needs to be capitalized to be correct (and look less strange). So, instead of:

`// flushes buffer to screen.`

It should be:

`// Flushes buffer to screen.`

**TL;DR**:
I think it's better to improve the quality of the comments before adding this type of rules that focus on details but as long as it is not enforced to all type of comments it's not a big deal for me if you want to have it.

PR #2321, opened 2024-08-06
Rename project to 'certsuite'

sebrandon1:
Changes and non-changes:
- All references to `test-network-function/cnf-certification-test` have been changed to `redhat-best-practices-for-k8s/certsuite` in preparation for the org and repo rename.  This includes all imports and docs links.
- References to _other_ repos that live under `test-network-function` have not been touched yet.  Such as `oct`, `privileged-daemonset`, `claim`, etc.


sebrandon1:
@ramperher Just a heads up, we're planning a rename of both our organization and the repos that live underneath it that may be disruptive if anyone is linking to the repo directly.  A lot of the renaming work has been done already, it's mostly just the names of the repos and the corresponding go imports that are left.

Our proposal, is that we will be changing the org:
`test-network-function` will be renamed to `k8s-best-practices`.

And the repos underneath it will be renamed as such:
`cnf-certification-test` --> `certsuite`
`test-network-function-claim` --> `claim`, `privileged-daemonset`, etc.

We'll also be pushing our images to a new quay.io organization as well: `k8s-best-practices`.

I marked this PR as a Proof-of-Concept so we'll be able to collaborate through the process.

PR #2352, opened 2024-08-16
Fix for the false positive of access-control's capabilities tcs.

greyerof:
The forbidden capabilities test cases in access-control test suite are giving false positives. The code was stringifying the whole "capabilities" field and searching for the capability in the resulting string, which led to false positives when the capability was set in the "drop" list, as we've seen in a partner.

The new code checks whether it was explicitly set in the Add list directly, matching by capability name or by "ALL".


tests/accesscontrol/suite.go
```diff
@@ -275,7 +295,7 @@ func checkForbiddenCapability(containers []*provider.Container, capability strin
 		switch {
 		case cut.SecurityContext == nil:
 		case cut.SecurityContext.Capabilities == nil:
-		case strings.Contains(cut.SecurityContext.Capabilities.String(), capability):
+		case isContainerCapabilitySet(cut.SecurityContext.Capabilities, capability):
```
edcdavid, line 298: The security context in the pod should be checked first and overridden by the security context in the container
see:
- https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-the-security-context-for-a-container
- https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-the-security-context-for-a-pod

greyerof: @edcdavid This is true for other fields of the securityContext, but not for this one. The capabilities can only be set at [container level](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.30/#securitycontext-v1-core), not at [pod level](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.30/#podsecuritycontext-v1-core).

tests/accesscontrol/suite_test.go
```diff
@@ -15,3 +15,112 @@
 // 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 
 package accesscontrol
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	v1 "k8s.io/api/core/v1"
```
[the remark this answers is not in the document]
greyerof: Done in [900e75b](https://github.com/redhat-best-practices-for-k8s/certsuite/pull/2352/commits/900e75b19beb3972202a423ad2b54a4c55179c05)

PR #2371, opened 2024-08-26
Return error string for SA lookup

sebrandon1:
We are seeing non-compliant pods (at least according to the test output) that have a blank reason for non-compliance.  This is the only place where a non-compliant object can come back with a blank reason string.


sebrandon1:
Closing in favor of: https://github.com/redhat-best-practices-for-k8s/certsuite/pull/2368

PR #2381, opened 2024-08-28
Fix OperatorAutomountTokens reversed result

edcdavid:
build-depends: https://github.com/dci-labs/dallas-pipelines/pull/1210

### DIFF
--- a/expected_results.yaml
+++ b/expected_results.yaml
@@ -55,7 +55,6 @@ testCases:
     - observability-crd-status
     - observability-pod-disruption-budget
     - observability-termination-policy
-    - operator-automount-tokens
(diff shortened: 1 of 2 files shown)


sebrandon1 (approved):

ramperher:
> from change #2381:
> 
> * FAILURE https://www.distributed-ci.io/jobs/825a1ca8-830a-48fd-a34e-df859eb7a067/jobStates

This is expected, in fact the test was passing before but it should be failing, so now the result is according to the behavior of the test. Let me add this as exception in our pipelines to avoid more noise here, so that you have a green job

ramperher (approved):

PR #2389, opened 2024-09-01
Add upload results-spreadsheet certsuite sub-command

shirmoran:
build-depends: https://github.com/redhatci/ansible-collection-redhatci-ocp/pull/421

`upload results-spreadsheet` certsuite sub command creates a `RedHat Best Practices for K8 Results Spreadsheet`, that is meant to help us keep track on workloads suite test results and document their conclusion.

The created spreadsheet contains:

1.  Given workloads raw results sheet.
2. Conclusions sheet to document information per-workload, including a link to a separate results spreadsheet for each workload (which is meant to be used by the tech lead\product manager of the specific workload).

The application is using the google sheet and drive APIs as well as a google cloud project - a project with permissions for the entire cert-and-platform-dev team was created.
In order to use the app, you'll have to download the credentials from the google project and move them to the `cmd/certsuite/upload/results_spreadsheet/credentials.json` in your certsuite cloned repo.

** Note: users who don't have permissions to the google cloud project, can use their own project and credentials (see [google sheet quick start](https://developers.google.com/sheets/api/quickstart/go))

**Usage:** `./certsuite upload results-spreadsheet -f <raw-results-file.csv> -d <dest-url> -v <ocp-version(optional)>`

Next Steps: Creating a private project to create an `RedHat Best Practices for K8 Results Spreadsheet` for specifically for operators - which will contain more specific aspects.


ramperher (commented):
Hi, is this a new, standard way of reporting the certsuite results for certification? If so, we'd be interested in knowing how it's working, thanks!

cmd/certsuite/upload/results_spreadsheet/sheet_utils.go
```diff
@@ -0,0 +1,161 @@
+package resultsspreadsheet
+
+import (
+	"fmt"
+
+	"google.golang.org/api/sheets/v4"
+)
+
+func getHeadersAsInterfaceList(sheet *sheets.Sheet) []interface{} {
```
greyerof, line 9: `getHeaders` is good enough, and you should document here why you need to return a `[]interface{}` instead of the straightforward `[]string`

cmd/certsuite/upload/results_spreadsheet/sheet_utils.go
```diff
@@ -0,0 +1,161 @@
...
+
+func getIndicesFromListByName(headers []interface{}, wantedColumnsNames []string) ([]int, error) {
+	indices := []int{}
+	for _, wantedColName := range wantedColumnsNames {
+		found := false
+		for i, val := range headers {
+			if wantedColName == val {
+				found = true
+				indices = append(indices, i)
+				continue
```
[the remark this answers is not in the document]
shirmoran: Suppose to be a break, right! thanks!

cmd/certsuite/upload/results_spreadsheet/results_spreadsheet.go
```diff
@@ -0,0 +1,367 @@
...
+	return rows
+}
+
+func createSingleWorkloadRawResultsSheet(rawResultsSheet *sheets.Sheet, workloadName string) (*sheets.Sheet, error) {
+	// Initialize sheet with headers
+	filteredRows := []*sheets.RowData{{Values: []*sheets.CellData{
+		{UserEnteredValue: &sheets.ExtendedValue{StringValue: stringToPointer("Owner/TechLead Conclusion")}},
+		{UserEnteredValue: &sheets.ExtendedValue{StringValue: stringToPointer("Next Step Actions")}},
+	}}}
+	filteredRows[0].Values = append(filteredRows[0].Values, rawResultsSheet.Data[0].RowData[0].Values...)
```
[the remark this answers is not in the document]
shirmoran: in line 142 I added a comment saying "Initialize sheet with headers", afterwards in line 156 I'm explaining the filter - "add to sheet only rows of given workload name".
But I get why it might be a bit confusing.. maybe I can initialize the headers in a different var and than add it to the `filteredRows`. What do you think?

cmd/certsuite/upload/results_spreadsheet/sheet_utils.go
shirmoran: I'll change this function to two different functions that will return `[]string`. Those functions will handle the different types that the headers need to be extracted from.

cmd/certsuite/upload/results_spreadsheet/results_spreadsheet.go
```diff
@@ -0,0 +1,367 @@
...
+				value = *rawResultsSheetrow.Values[operatorVersionColIndex].UserEnteredValue.StringValue
+
+			case ocpVersionConclusionsCol:
+				value = ocpVersion + " "
+
+			case workloadNameConclusionsCol:
+				value = workloadName
+
+			case resultsConclusionsCol:
+				workloadResultsSpreadsheet, err := createSingleWorkloadRawResultsSpreadSheet(sheetsService, driveService, workloadsResultsFolder, rawResultsSheet, workloadName)
```
edcdavid, line 258: Make workloadsResultsFolder a parameter of the createConclusionsSheet function

cmd/certsuite/upload/results_spreadsheet/results_spreadsheet.go
```diff
@@ -0,0 +1,367 @@
...
+	records, err := reader.ReadAll()
+	if err != nil {
+		return nil, err
+	}
+	return records, nil
+}
+
+func createSheetsAndDriveServices() (sheetService *sheets.Service, driveService *drive.Service, err error) {
+	ctx := context.Background()
+	b, err := os.ReadFile("cmd/certsuite/upload/results_spreadsheet/credentials.json")
```
edcdavid, line 87: Why not expect credentials.json and token.json in the same folder as the executable? Maybe optionally pass a credentials folder path?

edcdavid (approved):

cmd/certsuite/upload/results_spreadsheet/results_spreadsheet.go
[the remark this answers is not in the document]
greyerof: No, it's ok as it is now, and I finally understood it. I'd rather be a bit more verbose on the purpose of certain code blocks, like this one (lines 143-147). Also, I believe this functions should be properly docummented. If I'm not wrong, something like this would be fine:
```
// createSingleWorkloadRawResultsSheet creates a new spreadsheet with test case results of a single workload,
// extracted from rawResultsSheets (which may contain the results of several workloads).
// The spreadsheet will use the same header columns as the rawResultsSheet, but will also add two extra columns:
// - "Owner/TechLead Conclusion": the partner/user is expected to add the name of the workload owner that should lead the fix
//   of this test case result.
// - "Next Step Actions": the partner/user may use this column to add the follow-up actions to fix this test case result.
func createSingleWorkloadRawResultsSheet(rawResultsSheet *sheets.Sheet, workloadName string) (*sheets.Sheet, error) {
	// Initialize sheet with the two new column headers only.
	filteredRows := []*sheets.RowData{{Values: []*sheets.CellData{
		{UserEnteredValue: &sheets.ExtendedValue{StringValue: stringToPointer("Owner/TechLead Conclusion")}},
		{UserEnteredValue: &sheets.ExtendedValue{StringValue: stringToPointer("Next Step Actions")}},
	}}}

	// Add existing column headers from the rawResultsSheet
	filteredRows[0].Values = append(filteredRows[0].Values, rawResultsSheet.Data[0].RowData[0].Values...)
	...
```

cmd/certsuite/upload/results_spreadsheet/results_spreadsheet.go
```diff
@@ -0,0 +1,367 @@
...
+	headers := getHeadersFromSheet(rawResultsSheet)
+	indices, err := getHeaderIndicesByColumnNames(headers, []string{"CNFName"})
+	if err != nil {
+		return nil, err
+	}
+	workloadIndex := indices[0]
+
+	// add to sheet only rows of given workload name
+	for _, row := range rawResultsSheet.Data[0].RowData[1:] {
+		if len(row.Values) > workloadIndex {
```
greyerof, line 158: Suggestion: Recommended fast return if workload name is not found.
```
for _, row := range rawResultsSheet.Data[0].RowData[1:] {
  // short-cirtcuit exit if workload name not found
  if workloadIndex >= len(row.Values) {
    return nil, fmt.Errorf("Workload " + workloadName + " not found in raw spreadsheet.")
  }
  ...
```

cmd/certsuite/upload/results_spreadsheet/results_spreadsheet.go
```diff
@@ -0,0 +1,367 @@
...
+
+	headers := getHeadersFromSheet(rawResultsSheet)
+	indices, err := getHeaderIndicesByColumnNames(headers, []string{"CNFName"})
+	if err != nil {
+		return nil, err
+	}
+	workloadIndex := indices[0]
+
+	// add to sheet only rows of given workload name
+	for _, row := range rawResultsSheet.Data[0].RowData[1:] {
```
greyerof, line 157: Is `len(rawResultsSheet.Data[0].RowData) > 1` checked anywhere before calling this func? Otherwise it's a good pre-condition to add at the very begining of this one.

cmd/certsuite/upload/results_spreadsheet/results_spreadsheet.go
```diff
@@ -0,0 +1,367 @@
...
+
+	if err := MoveSpreadSheetToFolder(driveService, folder, workloadResultsSpreadsheet); err != nil {
+		return nil, err
+	}
+
+	return workloadResultsSpreadsheet, nil
+}
+
+//nolint:funlen
+func createConclusionsSheet(sheetsService *sheets.Service, driveService *drive.Service, rawResultsSheet *sheets.Sheet) (*sheets.Sheet, error) {
```
greyerof, line 210: Please document this function.

cmd/certsuite/upload/results_spreadsheet/results_spreadsheet.go
```diff
@@ -0,0 +1,367 @@
...
+	headers := getHeadersFromSheet(rawResultsSheet)
```
greyerof, line 211: Suggestion: `headers` --> `rawSheetHeaders`. Then this will help to distinguish it from the `conclusionSheetHeaders`.

cmd/certsuite/upload/results_spreadsheet/results_spreadsheet.go
```diff
@@ -0,0 +1,367 @@
...
+	// Initialize sheet with headers
+	conclusionsSheetRowsValues := []*sheets.CellData{}
+	for _, colHeader := range conclusionSheetHeaders {
+		headerCellData := &sheets.CellData{UserEnteredValue: &sheets.ExtendedValue{StringValue: stringToPointer(colHeader)}}
+		conclusionsSheetRowsValues = append(conclusionsSheetRowsValues, headerCellData)
+	}
+	conclusionsSheetRows := []*sheets.RowData{{Values: conclusionsSheetRowsValues}}
+
+	// Extract unique values from the CNFName column and fill sheet
+	uniqueMap := make(map[string]bool)
```
greyerof, line 230: Suggestion: `uniqueMap` --> `uniqueWorkloadNames`

cmd/certsuite/upload/results_spreadsheet/results_spreadsheet.go
```diff
@@ -0,0 +1,367 @@
...
+	for _, rawResultsSheetrow := range rawResultsSheet.Data[0].RowData[1:] {
```
greyerof, line 231: Same comment here as at line 157: `len(rawResultsSheet.Data[0].RowData)` should be checked to be `> 1`. Otherwise, it must be explicitly stated in the function documentation that the caller is responsible to check it before calling this func.

cmd/certsuite/upload/results_spreadsheet/results_spreadsheet.go
```diff
@@ -0,0 +1,367 @@
...
+
+				hyperlinkFormula := fmt.Sprintf("=HYPERLINK(%q, %q)", workloadResultsSpreadsheet.SpreadsheetUrl, "Results")
+				curCellData.UserEnteredValue.FormulaValue = &hyperlinkFormula
+
+			default:
+				// use space for empty values to avoid cells overlapping
+				value = " "
+			}
+
+			if colHeader != resultsConclusionsCol {
```
greyerof, line 271: It is worth having a comment here on why this column/cell is filtered now and when it's going to be set.

greyerof (approved):
Great work. Mostly suggestions comments and function documentation requests.

cmd/certsuite/upload/results_spreadsheet/results_spreadsheet.go
shirmoran: I agree, I will add this check in the `createConclusionSheet` func, and as the `createSingleWorkloadRawResultsSheet` is called only from there, the check will sufficient for both cases.

cmd/certsuite/upload/results_spreadsheet/results_spreadsheet.go
shirmoran: The token is created if does not exists, but the credentials can be passed if we want. I though it might be better to store the credentials and the token in the same folder as the sub-command to give it more context, but they can be moved to same folder as the executable.
What do you thinks?

cmd/certsuite/upload/results_spreadsheet/results_spreadsheet.go
greyerof: Now I see I was wrong here, since the value was already set as a formula value. Sorry for the noise :(

shirmoran: though I think my change made it clearer, take a look: https://github.com/redhat-best-practices-for-k8s/certsuite/blob/9ad972f30706b8c6a4dadafc457166dd1eeaa0d4/cmd/certsuite/upload/results_spreadsheet/results_spreadsheet.go#L260

cmd/certsuite/upload/results_spreadsheet/results_spreadsheet.go
edcdavid: I would prefer to use same folder as the executable or a configured folder optionally. The folder of the subcommand does not have a meaning when running the executable, it is just relevant at build time to organize the code. So if running the exe outside of our repo it would require a very specific directory to be created.

cmd/certsuite/upload/results_spreadsheet/drive_utils.go
```diff
@@ -0,0 +1,78 @@
...
+	return nil
+}
+
+func extractFolderIDFromURL(u string) (string, error) {
+	parsedURL, err := url.Parse(u)
+	if err != nil {
+		return "", err
+	}
+
+	pathSegments := strings.Split(parsedURL.Path, "/")
```
sebrandon1, line 74: Should there be a safeguard here after the split?

cmd/certsuite/upload/results_spreadsheet/results_spreadsheet.go
```diff
@@ -0,0 +1,372 @@
...
+
+	rawResultsSheet := &sheets.Sheet{
+		Properties: &sheets.SheetProperties{
+			Title:          "raw results",
+			GridProperties: &sheets.GridProperties{FrozenRowCount: 1},
+		},
+		Data: []*sheets.GridData{{RowData: rows}},
+	}
+
+	return rawResultsSheet, err
```
sebrandon1, line 318:
```suggestion
	return rawResultsSheet, nil
```

sebrandon1 (commented):
Some more comments

cmd/certsuite/upload/results_spreadsheet/drive_utils.go
shirmoran: you mean checking the len for the list? probably should

sebrandon1 (approved):

PR #2390, opened 2024-09-01
Avoid failing workflow when collectors sanity check fails

shirmoran:
This PR avoids failing `pre-main` workflow when collector's sanity check fails - instead it will only send a warning message to the our slack channel.


greyerof (approved):
This should (hopefully) work once it's merged.

PR #2417, opened 2024-09-10
Remove logrus as direct dependency

sebrandon1:
I noticed that we were only using `logrus` as a direct dependency in 3 lines.  By switching these lines over to the `log` package we can remove logrus as a direct dependency.

### DIFF
--- a/cmd/certsuite/generate/catalog/catalog.go
+++ b/cmd/certsuite/generate/catalog/catalog.go
@@ -27,7 +27,6 @@ import (
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/artifacts"
 	plibContainer "github.com/redhat-openshift-ecosystem/openshift-preflight/container"
 	plibOperator "github.com/redhat-openshift-ecosystem/openshift-preflight/operator"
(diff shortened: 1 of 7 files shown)


rdavid (approved):

cmd/certsuite/generate/catalog/catalog.go
```diff
@@ -138,7 +137,7 @@ func addPreflightTestsToCatalog() {
 	// Create artifacts handler
 	artifactsWriter, err := artifacts.NewMapWriter()
 	if err != nil {
-		logrus.Errorf("error creating artifact, failed to add preflight tests to catalog")
+		log.Error("error creating artifact, failed to add preflight tests to catalog")
```
greyerof, line 140: Please capitalize the first word's letter in the message. Also, include the error in the message like:
```
log.Error("Error creating artifacts catalog writer for preflight tests: %v", err)
```

Same for the other changes.

sebrandon1: Done.  I found some other instances of it as well.

greyerof (approved):

PR #2435, opened 2024-09-13
CRC Testing 3

sebrandon1:
Based on old changes from previous attempts: https://github.com/redhat-best-practices-for-k8s/certsuite/blob/cc447b17bb843f7d1ca238688621b648ae7e121d/.github/workflows/qe-crc.yml


sebrandon1:
Just entirely too slow still.  Not enough resources to both the environment and the tests at the same time.  Maybe someday there will be either a slimmer version of OCP or a beefier version of the github actions runners free tier.

PR #2443, opened 2024-09-16
Add Setter and Getter functions to results spreadsheet google creds

shirmoran:
This PR adds `credentials` as an argument of  `CreateSheetsAndDriveServices` func  in order to allow the use of it independently from using `upload results-spreadsheet` sub-command.
In current state, the credentials are sent using a flag to the mentioned sub command, which limits functions using this variable to run using the sub-command only.
In the `operator-results-spreadsheet` repo, we would like to allow the use of `CreateSheetsAndDriveServices` func (which uses the `credentials` variable) in order to add more columns to an existing sheet, and for that we need the credentials to be set manually.

### DIFF
--- a/cmd/certsuite/upload/results_spreadsheet/results_spreadsheet.go
+++ b/cmd/certsuite/upload/results_spreadsheet/results_spreadsheet.go
@@ -58,6 +58,14 @@ func NewCommand() *cobra.Command {
 	return uploadResultSpreadSheetCmd
 }
 
+func SetCredentials(credenitalsPath string) {
+	credentials = credenitalsPath
+}
+
+func GetCredentials() string {
+	return credentials
+}
+
 func readCSV(fp string) ([][]string, error) {
 	file, err := os.Open(fp)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


cmd/certsuite/upload/results_spreadsheet/results_spreadsheet.go
```diff
@@ -58,6 +58,14 @@ func NewCommand() *cobra.Command {
 	return uploadResultSpreadSheetCmd
 }
 
+func SetCredentials(credenitalsPath string) {
+	credentials = credenitalsPath
```
sebrandon1, line 62:
```suggestion
	credentials = credentialsPath
```

greyerof (approved):
This is a way to solve it, but it doesn't look like a good approach to me, as the setters/getters use an internal global/package var. I would probably change the `CreateSheetsAndDriveServices` to add an extra string argument for that credentails path.

shirmoran:
> This is a way to solve it, but it doesn't look like a good approach to me, as the setters/getters use an internal global/package var. I would probably change the `CreateSheetsAndDriveServices` to add an extra string argument for that credentails path.

I agree - That's a better idea. I'll do the change, thanks!

greyerof (approved):

sebrandon1 (approved):

PR #2497, opened 2024-10-07
Authenticate results spreadsheet creation using a google service account

shirmoran:
This PR changes the authentication method of the results spreadsheet (created by certsuite's executable `upload results-spreadsheet` sub command)

Up until now, the authentication process used a google account. The user running the app had to manually allow google cloud project to access the google account (than the app continued with the authorization) .

Now, the authentication process will use a google service account, while using it's private key as the credentials file (can be created from the google cloud project). Using this approach we can avoid the manual "allow" operation and the authcode and token exchange.


greyerof (approved):

edcdavid (approved):
I am getting a warning that drive is shared with a user outside of the organization.

sebrandon1 (approved):

PR #2508, opened 2024-10-09
operator: add olm.skipRange annotation test

sebrandon1:
Adds a test looking for the existence of the olm.skipRange annotation on the CSV.

Note: We will have to switch the expected result of this test to `skipped` when Hazelcast operator is removed.

build-depends: https://github.com/dci-labs/dallas-pipelines/pull/1245

### DIFF
--- a/CATALOG.md
+++ b/CATALOG.md
@@ -7,7 +7,7 @@ Depending on the workload type, not all tests are required to pass to satisfy be
 
 ## Test cases summary
 
-### Total test cases: 113
+### Total test cases: 114
 
 ### Total suites: 10
 
@@ -19,7 +19,7 @@ Depending on the workload type, not all tests are required to pass to satisfy be
 |manageability|2|
 |networking|11|
 |observability|5|
-|operator|8|
+|operator|9|
 |performance|6|
 |platform-alteration|13|
 |preflight|17|
@@ -36,11 +36,11 @@ Depending on the workload type, not all tests are required to pass to satisfy be
 |---|---|
 |7|1|
 
-### Non-Telco specific tests only: 66
+### Non-Telco specific tests only: 67
 
 |Mandatory|Optional|
 |---|---|
-|42|24|
+|42|25|
 
 ### Telco specific tests only: 27
 
@@ -1249,6 +1249,22 @@ Tags|common,operator
 |Far-Edge|Mandatory|
 |Non-Telco|Mandatory|
 |Telco|Mandatory|
+
+#### operator-olm-skip-range
+
+Property|Description
+---|---
+Unique ID|operator-olm-skip-range
+Description|Test that checks the operator has a valid olm skip range.
+Suggested Remediation|Ensure that the Operator has a valid OLM skip range.
+Best Practice Reference|https://redhat-best-practices-for-k8s.github.io/guide/#redhat-best-practices-for-k8s-cnf-operator-requirements
+Exception Process|No exceptions
+Tags|common,operator
+|**Scenario**|**Optional/Mandatory**|
+|Extended|Optional|
+|Far-Edge|Optional|
+|Non-Telco|Optional|
+|Telco|Optional|
 
 #### operator-pods-no-hugepages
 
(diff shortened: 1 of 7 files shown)


sebrandon1:
Tested against OCP 4.17 cluster with nginx-ingress-operator:

Prior to adding the annotation:

```
---------------------------------
| LOG (operator-olm-skip-range) |
---------------------------------
[INFO] [Oct 10 11:42:21.492] [check.go: 270] [operator-olm-skip-range] Running check (labels: [common operator-olm-skip-range operator])
[INFO] [Oct 10 11:42:21.492] [suite.go: 481] [operator-olm-skip-range] Testing Operator "csv: nginx-ingress-operator.v2.4.0 ns:brandontest subscription:nginx-ingress-operator targetNamespaces=[]"
[ERROR] [Oct 10 11:42:21.492] [suite.go: 484] [operator-olm-skip-range] OLM skipRange not found for Operator "csv: nginx-ingress-operator.v2.4.0 ns:brandontest subscription:nginx-ingress-operator targetNamespaces=[]"
[INFO] [Oct 10 11:42:21.492] [checksdb.go: 115] [operator-olm-skip-range] Recording result "FAILED", claimID: {Id:operator-olm-skip-range Suite:operator Tags:common}
```
After the annotation is added:

```
$ oc annotate csv nginx-ingress-operator.v2.4.0 -n brandontest olm.skipRange=v0.0.1
clusterserviceversion.operators.coreos.com/nginx-ingress-operator.v2.4.0 annotated
```

```
Running suite OPERATOR
[ PASS ] operator-install-status-succeeded
[ FAIL ] operator-install-status-no-privileges
[ PASS ] operator-install-source
[ PASS ] operator-semantic-versioning
[ SKIP ] operator-crd-versioning  (no operator crds found)
[ SKIP ] operator-crd-openapi-schema  (no operator crds found)
[ PASS ] operator-single-crd-owner
[ SKIP ] operator-run-as-user-id  (compliant and non-compliant objects lists are empty)
[ SKIP ] operator-run-as-non-root  (compliant and non-compliant objects lists are empty)
[ SKIP ] operator-automount-tokens  (compliant and non-compliant objects lists are empty)
[ SKIP ] operator-read-only-file-system  (compliant and non-compliant objects lists are empty)
[ PASS ] operator-olm-skip-range
```

sebrandon1:
@ramperher @manurodriguez We are looking to add new `operator` suite tests and this one will probably fail your DCI run for mongo operator.

manurodriguez:
@sebrandon1 Thanks for the heads up, I see exactly that test failed: `testcase': 'operator-olm-skip-range', 'passed': False` 
Let me know if you want me to skip that validation and I can prepare a PR as a dependency.

sebrandon1:
@manurodriguez That looks right to me.  It's an "Optional" test case but more than likely it is going to fail a lot of operators.

manurodriguez:
@sebrandon1 I added https://github.com/dci-labs/dallas-pipelines/pull/1245 as a dependency to ignore that test in our pipelines

/dci-rerun

greyerof (approved):
Let's see what happens, but I think this implementation might lead to too many operators failing this check (as hazelcast does). I know it's an optional test case but shouldn't it have an exception process for operators that (a) have one version only or (b) have good (compatibility?) reasons for not having an olm.skipRange.

sebrandon1:
We have had other tests in this same type of scenario where they might start failing everything such as the operator-run-as-non-root test (that's being name-changed in #2479).  We were expecting that one to fail quite a few operators when it was merged.

greyerof:
@sebrandon1 but this is different. If the operator controller's container runs as root, that's their fault, and it's a legit positive. If there's only one version available of that operator (the initial/first one) and we flag this tc as failed... that's our fault, it's a false positive, because there's no way (and it makes no sense) they can add that olm.skipRange yet.

ramperher:
@sebrandon1 , the PR cannot be tested because there are conflicting files comparing with main branch that need to be fixed prior to start testing this on DCI

PR #2514, opened 2024-10-11
networking: Add MTU value test for SRIOV NAD's

sebrandon1:
Adds a test that checks that Network Attachment Definitions are specifying their MTU values explicitly.

EDIT: I was trying to follow the quick start guide [here](https://github.com/openshift/sriov-network-operator/blob/master/doc/quickstart.md) for the SRIOV network operator but I don't have access to the necessary [hardware](https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/doc/supported-hardware.md).

We'll have to maybe figure out how to test this via QE on supported hardware in the future.

### DIFF
--- a/pkg/provider/pods.go
+++ b/pkg/provider/pods.go
@@ -266,6 +266,61 @@ func getCNCFNetworksNamesFromPodAnnotation(networksAnnotation string) []string {
 	return networkNames
 }
 
+// isNetworkAttachmentDefinitionSRIOVConfigMTUSet is a helper function to check whether a CNI config
+// string has any config for MTU for SRIOV configs only
+
+/*
+	{
+		"cniVersion": "0.4.0",
+		"name": "vlan-100",
+		"plugins": [
+			{
+				"type": "sriov",
+				"master": "ext0",
+				"mtu": 1500,
+				"vlanId": 100,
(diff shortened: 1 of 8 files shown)


pkg/provider/pods.go
```diff
@@ -265,6 +265,61 @@ func getCNCFNetworksNamesFromPodAnnotation(networksAnnotation string) []string {
 	return networkNames
 }
 
+// isNetworkAttachmentDefinitionSRIOVConfigMTUSet is a helper function to check whether a CNI config
```
bnshr, line 268: Remove this commented code.

greyerof (approved):

pkg/provider/pods.go
sebrandon1: Actually this was on purpose.  This is also done on other functions in this `pods.go` file to show examples of the JSON that is being parsed.

bnshr (approved):

bnshr:
Should there be a QE test for this which can be skipped based on the unavailability of the hardware?

sebrandon1:
@bnshr Yes, that would be the goal for this test I'd say.  We'll need some sort of hardware environment to test it.

PR #2551, opened 2024-11-07
Get subscription channel from package manifest if it is not present

edcdavid:
Depends-On: https://github.com/redhat-best-practices-for-k8s/oct/pull/221
When creating a subscription using a default channel, the "spec->channel" parameter is not required. However we need this parameter to perform the certification test for operators.
When the spec->channel parameter is empty, get the package manifest corresponding to the subscription, retrieve the default channel and use it for the operator certification test

Update: the channel is not an attribute of the operator but of the catalog, so I believe we can check certification without it. If an user creates a custom catalog with a channel name not matching the one in the catalog this test would fail otherwise when it should not.
Leaving the packageManifest discovery code in since it could be needed in the future


ramperher:
As you may see in the change, the failure reason is that the certsuite container image is not correctly built

edcdavid:
> As you may see in the change, the failure reason is that the certsuite container image is not correctly built, this probably needs to be reviewed on your side to make it work.

@ramperher thanks for the comment. I'll fix it

bnshr (approved):

pkg/autodiscover/autodiscover_operators.go
```diff
@@ -39,6 +44,34 @@ const (
 	istioDeploymentName = "istiod"
 )
 
+type PackageManifest struct {
```
greyerof, line 47: Since we're already importing the operator-lifecycle-manager repo, you could use the official type for the package manifest [here](https://github.com/operator-framework/operator-lifecycle-manager/blob/b3926787314dfbc750d84b83c754510b30d78596/pkg/package-server/apis/operators/v1/packagemanifest_types.go#L22). Maybe there's also a client/schema to be used to avoid the need of the dynamic client.

edcdavid: Cool Thx! I was looking for this but could not find it. I will update

edcdavid: Fixed

PR #2552, opened 2024-11-08
Remove unused .tar.gz artifacts from pre-main

sebrandon1:
These artifacts are never used and can be removed/no longer cached every run.

### DIFF
--- a/.github/workflows/pre-main.yaml
+++ b/.github/workflows/pre-main.yaml
@@ -239,30 +239,11 @@ jobs:
       - name: 'Show failed test cases and the non compliant objects'
         run: ./certsuite claim show failures -c certsuite-out/claim.json
 
-      - name: Upload smoke test results as an artifact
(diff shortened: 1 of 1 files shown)


.github/workflows/pre-main.yaml
```diff
@@ -239,31 +239,12 @@ jobs:
       - name: 'Show failed test cases and the non compliant objects'
         run: ./certsuite claim show failures -c certsuite-out/claim.json
 
-      - name: Upload smoke test results as an artifact
```
edcdavid, line 242: Are there any issues with keeping these? I used these artifacts several times in the past to debug issues or obtain recent claim files.

sebrandon1: I didn't know anyone was still using them.  We can keep them if you want.

sebrandon1:
Closing as these are still used.

PR #2577, opened 2024-11-20
Operand pods autodiscovery

greyerof:
This change adds an autodiscovery mechanism for operand pods that "belong" to CSVs/Operators under test. This is a separate and complementary way to the by-label/by-namespace already existing discovery modes.

Usually, operands are deployed by operators by means of a CR. Ideally, operators designers should also flag that CR as owner of those operand pods that are deployed in the cluster.

The autodiscovery mechanism for operand pods creates a list of CRDs owned by the CSVs under test. Then, it iterates over every pod found in the configured namespaces in order to get whether its owner is a CR of any of those CRD types. If so, it's added to the list of pods under test if it wasn't already there.

### DIFF
--- a/pkg/autodiscover/autodiscover.go
+++ b/pkg/autodiscover/autodiscover.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"path"
 	"regexp"
 	"strings"
 	"time"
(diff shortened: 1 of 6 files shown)


pkg/autodiscover/autodiscover.go
```diff
@@ -355,3 +367,61 @@ func getPodsOwnedByCsv(csvName, operatorNamespace string, client *clientsholder.
...
+		if len(ownedCrds) == 0 {
+			continue
+		}
+
+		for i := range ownedCrds {
+			crd := &ownedCrds[i]
+
+			_, group, found := strings.Cut(crd.Name, ".")
+			if !found {
+				return nil, fmt.Errorf("failed to parse resources and group form crd name %q", crd.Name)
```
edcdavid, line 391: typo form-> from

edcdavid (approved):

edcdavid:
tested with prometheus operator

bnshr (approved):

sebrandon1 (approved):

PR #2589, opened 2024-11-25
feat : Valid (Single/Multi-Namespaced) Operator Installation in tenant namespace

bnshr:
Refers [the issue](https://issues.redhat.com/browse/CNFCERT-1088)


tests/operator/suite.go
```diff
@@ -115,9 +116,69 @@ func LoadChecks() {
...
+		operatorTargetNamespaces := strings.Split(targetNamespacesStr, ",")
+		operatorGroup := findOperatorGroup(csv.Annotations["olm.operatorGroup"], operator.Namespace, env.OperatorGroups)
+
+		check.LogInfo("operatorNamespace %s, targetNamespaces %v, operatorGroup %s", operatorNamespace,
+			operatorTargetNamespaces, operatorGroup.Name)
+
+		isSingleNamespaceInstallMode := isInstallModeSingleNamespace(csv.Spec.InstallModes)
+		isCompliant := checkOperatorInstallationCompliance(operatorGroup, operatorNamespace, operatorTargetNamespaces, isSingleNamespaceInstallMode)
+
+		message := "Operator with%s is %sinstalled in tenant namespace "
```
sebrandon1, line 158:
```suggestion
		message := "Operator with %s is %s installed in tenant namespace "
```

tests/operator/helper.go
```diff
@@ -45,3 +50,34 @@ func SplitCsv(csv string) CsvResult {
...
+	opGroupTargetNamespaces := operatorGroup.Spec.TargetNamespaces
+	if isSingleNamespaceInstallMode {
+		return len(opGroupTargetNamespaces) == 1 && len(targetNamespaces) == 1 && opGroupTargetNamespaces[0] == targetNamespaces[0]
+	}
+	for _, ns := range opGroupTargetNamespaces {
+		if ns == operatorNamespace {
+			return false
+		}
+	}
+	return true
```
sebrandon1, line 82: Is `opGroupTargetNamespaces` a []string?

Then we could use `stringhelper.StringInSlice` and remove this loop.

tests/operator/helper.go
```diff
@@ -20,7 +20,12 @@ Package operator provides CNFCERT tests used to validate operator CNF facets.
 
 package operator
 
-import "strings"
+import (
+	"strings"
+
+	v1 "github.com/operator-framework/api/pkg/operators/v1"
```
sebrandon1, line 26:
```suggestion
	operatorsv1 "github.com/operator-framework/api/pkg/operators/v1"
```

Something to make `v1` more explicit as an import.

tests/operator/helper.go
```diff
@@ -20,7 +20,12 @@ Package operator provides CNFCERT tests used to validate operator CNF facets.
...
+	"github.com/operator-framework/api/pkg/operators/v1alpha1"
```
sebrandon1, line 27: Do the `InstallMode` variables not exist in `v1`?  That way we don't have multiple imports?

tests/operator/suite.go
sebrandon1: Instead of having `not ` with a space, for example.

sebrandon1 (commented):
Initial set of comments

tests/operator/helper.go
bnshr: Yes, []string. 
I don't think we need this function. The definition uses Contains function, which is like substring check, whereas we need here Equals check on strings. StringInSlice would be true for namespace abc if the slice contains xyzabcd.

sebrandon1: Just set the `contains` parameter to false for StringInSlice.

tests/operator/helper.go
bnshr: No, InstallMode does not exist.

tests/operator/suite.go
```diff
@@ -115,9 +115,62 @@ func LoadChecks() {
...
+
+	for _, operator := range env.Operators {
+		check.LogInfo("Checking operator %s in namespace %s ", operator.Name, operator.Namespace)
+
+		csv := operator.Csv
+		isSingleNamespaceInstallModeSupported := hasOperatorInstallModeSingleNamespace(csv.Spec.InstallModes)
+
+		csvNamespace := csv.Namespace
+		operatorNamespace := csv.Annotations["olm.operatorNamespace"]
+		targetNamespaces := operator.TargetNamespaces
```
greyerof, line 151: Maybe I'm wrong but, to me, "tenant app namespaces" are the ones declared in certsuite_config.yaml (field [targetNamespaces](https://github.com/redhat-best-practices-for-k8s/certsuite/blob/9718894c4bac4fd05a0acc35b4b5adb17d2fac18/config/certsuite_config.yml#L1)), not the operator.TargetNamespaces (which comes from [operatorGroup.spec.targetNamespaces](https://github.com/redhat-best-practices-for-k8s/certsuite/blob/9718894c4bac4fd05a0acc35b4b5adb17d2fac18/pkg/provider/operators.go#L192)).

So isn't it good enough just checking whether the operator installation namespace (operator.Csv.Annotations["olm.operatorNamespace"]) is in the certsuite_config.yaml's "targetNamespaces" field?

bnshr: As per the requirement, the tenant dedicated **operator** namespace is mentioned by the namespace where CR is allowed to be created. In any case, we are not considering any other namespace other than the config file specified namespaces for our testing. But the check on targetNamespace is better to be done in the operator specific way, instead of going back to our configuration.

edcdavid: The requirement for this test is:
```
Operators are not permitted to be installed into a tenant app namespace. Operators that are installed with the "Single Namespace" olm install mode must be installed into the tenants dedicated operator namespace. Upstream 3rd party operators will be installed Globally with "All Namespaces" olm install mode into the openshift-operators namespace.
```
In my understanding, the idea of this test is to make sure that tenants dedicate a special namespace for installing operator(s) and that it is not shared with other CNF application namespaces. So I would expect the dedicated operator namespace to be part of the certsuite_config.yaml's "targetNamespaces" filed as @greyerof  mentioned, since this hold the namespaces under test. Also, this reserved namespace should only be holding resources deployed/owned by the operator. So if non-operator pods are found in a namespace containing operator pods/resources, the test should fail.

tests/operator/suite.go
```diff
@@ -120,15 +116,52 @@ func LoadChecks() {
...
+		doesNamespaceContainsOnlyOperatorPods, err := checkIfNamespaceContainsOnlyOperatorPods(operatorNamespace)
+
+		if err != nil {
+			check.LogError("Skipped - cannot proceed with operator %s", operator.Name)
+			continue
+		}
+		if doesNamespaceContainsOnlyOperatorPods {
+			check.LogInfo("Operator %s has valid installation in tenant namespace %s ", operator.Name, operatorNamespace)
+			compliantObjects = append(compliantObjects, testhelper.NewOperatorReportObject(operator.Namespace, operator.Name,
+				"Operator has valid installation in tenant namespace ", true).AddField(testhelper.OperatorName, operator.Name))
```
greyerof, line 154: The operator's name is lready set in the NewOperatorReportObject() call. Is that `AddField()` with the name field actually needed?

tests/operator/suite.go
```diff
@@ -449,101 +472,3 @@ func testMultipleSameOperators(check *checksdb.Check, env *provider.TestEnvironm
 
 	check.SetResult(compliantObjects, nonCompliantObjects)
 }
-
-//nolint:funlen
-func testOperatorCatalogSourceBundleCount(check *checksdb.Check, env *provider.TestEnvironment) {
```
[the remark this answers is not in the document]
bnshr: Ah, may be caused by mistake in merge conflict resolution. Thanks!

tests/operator/helper.go
```diff
@@ -45,3 +57,62 @@ func SplitCsv(csv string) CsvResult {
...
+			if owner.Kind == v1alpha1.ClusterServiceVersionKind && owner.Namespace == namespace {
+				foundOperatorPods++
+				foundCsvs[owner.Name] = true
+				break
+			}
+		}
+	}
+
+	// Check if the found CSVs contain only valid operators under test
+	status := false
```
edcdavid, line 108: `status` seems to be true if one operator matches only one found csv and is not clusterwide. Instead, all found operators that are under test should not be clusterwide, something like:


```
    AllOperatorsFoundInNamespaceAreValid := true
    // report an error only if an operator under test is found to be clusterwide
     for _, operator := range env.Operators {
          for foundCsv := range foundCsvs {
			if operator.IsClusterWide && operator.Csv.Name == foundCsv {
				AllOperatorsFoundInNamespaceAreValid = false
                                  break
			}
		}
	}
```

tests/operator/helper.go
```diff
@@ -45,3 +57,62 @@ func SplitCsv(csv string) CsvResult {
...
+			singleNamespacedOperators = append(singleNamespacedOperators, operator)
+		}
+	}
+	return singleNamespacedOperators
+}
+
+// This function checks if the namespace contains only valid operator pods
+func checkIfNamespaceContainsOnlyOperatorPods(namespace string) (isOperatorOnlyNamespace bool, err error) {
+	// Get all pods from the target namespace
+	podsList, err := clientsholder.GetClientsHolder().K8sClient.CoreV1().Pods(namespace).List(context.TODO(), metav1.ListOptions{})
```
edcdavid, line 82: why not use the env->Pods list instead and filter the namespace?

tests/operator/suite.go
```diff
@@ -126,6 +126,46 @@ func LoadChecks() {
...
+
+		doesNamespaceContainsOnlyOperatorPods, err := checkIfNamespaceContainsOnlyOperatorPods(operatorNamespace)
+
+		if err != nil {
+			check.LogError("Skipped - cannot proceed with operator %s", operator.Name)
+			continue
+		}
+		if doesNamespaceContainsOnlyOperatorPods {
+			check.LogInfo("Operator %s has valid installation in tenant namespace %s ", operator.Name, operatorNamespace)
+			compliantObjects = append(compliantObjects, testhelper.NewOperatorReportObject(operator.Namespace, operator.Name,
```
edcdavid, line 160: Shouldn't the namespace be the compliance object? If a namespace has a mix of pods belonging to operators and not belonging, then the test fail for that namespace. We could also list the pod not belonging to operators under test and are responsible for the test failing and the list of operators belonging to the namespace.

tests/operator/helper.go
bnshr: That's because there could be some pods created by others running in the operator namespace which are not test pods. env.Pods would give only test pods, but we need to check for all pods in a namespace.

edcdavid: Should we really test all pods? If some pods need to be skipped because they do not belong to the cnf under test (like some specific cluster scenario), we can currently skip them by setting labels.  This looks like a similar scenario to me. By default, if there is no labels configured for pods, all pods in the namespace are considered anyways.

sebrandon1: We also already gather all pods in `env.AllPods` and we could just filter the namespace as well for David's first comment.

tests/operator/suite.go
```diff
@@ -126,6 +126,53 @@ func LoadChecks() {
...
+	var compliantObjects []*testhelper.ReportObject
+	var nonCompliantObjects []*testhelper.ReportObject
+
+	singleNamespacedOperators := filterSingleNamespacedOperatorUnderTest(env.Operators)
+
+	// namespaces to be considered
+	var namespaces []string
+	for i := range singleNamespacedOperators {
+		operator := singleNamespacedOperators[i]
+		namespaces = append(namespaces, operator.Namespace)
```
edcdavid, line 151: The list of namespaces under test should be the list of namespaces in which operator under test are installed. If an operator is clusterwide and  installed in the namespace, it should be part of the non-compliant logs (see: if operator.IsClusterWide && operator.Csv.Name == foundCsv)

bnshr: I do not think so, this test case considers only operators under test which are installed with single name space. So, may be we need a new skip method. Wdyt?

bnshr: @sebrandon1

sebrandon1: I'm okay with a specific test for 3rd party operators so we can be more explicit about which 3rd party operators are outside of `openshift-operators` namespace.

tests/operator/suite.go
```diff
@@ -126,6 +126,53 @@ func LoadChecks() {
...
+
+		if doesNamespaceContainsOnlyOperatorPods {
+			for _, csv := range csvs {
+				check.LogInfo("Csv %s has valid installation in tenant namespace %s ", csv, operatorNamespace)
+				compliantObjects = append(compliantObjects, testhelper.NewOperatorReportObject(operatorNamespace, csv,
+					"Operator has valid installation in tenant namespace ", true))
+			}
+		} else {
+			check.LogInfo("Operator namespace %s contains mixed invalid operator pods ", operatorNamespace)
+			nonCompliantObjects = append(nonCompliantObjects, testhelper.NewReportObject(operatorNamespace,
```
greyerof, line 171: Why not using `testhelper.NewNamespacedReportObject` instead?

bnshr: Updated!

config/certsuite_config.yml
```diff
@@ -1,11 +1,15 @@
 targetNameSpaces:
   - name: tnf
+  - name: certsuite-operator
+  - name: nginx-ops
+  - name: nginx-apps
```
sebrandon1, line 5: Should these be removed from the default config?

Makefile
```diff
@@ -83,7 +83,7 @@ lint:
 	typos
 	markdownlint '**/*.md'
 	yamllint --no-warnings .
-	shellcheck --format=gcc ${BASH_SCRIPTS}
+	shellcheck --nocomments --format=gcc ${BASH_SCRIPTS}
```
sebrandon1, line 86: Could this be put into another PR?

tests/operator/helper.go
```diff
@@ -45,3 +52,83 @@ func SplitCsv(csv string) CsvResult {
 	}
 	return result
 }
+
+func hasOperatorInstallModeType(installModes []v1alpha1.InstallMode, installModeType v1alpha1.InstallModeType) bool {
```
sebrandon1, line 56: This could be unit tested in another PR.

tests/operator/helper.go
```diff
@@ -45,3 +52,83 @@ func SplitCsv(csv string) CsvResult {
...
+	for i := 0; i < len(installModes); i++ {
+		if installModes[i].Type == installModeType && installModes[i].Supported {
+			return true
+		}
+	}
+	return false
+}
+
+func filterSingleNamespacedOperatorUnderTest(operators []*provider.Operator) (singleNamespacedOperators []*provider.Operator) {
```
sebrandon1, line 65: This could be unit tested in another PR.

tests/operator/helper.go
```diff
@@ -45,3 +52,83 @@ func SplitCsv(csv string) CsvResult {
...
+	for _, operator := range operators {
+		if hasOperatorInstallModeType(operator.Csv.Spec.InstallModes, v1alpha1.InstallModeTypeSingleNamespace) && len(operator.TargetNamespaces) == 1 {
+			singleNamespacedOperators = append(singleNamespacedOperators, operator)
+		}
+	}
+	return singleNamespacedOperators
+}
+
+func getAllPodsBy(namespace string) (podsInNamespace []*provider.Pod) {
```
sebrandon1, line 74: This could be unit tested in another PR.

tests/operator/helper.go
```diff
@@ -45,3 +52,83 @@ func SplitCsv(csv string) CsvResult {
...
+		pod := env.AllPods[i]
+		if pod.Namespace == namespace {
+			podsInNamespace = append(podsInNamespace, pod)
+		}
+	}
+	return podsInNamespace
+}
+
+// This function checks if the namespace contains only valid operator pods
+func checkIfNamespaceContainsOnlyOperatorPods(namespace string) (isOperatorOnlyNamespace bool, csvs []string, failedPods string, err error) {
```
sebrandon1, line 85: This could be unit tested in another PR.

config/certsuite_config.yml
```diff
@@ -1,11 +1,15 @@
...
 podsUnderTestLabels:
   - "redhat-best-practices-for-k8s.com/generic: target"
 operatorsUnderTestLabels:
   - "redhat-best-practices-for-k8s.com/operator:target"
   - "redhat-best-practices-for-k8s.com/operator1:new"
   - "cnf/test:cr-scale-operator"
+  - "operators.coreos.com/nginx-ingress-operator.nginx-ops:"
```
sebrandon1, line 12: Does this also need to be removed before merging?

sebrandon1 (commented):
Some more comments to look at.

Makefile
bnshr: I cannot remember now why this was added.

edcdavid (commented):
I am surprised to see so much changes, your previous implementation needed only a simple change, e.g. to go through all operator namespaces instead of all Single operator namespaces. I feel this implementation is more convoluted and seems to miss a test case since now we are testing also application namespaces.

tests/operator/suite.go
```diff
@@ -126,6 +127,72 @@ func LoadChecks() {
...
+			check.LogInfo(msg)
+			compliantObjects = append(compliantObjects, testhelper.NewNamespacedReportObject(msg, testhelper.Namespace, true, operatorNamespace))
+		} else {
+			const prefix = "Namespace with operators"
+			var msg string
+			if installedSingleNamespacedOperators != "" {
+				msg = fmt.Sprintf("%s (%s)", prefix, installedSingleNamespacedOperators)
+			} else {
+				msg = prefix
+			}
```
greyerof, line 177: Suggestion:
```
msg := "Namespace with operators"
if installedSingleNamespacedOperators != "" {
	msg += " " + installedSingleNamespacedOperators
}
```

tests/operator/helper.go
```diff
@@ -45,3 +52,87 @@ func SplitCsv(csv string) CsvResult {
...
+				continue
+			}
+
+			// Handle cluster-wide operators
+			if operator.IsClusterWide {
+				allOperatorsFoundInNamespaceAreValid = false
+				break
+			}
+
+			if hasOperatorInstallModeType(operator.Csv.Spec.InstallModes, v1alpha1.InstallModeTypeAllNamespaces) {
```
greyerof, line 122: The function `hasOperatorInstallModeType()` is only checking the supported installation types, not the actual installation method of this operator. An operator that supports all the four installMode types but was installed as singleNamespace would be flagged here as AllNamespacedCsv... that doesn't make sense to me.

The proper way to check which mode an operator was installed for is by checking the field `.spec.targetNamespaces` field of the operatorgroup in the installation namespace. This is actually how the `operator.IsClusterWide` field [was computed during the autodiscovery phase](https://github.com/redhat-best-practices-for-k8s/certsuite/blob/d9a18113b82f944f2ce3c6cbb091bdfcdecdb33b/pkg/provider/operators.go#L194).

Besides that, the AllNamespaces mode (cluster-wide) check was already done at line 117, so why do you need to check that again here?

tests/operator/helper.go
```diff
@@ -45,3 +52,87 @@ func SplitCsv(csv string) CsvResult {
...
+				allOperatorsFoundInNamespaceAreValid = false
+				allNamespacedCsvs = append(allNamespacedCsvs, foundCsv)
+				break
+			}
+			if hasOperatorInstallModeType(operator.Csv.Spec.InstallModes, v1alpha1.InstallModeTypeSingleNamespace) {
```
greyerof, line 127: Same here. There's no need to check the supported modes. It was installed in own/singleNamespace mode if its operatorgroup's `len(spec.targetNamespaces) == 1`.

tests/operator/suite.go
```diff
@@ -126,6 +127,72 @@ func LoadChecks() {
...
+
+			var allNamespacedOperators string
+			var nonCompliantMsg string
+			if len(allNamespacedCsvs) != 0 {
+				allNamespacedOperators = strings.Join(allNamespacedCsvs, ", ")
+				nonCompliantMsg = fmt.Sprintf("%s contains all-namespaced operators (%s) ", msg, allNamespacedOperators)
+			} else {
+				nonCompliantMsg = fmt.Sprintf("%s contains cluster-wide operators", msg)
```
greyerof, line 185: Is this message string ok? The all-namespaced (cluster-wide) operators were already checked at line 181.

tests/operator/helper.go
```diff
@@ -45,3 +52,87 @@ func SplitCsv(csv string) CsvResult {
...
+	allOperatorsFoundInNamespaceAreValid := true
+
+	for _, operator := range env.Operators {
+		for foundCsv := range foundCsvs {
+			if operator.Csv.Name != foundCsv {
+				continue
+			}
+
+			// Handle cluster-wide operators
+			if operator.IsClusterWide {
```
sebrandon1, line 117: I feel like this `IsClusterWide` check could be moved into the suite.go func line 147 to add a non-compliant object immediately for any ClusterWide operators right off the bat.  Then after that you know you're only dealing with non-ClusterWide operators for the rest of the test.

tests/operator/helper.go
bnshr: Exactly, thanks for sharing this. I will modify that.

tests/operator/suite.go
```diff
@@ -126,6 +127,69 @@ func LoadChecks() {
...
+			} else {
+				msg = prefix
+			}
+
+			var allNamespacedOperators string
+			var nonCompliantMsg string
+			if len(allNamespacedCsvs) != 0 { // cluster-wide operator
+				allNamespacedOperators = strings.Join(allNamespacedCsvs, ", ")
+				nonCompliantMsg = fmt.Sprintf("%s contains all-namespaced operators (%s) ", msg, allNamespacedOperators)
+			} else if len(podsBelongingToNoOperators) != 0 {
```
edcdavid: instead of 2 non-compliant cases, we could collapse the same reason in one:
`Namespace with installed operators (ptp, prometheus)  contains all-namespaced operators ( operator1, operator 2) and contains some application pods (pod1, pod2, )`

tests/operator/helper.go
```diff
@@ -45,3 +53,97 @@ func SplitCsv(csv string) CsvResult {
...
+
+	if len(podsBelongingToNoOperators) == 0 && allOperatorsFoundInNamespaceAreValid {
+		return true, singleNamespacedCsvs, allNamespacedCsvs, podsBelongingToNoOperators, nil
+	} else {
+		return false, singleNamespacedCsvs, allNamespacedCsvs, podsBelongingToNoOperators, err
+	}
+}
+
+func generateNonCompliantMessage(singleNamespacedOperators string, allNamespacedCsvs, podsBelongingToNoOperators []string) (nonCompliantMsg string) {
+	prefix := "Namespace with installed operators"
```
edcdavid, line 129: If there are no single namespaces, omit them, like:	
	prefix := "Namespace "
	if singleNamespacedOperators != "" {
		prefix += "with installed single namespace operators (" + singleNamespacedOperators + ")"
	}

CATALOG.md
```diff
@@ -1346,6 +1346,22 @@ Tags|common,operator
 |Non-Telco|Mandatory|
 |Telco|Mandatory|
 
+#### operator-valid-single-namespaced-operator-installation-tenant-namespace
```
edcdavid, line 1349: why not rename this test case to operator-only-single-namespace-mode-allowed-in-tenant-namespaces

tests/operator/helper.go
```diff
@@ -45,3 +53,97 @@ func SplitCsv(csv string) CsvResult {
...
+			if len(operator.TargetNamespaces) == 1 {
+				singleNamespacedCsvs = append(singleNamespacedCsvs, foundCsv)
+			}
+		}
+	}
+
+	if len(podsBelongingToNoOperators) == 0 && allOperatorsFoundInNamespaceAreValid {
+		return true, singleNamespacedCsvs, allNamespacedCsvs, podsBelongingToNoOperators, nil
+	} else {
+		return false, singleNamespacedCsvs, allNamespacedCsvs, podsBelongingToNoOperators, err
```
edcdavid, line 124: return nil here

tests/operator/suite.go
```diff
@@ -126,6 +127,53 @@ func LoadChecks() {
...
+		isValidSingleNamespacedOperatorInstallation, singleNamespacedCsvs, allNamespacedCsvs, podsBelongingToNoOperators, err := containsValidSingleNamespacedOperatorIn(operatorNamespace)
+
+		if err != nil {
+			check.LogError("Skipped - cannot proceed with operator namespace %s", operatorNamespace)
+			continue
+		}
+
+		var installedSingleNamespacedOperators string
+		if len(singleNamespacedCsvs) != 0 {
+			installedSingleNamespacedOperators = fmt.Sprintf("(%s)", strings.Join(singleNamespacedCsvs, ", "))
```
edcdavid, line 162: remove extra parenthesis:
installedSingleNamespacedOperators = strings.Join(singleNamespacedCsvs, ", ")

tests/operator/helper.go
```diff
@@ -45,3 +53,97 @@ func SplitCsv(csv string) CsvResult {
...
+		pod := allPods[index]
+		topOwners, err := podhelper.GetPodTopOwner(pod.Namespace, pod.OwnerReferences)
+		if err != nil {
+			return foundCsvs, podsBelongingToNoOperators, err
+		}
+
+		validOwnerFound := false
+		for _, owner := range topOwners {
+			if owner.Kind == v1alpha1.ClusterServiceVersionKind && owner.Namespace == namespace {
+				foundCsvs[owner.Name] = true
```
edcdavid, line 79: We may need more checks here. other valid top owner could be the operands created based on the crds installed by this operator. Actually these operands can only be installed in the single operator namespace. For instance, for prometheus, pods belonging to the prometheus object top owner should be valid. We could list all the crds installs by this operator by reading the crd label: "operatorframework.io/installed-alongside-..." or similar then use the crd name to identify valid top owners.

greyerof: @edcdavid that's only valid for operators installed with the [ownNamespace](https://olm.operatorframework.io/docs/advanced-tasks/operator-scoping-with-operatorgroups/#ownnamespace) mode. For [singleNamespace](https://olm.operatorframework.io/docs/advanced-tasks/operator-scoping-with-operatorgroups/#singlenamespace) operators, the controller is running in namespace **A** and watches for CRs in namespace **B**, so, in theory, operand pods should only be found in namespace **B**.

The problem is the requirement only referes to SingleNamespaced and AllNamespaced operators... so what about ownNamespaced and multiNamespaced ones?

CATALOG.md
bnshr: Thanks for the suggestion.

tests/operator/suite.go
```diff
@@ -126,6 +127,69 @@ func LoadChecks() {
...
+
+	var compliantObjects []*testhelper.ReportObject
+	var nonCompliantObjects []*testhelper.ReportObject
+
+	operatorNamespaces := make(map[string]bool)
+	for _, operator := range env.Operators {
+		operatorNamespaces[operator.Namespace] = true
+	}
+
+	for operatorNamespace := range operatorNamespaces {
```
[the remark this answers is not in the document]
bnshr: I mark this as resolved by adding a new helper function to generate all non-compliant object message which would help to understand what conditions fail this test. Unit test is added for the same as well.

tests/operator/helper.go
edcdavid: ah, you are right about single namespace vs own namespace, in single namespace only mode, the operand should be in only in namespace B, sorry for the confusion. So in this case, we still need a way to ignore the csv that will be created in namespace B otherwise it will be picked up as an operator namespace.

edcdavid: > The problem is the requirement only referes to SingleNamespaced and AllNamespaced operators... so what about ownNamespaced and multiNamespaced ones?
In my opinion, the multi namespace is similar to the single namespace and it would make sense to support it here.

tests/operator/helper.go
```diff
@@ -45,3 +53,97 @@ func SplitCsv(csv string) CsvResult {
...
+
+	if len(podsBelongingToNoOperators) == 0 && allOperatorsFoundInNamespaceAreValid {
+		return true, singleNamespacedCsvs, allNamespacedCsvs, podsBelongingToNoOperators, nil
+	} else {
+		return false, singleNamespacedCsvs, allNamespacedCsvs, podsBelongingToNoOperators, nil
+	}
+}
+
+func generateNonCompliantMessage(singleNamespacedOperators string, allNamespacedCsvs, podsBelongingToNoOperators []string) (nonCompliantMsg string) {
+	prefix := "Namespace"
```
edcdavid, line 129: "Namespace" -> "Operator Namespace"

greyerof:
Leaving this comment just to summarize what we've been talking in the internal meeting: 
```
ns1
  operatorgroup's.targetNamespaces = ns2
  subsOp1
  subsOp2
  csvOp1  -> valid
  csvOp2  -> valid

  op1-controller-pod  -> valid
  op2-controller-pod  -> valid
  
  csvOp3        -> invalid
  workload-pod  -> invalid
ns2
  csvOp1
  csvOp2
  
  Op1CR
  Op2CR
  op1-operand-pod1
  op2-operand-pods1
  
ns3
  workload-pod1
  workload-pod2
  
ns4
 operatogroup's.targetNamespaces = ns1
 subsOp3
```
In **ns1**, due to operatogroup's.targetNamespaces = ns2, only singleNamespaced operators are allowed, therefore there should be only controller pods of Op1 and Op2. Any workload pod, operand pod or csv from singleNamespace operators installed in a different ns (e.g. csvOp3) should make the tc to fail.

bnshr:
> In **ns1**, due to operatogroup's.targetNamespaces = ns2, only singleNamespaced operators are allowed, therefore there should be only controller pods of Op1 and Op2. Any workload pod, operand pod or csv from singleNamespace operators installed in a different ns (e.g. csvOp3) should make the tc to fail.

What messages do we expect for non-compliant namespaces / operators?

greyerof:
> > In **ns1**, due to operatogroup's.targetNamespaces = ns2, only singleNamespaced operators are allowed, therefore there should be only controller pods of Op1 and Op2. Any workload pod, operand pod or csv from singleNamespace operators installed in a different ns (e.g. csvOp3) should make the tc to fail.
> 
> What messages do we expect for non-compliant namespaces / operators?

Not a strong opinion on this, but I think we should use the ReportObject's "AddField" method to include every resource found that makes that namespace not compatible with the operator under test. E.g.:
```
nonCompliantNs := NewNamespacedReportObject("Namespace ns1 is not a valid dedicated namespace for singleNamespace operators", testhelper.Namespace, false, operatorNamespace)
// For any workload pod (not belonging to any operator)
for _, podName := range podsBelongingToNoOperators {
	nonCompliantNs.AddField("Invalid pods found",podName)
}

// For CSV founds in ns1 that belong to ownNamespace/multiNamespace operators not installed in ns1.
for _, csvName := range csvNotBelongingToSingleNamespaceOperators {
	nonCompliantNs.AddField("Invalid CSV found",csvName)
}

// For any operand pod not belonging to the singleNamespace operators installed in ns1
for _, podName := range operandPodNotBelongingToSingleNamespaceOperators {
	nonCompliantNs.AddField("Invalid operand pod found",csvName)
}
```

tests/operator/helper.go
```diff
@@ -45,3 +52,95 @@ func SplitCsv(csv string) CsvResult {
...
+	}
+	return operatorsInNamespace
+}
+
+func isSingleNamespacedOperator(operator *provider.Operator) bool {
+	return len(operator.TargetNamespaces) == 1 && operator.Namespace != operator.TargetNamespaces[0]
+}
+
+func isMultiNamespacedOperator(operator *provider.Operator) bool {
+	return len(operator.TargetNamespaces) > 1 && !stringhelper.StringInSlice(operator.TargetNamespaces, operator.Namespace, false)
```
sebrandon1, line 80: I'm not sure the check for `len(operator.TargetNamespaces) > 1` is actually needed because `StringInSlice` will already return false for a slice that's 0 length.

tests/operator/helper.go
```diff
@@ -45,3 +52,95 @@ func SplitCsv(csv string) CsvResult {
 	}
 	return result
 }
+
+func getAllPodsBy(namespace string) (podsInNamespace []*provider.Pod) {
+	for i := range env.AllPods {
```
sebrandon1, line 57: Just in general I feel like these filtering functions should accept `env.AllPods` as a parameter instead of just referencing the global variable.  Not a huge deal but for unit testing functions it makes it easier.

sebrandon1 (commented):
Some additional comments.

tests/operator/helper.go
bnshr: Makes sense.

tests/operator/helper.go
bnshr: We need that, imo. We want to avoid that situation for ClusterWide operators. Does it make sense?

tests/operator/helper.go
```diff
@@ -45,3 +52,90 @@ func SplitCsv(csv string) CsvResult {
...
+
+func checkValidOperatorInstallation(namespace string) (isDedicatedOperatorNamespace bool, validInstalledOperators, invalidInstalledOperators, invalidNonInstalledOperators, unlabeledInstalledOperators, podsNotBelongingToOperators []string) {
+	// 1. operator installation checks
+	for _, operator := range getAllOperatorsBy(namespace, env.AllOperators) {
+		if namespace == operator.Csv.Annotations["olm.operatorNamespace"] {
+			if checkIfOperatorUnderTest(operator) {
+				if isSingleNamespacedOperator(operator) || isMultiNamespacedOperator(operator) {
+					validInstalledOperators = append(validInstalledOperators, operator.Name)
+				} else {
+					invalidInstalledOperators = append(invalidInstalledOperators, operator.Name)
```
edcdavid, line 101: Replace across  PR:
invalidInstalledOperators -> singleOrMultiNamespaceOperators
validInstalledOperators -> nonSingleOrMultiNamespaceOperators

tests/operator/helper.go
```diff
@@ -51,6 +54,97 @@ func SplitCsv(csv string) CsvResult {
...
+				if isSingleNamespacedOperator(operator) || isMultiNamespacedOperator(operator) {
+					singleOrMultiNamespaceOperators = append(singleOrMultiNamespaceOperators, operator.Name)
+				} else {
+					nonSingleOrMultiNamespaceOperators = append(nonSingleOrMultiNamespaceOperators, operator.Name)
+				}
+			} else {
+				operatorsFoundButNotUnderTest = append(operatorsFoundButNotUnderTest, operator.Name)
+			}
+		} else {
+			csvsFoundButNotInOperatorInstallationNamespace = append(csvsFoundButNotInOperatorInstallationNamespace, operator.Name)
```
edcdavid, line 110: csvs from All namespace operators should be fine since they are ok from requirements, unless they are actually installed in the tenant namespace. I would add:
} else if len(operator.TargetNamespaces) !=0  {

tests/operator/helper.go
```diff
@@ -51,6 +54,97 @@ func SplitCsv(csv string) CsvResult {
...
+	}
+
+	// 2. existing pods check
+	podsNotBelongingToOperators, err := findPodsNotBelongingToOperators(namespace)
+	if err != nil {
+		return false, singleOrMultiNamespaceOperators, nonSingleOrMultiNamespaceOperators, csvsFoundButNotInOperatorInstallationNamespace, operatorsFoundButNotUnderTest, podsNotBelongingToOperators
+	}
+
+	return len(podsNotBelongingToOperators) == 0 && len(singleOrMultiNamespaceOperators) != 0, singleOrMultiNamespaceOperators, nonSingleOrMultiNamespaceOperators,
+		csvsFoundButNotInOperatorInstallationNamespace, operatorsFoundButNotUnderTest, podsNotBelongingToOperators
```
edcdavid, line 121: Fix condition to include all namespace operators in passing case when they are not installed in tenant namespaces

tests/operator/suite.go
```diff
@@ -128,6 +129,75 @@ func LoadChecks() {
...
+
+			if len(nonSingleOrMultiNamespaceOperators) != 0 {
+				msg += " - invalid operators " + strings.Join(nonSingleOrMultiNamespaceOperators, ", ")
+			}
+
+			if len(nonSingleOrMultiNamespaceOperators) != 0 {
+				msg += "- invalid installed operators (not single/multiple namespaced) " + strings.Join(nonSingleOrMultiNamespaceOperators, ", ")
+			}
+			if len(csvsFoundButNotInOperatorInstallationNamespace) != 0 {
+				msg += "- invalid installed operators (installed in other namespace) " + strings.Join(csvsFoundButNotInOperatorInstallationNamespace, ", ")
```
edcdavid, line 186: msg += "- invalid installed operators (installed in other namespace) " + strings.Join(csvsFoundButNotInOperatorInstallationNamespace, ", ")
->
msg += "- this namespace is the target namespace of other operators (" + strings.Join(csvsFoundButNotInOperatorInstallationNamespace, ", ")+")\n"

tests/operator/suite.go
```diff
@@ -128,6 +129,75 @@ func LoadChecks() {
...
+			}
+			if len(operatorsFoundButNotUnderTest) != 0 {
+				msg += "- operators not under test found " + strings.Join(operatorsFoundButNotUnderTest, ", ")
```
edcdavid, line 189: msg += "- operators not under test found " + strings.Join(operatorsFoundButNotUnderTest, ", ")
->
msg += "- operators not under test found (" + strings.Join(operatorsFoundButNotUnderTest, ", ")+")\n"

tests/operator/suite.go
```diff
@@ -128,6 +129,75 @@ func LoadChecks() {
...
+			}
+			if len(podsNotBelongingToOperators) != 0 {
+				msg += "- invalid non operator pods found = " + strings.Join(podsNotBelongingToOperators, ", ")
```
edcdavid, line 192: msg += "- invalid non operator pods found = " + strings.Join(podsNotBelongingToOperators, ", ")
->
msg += "- invalid non operator pods found (" + strings.Join(podsNotBelongingToOperators, ", ")+")"

tests/operator/helper.go
```diff
@@ -83,3 +88,112 @@ func OperatorInstalledMoreThanOnce(operator1, operator2 *provider.Operator) bool
...
+
+		var targetNameSpaces []string
+		if len(targetNamespacesStr) != 0 {
+			targetNameSpaces = strings.Split(targetNamespacesStr, ",")
+		}
+
+		if namespace == operatorNamespace {
+			if checkIfCsvUnderTest(csv) {
+
+				isValidInstallation := isSingleNamespacedOperator(operatorNamespace, targetNameSpaces) || isMultiNamespacedOperator(operatorNamespace, targetNameSpaces)
```
edcdavid, line 146: isValidInstallation -> isSingleOrMultiInstallation

tests/operator/helper.go
```diff
@@ -83,3 +88,112 @@ func OperatorInstalledMoreThanOnce(operator1, operator2 *provider.Operator) bool
...
+					singleOrMultiNamespaceOperators = append(singleOrMultiNamespaceOperators, csv.Name)
+				} else {
+					nonSingleOrMultiNamespaceOperators = append(nonSingleOrMultiNamespaceOperators, csv.Name)
+				}
+			} else {
+				operatorsFoundButNotUnderTest = append(operatorsFoundButNotUnderTest, csv.Name)
+			}
+		} else {
+			if len(targetNameSpaces) != 0 { // check for non-cluster wide operators
+				csvsFoundButNotInOperatorInstallationNamespace = append(csvsFoundButNotInOperatorInstallationNamespace, csv.Name)
```
edcdavid, line 157: csvsFoundButNotInOperatorInstallationNamespace -> csvsTargetingNamespace

tests/operator/helper.go
```diff
@@ -83,3 +88,112 @@ func OperatorInstalledMoreThanOnce(operator1, operator2 *provider.Operator) bool
...
+			}
+		}
+	}
+
+	// 2. existing pods check
+	podsNotBelongingToOperators, err := findPodsNotBelongingToOperators(namespace)
+	if err != nil {
+		return false, singleOrMultiNamespaceOperators, nonSingleOrMultiNamespaceOperators, csvsFoundButNotInOperatorInstallationNamespace, operatorsFoundButNotUnderTest, podsNotBelongingToOperators
```
edcdavid, line 165: I would add another return (err) to report an internal error reason and display the err

edcdavid (commented):
Almost there! few comments on the results presentation

PR #2614, opened 2024-12-06
Add WaitingForRelease feature flag

sebrandon1:
As of right now, when we are adding new tests to the repo we can't really release a new version of the certsuite until everything has been merged in.  This will allow us to set tests as disabled and they won't affect existing release version of the certsuite.

Changes:
- Adds a `WaitingForRelease` tag similar to `TagCommon`, `TagTelco`, etc.
- v1.6 tests that have been newly added have had `WaitingFoRelease` added to them.  This removes their entries from the CATALOG.md temporarily until we are ready to fully release them to partners.

To keep the tests enabled in our CI's we are setting an environment variable `REDHAT_CI` which will always set the test cases to enabled regardless of `waiting-for-release" tag.

### DIFF
--- a/pkg/checksdb/check.go
+++ b/pkg/checksdb/check.go
@@ -56,6 +56,8 @@ type Check struct {
 	Timeout            time.Duration
 	Error              error
 	abortChan          chan string
+
+	enabled bool
 }
 
 func NewCheck(id string, labels []string) *Check {
@@ -64,10 +66,17 @@ func NewCheck(id string, labels []string) *Check {
 		Labels:     labels,
 		Result:     CheckResultPassed,
 		logArchive: &strings.Builder{},
+		enabled:    true,
 	}
 
 	check.logger = log.GetMultiLogger(check.logArchive, cli.CliCheckLogSniffer).With("check", check.ID)
 
+	return check
+}
+
+// SetDisabled disables the check.
+func (check *Check) SetDisabled() *Check {
+	check.enabled = false
 	return check
 }
 
--- a/pkg/checksdb/check.go
+++ b/pkg/checksdb/check.go
@@ -56,8 +56,6 @@ type Check struct {
 	Timeout            time.Duration
 	Error              error
 	abortChan          chan string
-
-	enabled bool
 }
 
 func NewCheck(id string, labels []string) *Check {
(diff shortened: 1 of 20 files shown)


pkg/checksdb/check.go
```diff
@@ -64,13 +66,27 @@ func NewCheck(id string, labels []string) *Check {
...
 	check.logger = log.GetMultiLogger(check.logArchive, cli.CliCheckLogSniffer).With("check", check.ID)
 
 	return check
 }
 
+// SetDisabled disables the check.
+func (check *Check) SetDisabled() *Check {
+	// Check if the environment variable REDHAT_CI is set.
+	// If it is set, the check is enabled.
+	if os.Getenv("REDHAT_CI") != "" {
```
rdavid, line 81:
```suggestion
	if os.Getenv("REDHAT_CI") == "true" {
```

sebrandon1:
This is probably going to fail DCI until they add the environment variable.  Right now I have it set as `REDHAT_CI` but I could be persuaded to use another variable or way to differentiate how to add the checks into the group that are waiting for release.

edcdavid (approved):

greyerof:
Just my two cents here: if a test case made it (after the review+approval process) to the main branch, it is then supposed to be fully functional and, at least, it should have been manually tested, so appearing in the catalog as a new official test case in the next release cut seems 100% legit to me in that case. Why should we hide them and wait for some X.Y.Z release to announce all of them at once?

In case we want/need to go with half-baked test cases, why not just flag them with some extra label like that "waiting-for-release" or similar ("release-pending", "not-released", "in-progress"...)? It should be sufficient to state in our official documentation that the results of those test cases, whether they pass or fail, should not be considered official, as they are not officially released/supported yet. Just removing that label when they're 100% ready or before the release cut should work. Adding extra env vars shouldn't be necessary.

sebrandon1:
That's fair.  There's nothing "wrong" with just releasing what we currently have merged into `main` other than just keeping it more organized for users.  I agree the environment variable makes it ugly.  I'll mark this as closed and we can discuss it more in the future if needed.

In the meantime, I think I will just release `5.4.x` with the currently merged test cases.  That's how we've been doing it in the past if there are any new test cases by just bumping the minor version. (major.minor.patch).

PR #2617, opened 2024-12-10
Fix for access-control-security-context-non-root-user-id-check

greyerof:
This test case had the same issue that was fixed here: https://github.com/redhat-best-practices-for-k8s/certsuite/pull/2557
With the prior implementation, the test case failed if pod's level SCC's runAsUser was set to 0 but container's SCC runAsUser was set to any non-zero value, like 1001. This should pass as container's SCC runAsUser value takes precedence over pod's one.

I followed the same approach: the field RunAsUser in container's SCC takes precedence over pod's level one.

Also, replaced the helper function stringhelper.BoolToString with stringhelper.PointerToString(), which works for any type now, as long as it has a stringer method implementation.

### DIFF
--- a/pkg/provider/containers.go
+++ b/pkg/provider/containers.go
@@ -197,10 +197,26 @@ func (c *Container) IsReadOnlyRootFilesystem(logger *log.Logger) bool {
 
 func (c *Container) IsContainerRunAsNonRoot(podRunAsNonRoot *bool) (isContainerRunAsNonRoot bool, reason string) {
 	if c.SecurityContext != nil && c.SecurityContext.RunAsNonRoot != nil {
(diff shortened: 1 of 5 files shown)


sebrandon1 (approved):

greyerof:
certsuite-qe PR https://github.com/redhat-best-practices-for-k8s/certsuite-qe/pull/1014 needs to be merged first so the qe workflow in ubuntu can pass.

sebrandon1:
https://github.com/redhat-best-practices-for-k8s/certsuite-qe/pull/1014 has been merged and I rekicked the QE run in the CI.

edcdavid (approved):

PR #2629, opened 2024-12-11
Add upload of tar.gz to Red Hat Connect endpoint

sebrandon1:
Adds the ability to automatically upload the certsuite `tar.gz` file directly to the Red Hat Connect API portal.  

Marked WIP until the Connect Portal is out of development.

This adds 5 flags to the certsuite config file for the:
- API Key
- Project ID
- Connect API Base URL
- API Proxy URL (if needed, can probably be removed after moved out of development)
- API Proxy Port (if needed, can probably be removed after moved out of development)

You currently need to be on the Red Hat internal VPN for this to work in its current state with the proxy enabled.


config/certsuite_config.yml
```diff
@@ -38,3 +38,8 @@ executedBy: ""
 partnerName: ""
 collectorAppPassword: ""
 collectorAppEndpoint: "http://claims-collector.cnf-certifications.sysdeseng.com"
+connectAPIKey: ""
```
greyerof, line 41: Suggestion: make a placeholder field for all these params, with a bit more descriptive name like"redhat-connect-api":
```
redhat-connect-api:
  projectID: ...
  key: ...
  baseURL: ...
  ...
```
or
```
redhat-connect:
  projectID: ...
  api: 
    key: ...
    baseURL: ...
    ...
```

internal/results/archiver.go
```diff
@@ -38,14 +44,14 @@ func getFileTarHeader(file string) (*tar.Header, error) {
 }
 
 // Creates a zip file in the outputDir containing each file in the filePaths slice.
-func CompressResultsArtifacts(outputDir string, filePaths []string) error {
+func CompressResultsArtifacts(outputDir string, filePaths []string) (string, error) {
```
greyerof, line 47: Please update func's comment to include this new string output param.

internal/results/archiver.go
```diff
@@ -60,25 +66,248 @@ func CompressResultsArtifacts(outputDir string, filePaths []string) error {
...
+	zipFilePath, err = filepath.Abs(zipFilePath)
+	if err != nil {
+		return "", fmt.Errorf("failed to get absolute path for %s: %v", zipFilePath, err)
+	}
+
+	// Return the entire path to the zip file
+	return zipFilePath, nil
+}
+
+func createFormField(w *multipart.Writer, field, value string) error {
```
greyerof, line 99: What about moving all these funcs to a new file like "internal/results/rhconnect.go"?

sebrandon1: Good point.  I started it out in archiver but that would make more sense.  I'll move them.

internal/results/archiver.go
sebrandon1: Yup I can do that.  👍

edcdavid (approved):

config/certsuite_config.yml
sebrandon1: Done!

PR #2648, opened 2024-12-17
Update preflight to v1.11.1

sebrandon1:
https://github.com/redhat-openshift-ecosystem/openshift-preflight/releases/tag/1.11.1


sebrandon1:
@ramperher This new update to preflight v1.11.1 adds a new test which is failing in the DCI run:

```
'actual_result': {'testcase': 'preflight-HasRequiredLabel', 'passed': False}}]"}
```

rdavid (approved):

itroyano:
The good news is the new test ran :)

ramperher:
> @ramperher This new update to preflight v1.11.1 adds a new test which is failing in the DCI run:
> 
> ```
> 'actual_result': {'testcase': 'preflight-HasRequiredLabel', 'passed': False}}]"}
> ```

I've added an exception for this test in our pipeline, but note we're testing your sample workload in all these jobs, so if this test is failing, this means your workload needs to be updated to pass this test, and also the other one that is failing, which is `preflight-HasUniqueTag`.

itroyano:
cc: @acornett21 @komish
Looks like a test image that doesn't set a "maintainer" level,  sets the name at "Red Hat",  or both.

PR #2673, opened 2025-01-08
feat : Restrict the size of operator catalog < 200GB

bnshr:
Refers https://issues.redhat.com/browse/CNFCERT-1087


tests/operator/catalogsource/catalogsource.go
```diff
@@ -28,3 +37,115 @@ func SkipPMBasedOnChannel(channels []olmpkgv1.PackageChannel, csvName string) bo
 
 	return skipPMBasedOnChannel
 }
+
+// TODO: Get all catalog sources of application operators
+func getOperatorCatalogSources() {
```
sebrandon1, line 42: We get `AllCatalogSources` in the autodiscover here: https://github.com/redhat-best-practices-for-k8s/certsuite/blob/main/pkg/provider/provider.go#L121

So they are already in the `env`.  You could use that in the lookup you use below as well in `func getCatalogSize(catalogNamespace, catalogSourceName string)` where you could just loop through the `env` and look for the matching name/namespace combo.

PR #2700, opened 2025-01-17
Update the RHCOS script

sebrandon1:
Improving on the existing RHCOS script:
- Added logic to pull in existing mapped version `x.y.z / rhcos` key value mappings from the file prior to running the rest of the script.  This makes the runtime of the script much faster as it doesn't have to re-query the RHCOS versions (which never change) from the API.
- Removed channels 4.4, 4.3, 4.2, and 4.1 from the list of channels to query because they do not exist any more in `oc adm` API calls.
- Added some counters to keep track of successful and failed API attempts.
- Added logic to skip lookups for already deemed missing versions that do not return valid RHCOS versions from the API.
- If the `oc adm` API query fails with a non-zero exit code, retry up to 5 times with a 1 second sleep in between each.


greyerof (approved):

script/rhcos_versions.sh
```diff
@@ -1,34 +1,95 @@
 #!/usr/bin/env bash
 
-set -x
-set -e
```
rdavid, line 4: `set -o errexit -o pipefail` means exiting on any error (`errexit`) and treating errors in pipes as failures (`pipefail`). Solid settings.

rdavid (approved):

script/rhcos_versions.sh
sebrandon1: Done.

PR #2741, opened 2025-01-29
connect api: override rhconnect vars from cmdline

sebrandon1:
Override env vars with cmdline flags if available.

Follow up to:
#2629 

Test-Hints: no-check


ramperher:
recheck

ramperher:
@sebrandon1 , this is now working good, I've validated it in my last run

ramperher (approved):
LGTM

PR #2795, opened 2025-02-13
Update Go to v1.24.0

sebrandon1:
https://go.dev/doc/go1.24

Changes:
- Using `go-version-file` option thanks to @acornett21 
- Go 1.24.x now complains if you don't properly format your strings so this is now fixed.

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -13,10 +13,10 @@ runs:
         echo '{ "auths": {} }' >> ${PFLT_DOCKERCONFIG}
       shell: bash
 
-    - name: Set up Go 1.23
+    - name: Set up Go
       uses: actions/setup-go@v4
       with:
-        go-version: 1.23.6
+        go-version-file: go.mod
         cache: false
 
     - name: Disable default go problem matcher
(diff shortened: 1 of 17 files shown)


.github/actions/setup/action.yml
```diff
@@ -13,10 +13,10 @@ runs:
         echo '{ "auths": {} }' >> ${PFLT_DOCKERCONFIG}
       shell: bash
 
-    - name: Set up Go 1.23
+    - name: Set up Go 1.24
       uses: actions/setup-go@v4
       with:
-        go-version: 1.23.6
+        go-version: 1.24.0
```
acornett21, line 19: Instead of all of these changes for ever go version, would it make sense to switch to 
```suggestion
        go-version-file: go.mod
```

acornett21 (approved):

sebrandon1:
Other repos that were dealing with the Go fmt fix for example/reference:
- https://github.com/opentofu/opentofu/pull/2390
- https://github.com/codefresh-io/argo-rollouts/pull/313
- https://github.com/openshift/kubernetes/pull/2094
- https://github.com/kubernetes-sigs/structured-merge-diff/issues/281
- https://github.com/kubernetes/kubernetes/pull/129519

greyerof (approved):

PR #2796, opened 2025-02-13
Making Operator tenant test case optional

edcdavid:
The Following test case `single-or-multi-namespaced-allowed-in-tenant-namespaces` is only mandatory in the extended scenario. Updated the catalog to reflect this.
Catalog updated also picked up the latest testcase


sebrandon1 (approved):

greyerof (approved):

ramperher (approved):
Thanks for addressing this!

PR #2812, opened 2025-02-17
fix :  Logic for detecting the targeting namespace operators from valid cluster-wide operators

bnshr:


sebrandon1 (approved):
This could use a simple unit test.

PR #2868, opened 2025-03-19
Update documentation about running intrusive tests

rocrisp:
because the newest version of the tooling is managed by the flag and not by env

build-depends: https://github.com/redhatci/ansible-collection-redhatci-ocp/pull/621
build-depends: https://github.com/dci-labs/dallas-pipelines/pull/1398

Test-Hints: no-check


cmd/certsuite/run/run.go
```diff
@@ -34,7 +34,7 @@ func NewCommand() *cobra.Command {
 	runCmd.PersistentFlags().String("log-level", "debug", "Sets the log level")
 	runCmd.PersistentFlags().String("offline-db", "", "Set the location of an offline DB to check the certification status of for container images, operators and helm charts")
 	runCmd.PersistentFlags().String("preflight-dockerconfig", "", "Set the dockerconfig file to be used by the Preflight test suite")
-	runCmd.PersistentFlags().Bool("non-intrusive", false, "Run only the test that do not disrupt the test environment")
+	runCmd.PersistentFlags().Bool("non-intrusive", false, "Run intrusive tests that may disrupt the test environment")
```
acornett21, line 37: Why the comment change? Was this incorrect in the first place? Or is this a flag where it's truthiness does the opposite of what the flag states?

acornett21 (commented):
Just a couple of questions.

docs/cluster-deploy/kustomization.yaml
```diff
@@ -34,5 +34,6 @@ resources:
 #       name: certsuite
 #     patch: |
 #       - op: replace
-#         path: /spec/containers/0/env/0/value
-#         value: false
+#         path: /spec/containers/0/env/1
```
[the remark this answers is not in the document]
rocrisp: oh dang, there's an error. The correct path should be:
path: /spec/containers/0/arg/1
We removed the env and replaced it with a flag.
I'll make a correction now.

cmd/certsuite/run/run.go
rocrisp: I believe it was incorrect in the first place.
According to [doc](https://redhat-best-practices-for-k8s.github.io/certsuite/runtime-env/#ocp-412-labels), to enable intrusive tests, export CERTSUITE_NON_INTRUSIVE_ONLY=false.

acornett21: Maybe I'm reading this incorrectly, but it seems the flags name, it's truthiness, or it's description are incorrect. 

@sebrandon1 Am I just reading this incorrectly?

acornett21: I know that this increases the scope of this ask, but if we can fix this confusion now and switch to `--intrusive: true` as the default and globally refactor to drop the `non` elsewhere, I'm all for that to avoid confusion in the future.

sebrandon1: Agreed, let's just adjust in one shot.

rocrisp: I will refactor shortly.

rocrisp: @sebrandon1 Do you know what is the DISABLE_INTRUSIVE_TESTS in the workflows set to?
I'm not so sure how is this env related to --intrusive.
- name: Run the tests (against binary)
        uses: nick-fields/retry@ce71cc2ab81d554ebbe88c79ab5975992d79ba08 # v3.0.2
        with:
          timeout_minutes: 60
          max_attempts: 3
          command: cd ${GITHUB_WORKSPACE}/certsuite-qe; FEATURES=${{matrix.suite}} CERTSUITE_REPO_PATH=${GITHUB_WORKSPACE} USE_BINARY=true JOB_ID=${{github.run_id}} DISABLE_INTRUSIVE_TESTS=false ENABLE_PARALLEL=true ENABLE_FLAKY_RETRY=true make test-features

sebrandon1: So let's just leave those, because those flags are for the [certsuite-qe](https://github.com/redhat-best-practices-for-k8s/certsuite-qe) repo and they just cause certain tests over there to be skipped completely.

The flag is here for reference: https://github.com/redhat-best-practices-for-k8s/certsuite-qe/blob/8864991067367859352ecddfc22498b00a8374b9/tests/utils/config/config.go#L39

ramperher:
I've added the support of this new feature in the k8s_best_practices_certsuite role: https://github.com/redhatci/ansible-collection-redhatci-ocp/pull/621. Please wait till having a successful DCI job to confirm this is working.

Also, please make sure of bumping certsuite version once this is merged, since this is a really breaking change in the definition of default variables. Thanks.

rocrisp:
@ramperher Is there anything I need todo here?

ramperher:
> @ramperher Is there anything I need todo here?

No, with what regards DCI, it's currently fine

sebrandon1 (approved):

acornett21 (approved):

PR #2875, opened 2025-03-25
golangci-lint: migration to v2

sebrandon1:
https://golangci-lint.run/product/migration-guide/

I migrated the config over to their `v2` style with the `migrate` command.  Also I fixed a number of things it started flagging.


Makefile
```diff
@@ -114,10 +113,6 @@ install-mac-brew-tools:
 		hadolint \
 		shfmt
 
-# Installs linters
```
acornett21, line 117: Wouldn't we want to keep this so people can test linting locally? but maybe put it in `./bin` local to the project and have that folder in `.gitignore`?

sebrandon1: I'll leave it in there.

PR #2903, opened 2025-04-08
fix bin directory to no longer have folders and duplicate binaries

acornett21:
## Motivation
When trying to figure out where the `certsuite` binary lived in the container, I noticed the `tree` output, had the binary duplicated in two places within the `/usr` location.

## Changes
Fix the duplication, and remove the folders in `/usr/local/bin`, creating a flat directory of all binaries.

Test-Hints: no-check

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -142,11 +142,10 @@ ENV \
 	OSDK_BIN=/usr/local/osdk/bin
 
 # Install the certsuite binary
-COPY --from=build ${CERTSUITE_DIR} ${CERTSUITE_DIR}
-RUN cp ${CERTSUITE_DIR}/certsuite /usr/local/bin
+COPY --from=build ${CERTSUITE_DIR}/certsuite /usr/local/bin/certsuite
 
 # Add operatorsdk binary to image
-COPY --from=build ${OSDK_BIN} ${OSDK_BIN}
+COPY --from=build ${OSDK_BIN} /usr/local/bin/operator-sdk
 
 # Update the CNF containers, helm charts and operators DB
 ENV \
@@ -154,7 +153,6 @@ ENV \
 	OCT_DB_PATH=/usr/oct/cmd/tnf/fetch
 COPY --from=db ${OCT_DB_PATH} ${CERTSUITE_OFFLINE_DB}
 
-ENV PATH="${OSDK_BIN}:${PATH}"
 WORKDIR ${CERTSUITE_DIR}
 ENV SHELL=/bin/bash
 CMD ["certsuite", "-h"]


Dockerfile
```diff
@@ -142,19 +142,17 @@ ENV \
 	OSDK_BIN=/usr/local/osdk/bin
 
 # Install the certsuite binary
-COPY --from=build ${CERTSUITE_DIR} ${CERTSUITE_DIR}
```
sebrandon1, line 145: So `CERTSUITE_DIR` is the actual DIR where the `CERTSUITE_DIR/certsuite` is the binary.  I think you just need: 

`COPY --from=build ${CERTSUITE_DIR}/certsuite /usr/local/bin/certsuite`

acornett21: ahh I was reading it as the binary name, but that makes sense based on the `tree` output.

sebrandon1: Although, now that I think about it, I think you might hit a problem where you also need the `rhcos_version_map` file (tests/platform/operatingsystem/files/rhcos_version_map) if you aren't copying the entire directory for that test to work.

ramperher:
@acornett21 , @sebrandon1  DCI job is failing because it's not able to find `certsuite` binary. Where's the correct place to look at it? We invoke it currently as `./certsuite run...`

acornett21:
@ramperher Are you saying that you change the entry point to explicitly call `./` when running the container in DCI? I'm confused why anything would need to be specified since the binary was in `$PATH` previously.

Dockerfile
acornett21: @sebrandon1 The code shows that `go:embed` is being used to include the file at binary build time, is that not working correctly?

ramperher:
> @ramperher Are you saying that you change the entry point to explicitly call `./` when running the container in DCI? I'm confused why anything would need to be specified since the binary was in `$PATH` previously.

@acornett21 , we're just using what's documented in the [certsuite official docs](https://redhat-best-practices-for-k8s.github.io/certsuite/test-run/):

_The Test Suite can be run using the Certsuite tool directly or through a container._

_To run the Test Suite direct use:_

`./certsuite run -l <label-filter> -c <certsuite-config> -k <kubeconfig> -o <output-dir> [<flags>]`

So, we consequently use `./certsuite run`: https://github.com/redhatci/ansible-collection-redhatci-ocp/blob/main/roles/k8s_best_practices_certsuite/tasks/tests.yml#L53

As I said, that's not working in the test launched with the image generated from this PR, so if something is changing, I guess the docs should be changed too.

acornett21:
@ramperher It's odd to me that the docs would list the `./` in the first place, especially for the container, since the binary is in `$PATH` and the directory shouldn't matter to the calling user.

ramperher:
> @ramperher It's odd to me that the docs would list the `./` in the first place, especially for the container, since the binary is in `$PATH` and the directory shouldn't matter to the calling user.

That's fine for us, but as I said, we were mainly following the docs from this repo, so I believe the docs should be changed accordingly.

ramperher (approved):
LGTM

sebrandon1 (approved):

PR #2906, opened 2025-04-10
fsdiff: add retries around runPodmanDiff command

sebrandon1:
We have seen where a container is initially `Running` during autodiscover, it might fall into a `CrashLoopBackOff` state before the test is executed.  Let's add some retries here to see if we can smooth out any potential errors.

### DIFF
--- a/tests/platform/cnffsdiff/fsdiff.go
+++ b/tests/platform/cnffsdiff/fsdiff.go
@@ -20,6 +20,8 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"strings"
+	"time"
 
 	"github.com/Masterminds/semver/v3"
 	"github.com/redhat-best-practices-for-k8s/certsuite/internal/clientsholder"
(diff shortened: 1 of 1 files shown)


tests/platform/cnffsdiff/fsdiff.go
```diff
@@ -165,7 +168,25 @@ func (f *FsDiff) RunTest(containerUID string) {
...
 	f.check.LogInfo("Running \"podman diff\" for container id %s", containerUID)
-	output, err := f.runPodmanDiff(containerUID)
+
+	var output string
+	var err error
+	for i := range [5]int{} {
+		output, err = f.runPodmanDiff(containerUID)
+		if err == nil {
+			break
+		}
```
rdavid, line 178: You can combine break with both no error and errors other than 125. There's no need for continue.

sebrandon1: I'm not quite sure, because we need the continue to restart the loop if we're waiting for error 125 to go away.  

For the first break, that's because there was no error (happy path).  For the second break, that's if the error != nil and doesn't relate to error 125.  

Although now you're making me think about it 😄 I think this should be good.

PR #2925, opened 2025-04-23
update test run docs to remove ./ for clarity

acornett21:
## Motivation
A follow up to the below, where it was pointed out the docs should be updated, since it wasn't clear that for a container the binary was in $PATH.
- #2903

## Changes
Remove references to `./` in the `test-run` page. It still makes sense to keep this in the developer pages/guide.

Test-Hints: no-check


ramperher:
recheck

ramperher (approved):
Thanks for this update!

sebrandon1 (approved):

PR #2956, opened 2025-05-05
Dockerfile: Adjust multi-stage build to save space

sebrandon1:
Changes:
* Consolidated the build and cleanup steps in the `RUN` command to streamline the process and remove unnecessary intermediate files, reducing the image size. Removed a redundant cleanup `RUN` block. (`Dockerfile`, [DockerfileL109-L131](diffhunk://#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557L109-L131))
* Added a `docker image prune -f` command to the `build-image-local` target to automatically remove dangling images after the build, further minimizing disk usage. (`Makefile`, [MakefileL154-R155](diffhunk://#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L154-R155))


Makefile
```diff
@@ -151,7 +151,8 @@ build-image-local:
 		-t ${REGISTRY}/${CERTSUITE_IMAGE_NAME}:${IMAGE_TAG} \
 		-t ${REGISTRY_LOCAL}/${CERTSUITE_IMAGE_NAME_LEGACY}:${IMAGE_TAG} \
 		-t ${REGISTRY}/${CERTSUITE_IMAGE_NAME_LEGACY}:${IMAGE_TAG} \
-		-f Dockerfile .
+		-f Dockerfile . ;\
+	docker image prune -f
```
tonyskapunk, line 155: This one is risky, it may wipe other images that might be in use, if you could determine what intermediate images can be wiped out that would be great

sebrandon1: I'll take out the `docker image prune -f` now and we can merge the rest of the changes.  It reduces that dangling layer size from 5GB to 1GB and if DCI hit any more storage problems we can address it in another PR.

PR #3014, opened 2025-06-11
Suggestion of replacing debug log with a concrete diff result

### LINKED ISSUE #2970: Reduce DEBUG logging for fsdiff test for clearer logs
The fsdiff test prints out all changed or added files and folders detected during `fsdiff` however we have had a request to reduce the debug logging to not show the entire output of fsdiff.

itroyano:
Moving the debug output to show a concrete diff result instead of the entire podman diff output.
Fixes #2970.

### DIFF
--- a/tests/platform/cnffsdiff/fsdiff.go
+++ b/tests/platform/cnffsdiff/fsdiff.go
@@ -193,9 +193,6 @@ func (f *FsDiff) RunTest(containerUID string) {
 		return
 	}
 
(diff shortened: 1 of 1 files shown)


sebrandon1 (approved):
This is good, it should cut down on noise.

itroyano:
Thanks @sebrandon1 !
Should it be a debug level, or promoted to info level?

PR #3079, opened 2025-07-09
re-organizing dependencies to be more readable

acornett21:
## Motivation
The mod file as it stands is not human readable since there is a mix of direct and indirect dependencies scattered all over the place. 

## Changes
This changes the mod file to contain
- 2 required sections
   - 1 direct
   - 1 indirect
- 1 replaces section

## Notes
I do notice that some indirect dependencies changes versions when `go mod tidy` regenerated this section, if we feel that these need to be manually adjusted those can be as well.


sebrandon1 (approved):
I think this should work fine.  Not quite sure why go.mod tidy made it look like that in the first place.

PR #3084, opened 2025-07-14
adding new pod-count test to the observability suite

acornett21:
## Motivation
To implement a new test that conforms to the business ask in [EET-4647](https://issues.redhat.com/browse/EET-4647).

## Changes
- Exported autodiscover functions
   - `CreateLabels`
   - `FindPodsByLabels`
- Added function `testComparePodCount` to `tests/observability/suite.go`
   - Included the above as a test identifier 

## Notes/Questions
- ~~Do we want to see if there are _extra_ pods running after the fact?~~
- If this is the direction we want to go for this, tests can be added to this PR.

Assisted-by: Cursor

### DIFF
--- a/tests/observability/suite.go
+++ b/tests/observability/suite.go
@@ -24,16 +24,16 @@ import (
 	"strings"
 
 	"github.com/Masterminds/semver/v3"
-	"github.com/redhat-best-practices-for-k8s/certsuite/tests/common"
-	"github.com/redhat-best-practices-for-k8s/certsuite/tests/identifiers"
-	pdbv1 "github.com/redhat-best-practices-for-k8s/certsuite/tests/observability/pdb"
-
 	apiserv1 "github.com/openshift/api/apiserver/v1"
 	"github.com/redhat-best-practices-for-k8s/certsuite/internal/clientsholder"
 	"github.com/redhat-best-practices-for-k8s/certsuite/internal/log"
+	"github.com/redhat-best-practices-for-k8s/certsuite/pkg/autodiscover"
 	"github.com/redhat-best-practices-for-k8s/certsuite/pkg/checksdb"
 	"github.com/redhat-best-practices-for-k8s/certsuite/pkg/provider"
 	"github.com/redhat-best-practices-for-k8s/certsuite/pkg/testhelper"
+	"github.com/redhat-best-practices-for-k8s/certsuite/tests/common"
+	"github.com/redhat-best-practices-for-k8s/certsuite/tests/identifiers"
+	pdbv1 "github.com/redhat-best-practices-for-k8s/certsuite/tests/observability/pdb"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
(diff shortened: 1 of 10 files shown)


tests/observability/suite.go
```diff
@@ -423,3 +430,70 @@ func testAPICompatibilityWithNextOCPRelease(check *checksdb.Check, env *provider
 	// Add test results
 	check.SetResult(compliantObjects, nonCompliantObjects)
 }
+
+// Function to compare the number of running pods to those loaded during autodiscover at the start of test execution.
+func testComparePodCount(check *checksdb.Check, env *provider.TestEnvironment) {
+	oc := clientsholder.GetClientsHolder()
+
+	originalPods := env.Pods
```
edcdavid, line 438: Instead of comparing directly pods, why not compare the status of the deployment and stateful set instead ( ready replicas for instance). So in this test, you could use the <statefulset/deployment name>-replica-<replica number>/<namespace> to add a stable reference to pods in your results list. Otherwise, we would have a lot of false positives if the pod recreation test is triggered as a different uuid is appended to the pod after they are deleted and recreated.
For the orphan pods (the pods with owner references that are not a statefulset or replicaset), if any, we could compare them as already described here. See owner reference example: testPodsOwnerReference

acornett21: I wrote the test the way it was described in the jira, if we want something other then pods before and pods after, then is it really a pod comparison tests?

I'm happy to write whatever, but it's odd that this question came up, so is what we are trying to accomplish even needed, if we are unsure of what we really want?

edcdavid: > I wrote the test the way it was described in the jira, if we want something other then pods before and pods after, then is it really a pod comparison tests?

The jira describes counting the number of ready and non ready pods before and after running the suite. See below:
`Add a new test case that checks and collects the number of ready and non ready pods before and after running the certsuite to have an idea of any changes that have occurred while running the suite. This is a quick gauge to check the stability of the workload while running the suite.`

I like the idea of adding more details in term of which pods became not ready (or ready) after running the suite, but in my understanding this would work only if the pod names are stable after being re-created. Most pods names that belong to deployment/statefulsets will be different after pod re-creation because of the randomized identifier added at the end. For instance in the self-node-remediation-ds-g26dh pod name "-g26dh" will change when this pod is re-created.

acornett21: I get what you are saying, but if a pod gets a new uuid, doesn't that mean that it's not stable, and should be reported? 
> This is a quick gauge to check the stability of the workload while running the suite.

It was already stated that test will have false positives/negatives, this is why it's not mandatory anywhere (which IMO is odd, since other certifications don't work this way, but that's another story).

acornett21: How many tests in the test suite manipulate pods? Can tests be ordered / have priority?

hamadise:
test

tests/observability/suite.go
edcdavid: The tests that manipulate pods are identified by this function: GetNotIntrusiveSkipFn. This allows to have a switch to skip them. 4 tests: lifecycle-crd-scaling, lifecycle-deployment-scaling, lifecycle-statefulset-scaling, lifecycle-pod-recreation.
The tests are ordered in the order they are added in the code, we have not implemented changing the order or priority.
Maybe this test could be re-run after each intrusive tests?

acornett21:
going another direction, so going to close this and create a fresh pr.

PR #3095, opened 2025-07-18
PoC: Remove sriov-network-operator

sebrandon1:
Replace using the `sriov-network-operator` with using GVR resources and the `unstructured` library to query these objects.

This will remove the dependency completely and unblock us from updating the k8s library and others.


pkg/autodiscover/autodiscover_sriov.go
```diff
@@ -3,17 +3,32 @@ package autodiscover
...
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 )
 
-func getSriovNetworks(client *clientsholder.ClientsHolder, namespaces []string) (sriovNetworks []sriovNetworkOp.SriovNetwork, err error) {
-	var sriovNetworkList []sriovNetworkOp.SriovNetwork
+// SriovNetworkGVR defines the GroupVersionResource for SriovNetwork
+var SriovNetworkGVR = schema.GroupVersionResource{
+	Group:    "sriovnetwork.openshift.io",
```
acornett21, line 15: Question: Was this always the group, even those this repo/dep wasn't coming from the OCP org?

sebrandon1: Yes, that's what I've seen in their repo as their group.

acornett21: Odd...but okay! Thanks for clarifying.

PR #3113, opened 2025-07-28
adding a new PodStates struct to env for ready/non-ready pod comparison

acornett21:
## Implementation
- Created a new `PodStates` struct in the `env`, this is unstructured, but could be switched to exact `Phases` if we feel we want to keep track of all phases
   - I don't love the unstructured nature, but wasn't sure how granular we wanted to be. 
- Add logic to compare before/after `ready` states and log the diff
- Updated `NewClaimBuilder` to take in an `env` so that this can be modified, since `env` by nature is a singleton.
- Exported various methods that needed to be called outside of the `autodiscover` pkg.

# Testing

Below is the output of the `claim.json` with this. 
```json
❯ cat results/claim.json | jq .claim.configurations.podStates
{
  "AfterExecution": {
    "non-ready": 0,
    "ready": 2
  },
  "BeforeExecution": {
    "non-ready": 0,
    "ready": 3
  }
}
```

Below is the sample log file (though it seems to only log to the file and not the console, but that seem by design of the suite)
```log
[WARN] [Jul 28 11:36:24.356] [certsuite.go: 152] Some pods were not ready during entire test execution. See results/claim.json podStates section for more details
```

Below is an image from the results html parser
Screenshot From 2025-07-29 06-54-04

### DIFF
--- a/pkg/autodiscover/autodiscover.go
+++ b/pkg/autodiscover/autodiscover.go
@@ -59,8 +59,14 @@ const (
 	labelTemplate              = "%s/%s"
 )
 
+type PodStates struct {
+	BeforeExecution map[string]int
+	AfterExecution  map[string]int
+}
+
 type DiscoveredTestData struct {
 	Env                          configuration.TestParameters
+	PodStates                    PodStates
 	Pods                         []corev1.Pod
 	AllPods                      []corev1.Pod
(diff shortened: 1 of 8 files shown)


edcdavid (approved):
looks good to me, could you also please put a picture of the parser showing the new fearture?

PR #3120, opened 2025-07-30
Remove ARM64 job

sebrandon1:
The ARM server we have used for testing has been removed unexpectedly.  This needs to be removed as to not cause problems with PRs.


sebrandon1:
Actually, we just disabled the job in the UI for now and we can revisit this later if needed.

PR #3180, opened 2025-08-28
Update preflight to v1.14.1

sebrandon1:
https://github.com/redhat-openshift-ecosystem/openshift-preflight/releases/tag/1.14.1


acornett21:
@sebrandon1 That is exactly the intent, to make sure that partners set their own labels (or overwrite RH labels that are inherited from UBI) and ensure they do not infringe on Red Hat trademarks. The policy guide is [here](https://docs.redhat.com/en/documentation/red_hat_software_certification/2025/html-single/red_hat_openshift_software_certification_policy_guide/index#con-image-metadata-requirements_openshift-sw-cert-policy-container-images) and the trademark guide is [here](https://www.redhat.com/en/about/trademark-guidelines-and-policies).

This would have failed all the way back to version [1.11.0](https://github.com/redhat-openshift-ecosystem/openshift-preflight/releases/tag/1.11.0) and then became it's own check in [1.13.0](https://github.com/redhat-openshift-ecosystem/openshift-preflight/releases/tag/1.13.0) (which looks like the version being used here). So it's really unclear to me why you are seeing this now, since this is 10 months old at this point.

tests/identifiers/impact.go
```diff
@@ -149,6 +149,7 @@ const (
 	PreflightFollowsRestrictedNetworkEnablementGuidelinesImpact = `Non-compliance with restricted network guidelines can prevent deployment in air-gapped environments and violate security policies.`
 	PreflightHasLicenseImpact                                   = `Missing license information can create legal compliance issues and prevent proper software asset management.`
 	PreflightHasModifiedFilesImpact                             = `Modified files in containers can introduce security vulnerabilities, create inconsistent behavior, and violate immutable infrastructure principles.`
+	PreflightHasNoProhibitedLabelsImpact                        = `Prohibited labels can introduce security vulnerabilities, licensing issues, and compliance violations.`
```
acornett21, line 152: I would put something close to what's in preflights [suggestion](https://github.com/redhat-openshift-ecosystem/openshift-preflight/blob/main/internal/policy/container/has_prohibited_labels.go#L68), since this is just for trademark infringement.

sebrandon1: Sounds good I can update it.

PR #3222, opened 2025-09-15
Add parser update workflow

sebrandon1:
Similar to: https://github.com/redhat-best-practices-for-k8s/certsuite/pull/3195

Performs a check for new parser releases and then creates a new PR if one is available.

### DIFF
--- a/.github/workflows/parser-update.yaml
+++ b/.github/workflows/parser-update.yaml
@@ -32,31 +32,34 @@ jobs:
 
       - name: Determine current and latest parser versions
         id: versions
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
           set -euo pipefail
 
           CURRENT=$(jq -r '.parserTag' version.json)
           echo "current=${CURRENT}" >> $GITHUB_OUTPUT
 
-          # Fetch latest release with retry and exponential backoff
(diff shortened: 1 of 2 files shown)


.github/workflows/parser-update.yaml
```diff
@@ -0,0 +1,107 @@
...
+            fi
+
+            if [[ -n "${LATEST}" && "${LATEST}" != "null" ]]; then
+              break
+            fi
+
+            SLEEP=$((2 ** (attempt-1)))
+            echo "Attempt ${attempt} failed (status=${STATUS}). Retrying in ${SLEEP}s..."
+            sleep "${SLEEP}"
+          done
```
acornett21, line 61: I think what you have works, but do you really want to maintain all of this custom logic? `gh` cli ships with the action runners, so I think something as simple as the below would work and you can drop most (all?) of the code.

`❯ gh release list -R redhat-best-practices-for-k8s/parser --json name,isLatest --jq '.[] | select(.isLatest)|.name'
v0.5.3
`

tonyskapunk:
> from change #3222:
> 
>     * FAILURE https://www.distributed-ci.io/jobs/70cfff7c-3e41-481e-95dd-52cf5571475c/jobStates

The failure is unrelated to this change, a timeout was reached for downloading of a tool.

PR #3249, opened 2025-09-26
Add --unique-probe-name boolean config variable

sebrandon1:
Adds a new (development only) `--unique-probe-name` variable that QE can use when we are running the certsuite in parallel.

When enabled, the code will randomly add a suffix string to the certsuite-probe name so each run of the certsuite will have its own version of the probe.


cmd/certsuite/run/run.go
```diff
@@ -52,6 +52,8 @@ func NewCommand() *cobra.Command {
 	runCmd.PersistentFlags().String("connect-api-base-url", "", "Base URL for Red Hat Connect API")
 	runCmd.PersistentFlags().String("connect-api-proxy-url", "", "Proxy URL for Red Hat Connect API")
 	runCmd.PersistentFlags().String("connect-api-proxy-port", "", "Proxy port for Red Hat Connect API")
+	// When enabled, certsuite will deploy a uniquely named probe daemonset per run
+	runCmd.PersistentFlags().Bool("unique-probe-name", false, "Deploy a uniquely named probe daemonset per run")
```
acornett21, line 56: Wouldn't we also need to use cobra to mark this mutually-exclusive to another flag, so that this is truly never used by partners, and only for development?

PR #3285, opened 2025-10-20
Add consolidated free-disk-space action

sebrandon1:
We have been seeing out of disk space errors in our build-and-store actions runs because they're running out of disk space.  Also I found some areas where consolidation would make things "cleaner" instead of a lot of copy pasting.

Just as a note, these are the "pre" runs that builds the image and binary, not the actual test runs which are already taking care of their disk space.

### DIFF
--- a/.github/actions/free-disk-space/action.yml
+++ b/.github/actions/free-disk-space/action.yml
@@ -1,0 +1,17 @@
+name: 'Free Disk Space'
+description: 'Frees up disk space on Ubuntu runners by removing unnecessary packages and files'
+runs:
+  using: "composite"
+  steps:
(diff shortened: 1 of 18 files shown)


acornett21:
I think adding the cleanup makes sense, I've had to use this in other projects, but does it really make sense to add another action for this? Especially one that hasn't been touched in over 2 years? This seems like something we'd want manual/full control over.

When I looked into this for another project I came up with the below, based on all the issues out there on this for hosted runners.
https://github.com/operator-framework/operator-sdk/blob/89b409badcc4f38d1816ddfa13b4d7bfda1a5ce4/.github/workflows/clean-unused-disk-space.sh

PR #3376, opened 2025-12-17
Feedback review

### LINKED ISSUE #3267: add Red Hat review comments window in result.html
currently the result.html has partner feedback input window for partner to add feedback or justification for each failed test case,  but not input window for Red Hat's feedback review.  this request is for adding new input window for RedHat certification team to add comments and can be sent back to partner

dcurran90:
Adds a section for a response/review for feedback on failed certsuite tests in the results.html

> **BLOCKED**: This PR depends on changes in redhat-best-practices-for-k8s/parser#89.

closes #3267


ramperher:
How is this feedback review block going to appear in the HTML report? I'm checking the autogenerated report in the last DCI job and I don't see any difference with current executions with stable branch.
Note that we use the feedback text block to automatically include the feedback with DCI, we create the feedback.jsfile that is embedded in the HTML web page, so that the feedback appears when you open it. We're interested in knowing if this feature will change or not in this change, so that we can apply the proper modifications in our side.

dcurran90:
I see what you mean now. It's blocked by the changes to the parser located [here](https://github.com/redhat-best-practices-for-k8s/parser/pull/89) 
Once that is merged/released this PR should work appropriately and the changes reflected in the DCI run

PR #3611, opened 2026-05-01
Add parallel check execution for I/O-heavy test functions

sebrandon1:
## Summary

- Add `ForEachParallel[T any]` helper to `pkg/checksdb/` enabling bounded-concurrency parallel iteration over pods/containers/nodes
- Thread-safe `ParallelResult` accumulator with panic recovery and batch add methods to reduce mutex contention
- Convert 13 I/O-heavy test functions to parallel execution:
  - `testIsRedHatRelease`, `testIsSELinuxEnforcing`, `testContainersFsDiff`
  - `testNoSSHDaemonsAllowed`, `testOneProcessPerContainer`
  - `testUndeclaredContainerPortsUsage`, `testContainersLogging`
  - `testSchedulingPolicyInCPUPool`, `testRtAppsNoExecProbes`
  - `testHyperThreadingEnabled`, `testTainted`
  - `testOCPReservedPortsUsage`, `testPartnerSpecificTCPPorts`
- `testContainersFsDiff` uses per-node concurrency limit to avoid OOM from heavy `podman diff` operations
- Remove dead code from `netcommons` (~100 lines) after inlining reserved port checks into parallel callers

### CI QE Suite Comparison (main vs. this PR)

Compared against the latest QE run on `main` ([run 25471875821](https://github.com/redhat-best-practices-for-k8s/certsuite/actions/runs/25471875821)). Only suites containing parallelized checks shown:

| Suite | main (sequential) | PR (parallel) | Saved | % |
|---|---|---|---|---|
| accesscontrol4-k8s | 30m 18s | 17m 15s | 13m 03s | **43%** |
| networking2-k8s | 53m 21s | 30m 36s | 22m 45s | **43%** |
| networking1-k8s | 49m 23s | 29m 51s | 19m 32s | **40%** |
| networking3-k8s | 49m 56s | 30m 12s | 19m 44s | **40%** |
| accesscontrol11-k8s | 22m 45s | 13m 48s | 8m 57s | **39%** |
| lifecycle6-k8s | 29m 53s | 18m 46s | 11m 07s | **37%** |
| accesscontrol7-k8s | 20m 46s | 13m 00s | 7m 46s | **37%** |
| accesscontrol10-k8s | 21m 51s | 13m 58s | 7m 53s | **36%** |
| accesscontrol3-k8s | 14m 46s | 9m 29s | 5m 17s | **36%** |
| observability-k8s | 29m 17s | 19m 43s | 9m 34s | **33%** |
[...]

PR #3671, opened 2026-05-26
Bug fix for platform-alteration-base-image: use per-node mutex

greyerof:
Parallel check execution added in PR #3611 makes this check to fail in compact/MNO clusters if there's more than one container under test on the same node. It only fails on OCP < 4.13 that need the probe's podman to be mounted on the node's tmp folder.

The mount/unmount and deleting operation of that custom podman must be done sequentially in the same node. This fix is still compatible with parallel execution of the check, but only for containers deployed in different nodes.

### DIFF
--- a/tests/platform/suite.go
+++ b/tests/platform/suite.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"strconv"
 	"strings"
+	"sync"
 
 	clientsholder "github.com/redhat-best-practices-for-k8s/certsuite/internal/clientsholder"
 	"github.com/redhat-best-practices-for-k8s/certsuite/internal/log"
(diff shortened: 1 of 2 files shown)


sebrandon1 (approved):

tests/platform/suite.go
```diff
@@ -225,8 +226,16 @@ func testServiceMesh(check *checksdb.Check, env *provider.TestEnvironment) {
 // testContainersFsDiff test that all CUT did not install new packages are starting
 func testContainersFsDiff(check *checksdb.Check, env *provider.TestEnvironment) {
 	// podman diff is resource-heavy; limit concurrency to avoid OOM on probe pods
+
+	// We also need a per-node map of mutexes to limit the concurrency per node on ocp < 4.13
+	// This avoids conflicts while mounting/unmounting and deleting the podman temp folder on the node.
+	mutexPerNode := make(map[string]*sync.Mutex)
+	for _, node := range env.Nodes {
+		mutexPerNode[node.Data.Name] = &sync.Mutex{}
+	}
```
sebrandon1, line 235: Looking at this again, should this be a mutex per `ProbePods` instead of per node?

greyerof: I think it's the same, as there's one probe per node. Strictly speaking, only the nodes with workload under test should have a mutex, but I just didn't want to add more code. Or maybe I'm missing something else?

sebrandon1: Just a thought since we only really care about mutex'ing the probe pods themselves but yeah I think we should be okay with a mutex per node anyways.  👍

PR #3698, opened 2026-06-03
fix: correct copy-paste bug in updateTnf managed statefulsets loop

yaronh12:
The `updateTnf` function in `webserver/webserver.go` iterated over `data.ManagedDeployments` when populating `managedStatefulsets`, causing statefulsets to silently receive deployment names instead of their own values.

**Fix:** Change the data source from `data.ManagedDeployments` to `data.ManagedStatefulsets` in the second loop.

Added table-driven regression test in `webserver/webserver_test.go` that verifies deployments and statefulsets are stored independently


sebrandon1 (approved):
Nice find!

PR #3734, opened 2026-06-17
fix: add nil check for probe pod in hyperthreading detection

### LINKED ISSUE #3732: platform-alteration-hyperthread-enable panics with nil pointer dereference during parallel execution
During certsuite execution on a multi-node OCP cluster, the platform-alteration-hyperthread-enable test panics repeatedly with:
[ERROR] [parallel.go: 64] [platform-alteration-hyperthread-enable] Panic during parallel check execution: runtime error: invalid memory address or nil pointer dereference
The panic occurs 6 times (once per node) but is caught by Go's recover, so the suite completes. Likely cause: the test assumes hyperthreading info is present in the node's hardware inventory, but the data is nil/empty on these nodes.
Impact: Test result for platform-alteration-hyperthread-enable is unreliable. Suite completes but the panic clutters logs.
Environment: OCP 4.18.22
K8s v1.31.11
Certsuite v5.5.21

sebrandon1:
## Summary

Fixes the nil pointer panic in the `platform-alteration-hyperthread-enable` test when probe pods are missing for nodes.

## Problem

The test was panicking with:
```
[ERROR] [parallel.go: 64] [platform-alteration-hyperthread-enable] Panic during parallel check execution: runtime error: invalid memory address or nil pointer dereference
```

This occurred because `env.ProbePods[nodeName]` was accessed without checking if the entry exists. On clusters where probe pods aren't deployed to all nodes, this causes repeated panics (once per node without a probe pod).

## Solution

Added a nil check before accessing the probe pod:
- Check if probe pod exists in the map
- Check if the pod itself is nil
- Return a descriptive error instead of panicking

## Impact

**Before:** Runtime panic (6 times on 6-node cluster)  
**After:** Returns error "no probe pod found for node X"

The test result is now reliable and logs are clean. The test properly reports nodes without probe pods as errors rather than crashing.

## Testing

- ✅ Package compiles successfully
- ✅ All 96 provider package tests pass
- ✅ golangci-lint reports 0 issues

Fixes #3732

### DIFF
--- a/pkg/provider/nodes.go
+++ b/pkg/provider/nodes.go
@@ -142,10 +142,17 @@ const (
 func (node *Node) IsHyperThreadNode(env *TestEnvironment) (bool, error) {
 	o := clientsholder.GetClientsHolder()
 	nodeName := node.Data.Name
-	ctx := clientsholder.NewContext(env.ProbePods[nodeName].Namespace, env.ProbePods[nodeName].Name, env.ProbePods[nodeName].Spec.Containers[0].Name)
+
+	// Check if probe pod exists for this node
+	probePod, exists := env.ProbePods[nodeName]
+	if !exists || probePod == nil {
+		return false, fmt.Errorf("no probe pod found for node %s", nodeName)
+	}
+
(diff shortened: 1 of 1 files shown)


sebrandon1:
Superseded by clean branch with review fixes

PR #3737, opened 2026-06-18
Fix nil pointer dereference in testLimitedUseOfExecProbes

Ruclo:
The StartupProbe and ReadinessProbe log messages incorrectly referenced cut.LivenessProbe.PeriodSeconds instead of the corresponding probe, causing a panic when LivenessProbe was nil.


sebrandon1 (approved):
Verified the bug — the StartupProbe and ReadinessProbe log lines incorrectly reference `cut.LivenessProbe.PeriodSeconds`, which panics when LivenessProbe is nil. Fix is correct and minimal.

PR #3790, opened 2026-07-20
fix: keep nodes without MachineConfig on hosted control planes

yaronh12:
### Why
Konflux EaaS run of certsuite v5.5.23 against ptp-operator on a HyperShift guest (OCP ~4.21) left `env.Nodes` empty even though three workers and healthy probe pods existed.
Guest nodes lacked `machineconfiguration.openshift.io/currentConfig`. `createNodes` treated the cluster as OCP, logged "Failed to get machineConfig name", and dropped every node. That emptied `claim.nodes.nodeSummary` while ProbePods / `nodesHwInfo` still listed the nodes.
Fallout included:
- panic in `platform-alteration-boot-params` (`GetMcKernelArguments`)
- fail in `platform-alteration-base-image` ("No mutex found for node")
- cascade skips for other platform checks
- lifecycle HA/scaling "not enough nodes" via empty `GetWorkerCount()`
HyperShift guests are OCP without usable guest MCO annotations; management-cluster MCO does not apply to guest workers.

certsuite assumed every OCP cluster exposes usable guest MachineConfig (MCO). HyperShift guests are still detected as OCP, but worker nodes often lack `machineconfiguration.openshift.io/currentConfig` (MCO lives on the management cluster / NodePool side, not on guest workers).

### What
- Keep nodes in `env.Nodes` when Mc annotation is missing or Mc fetch fails
- `GetMcKernelArguments` returns `ErrNoMachineConfig` instead of panicking when Mc is absent
- Boot-params / sysctl callers use `errors.Is(..., ErrNoMachineConfig)` → skip (do not fail)

Regular OCP with MCO is unchanged.

### DIFF
--- a/tests/platform/suite.go
+++ b/tests/platform/suite.go
@@ -434,9 +434,14 @@ func testHugepages(check *checksdb.Check, env *provider.TestEnvironment) {
 		}
 
 		hpTester, err := hugepages.NewTester(&node, probePod, clientsholder.GetClientsHolder())
+		if errors.Is(err, provider.ErrNoMachineConfig) {
+			check.LogInfo("Skipping node %q: %v", nodeName, err)
+			continue
+		}
 		if err != nil {
 			check.LogError("Unable to get node hugepages tester for node %q, err: %v", nodeName, err)
 			nonCompliantObjects = append(nonCompliantObjects, testhelper.NewNodeReportObject(nodeName, "Unable to get node hugepages tester", false))
+			continue
 		}
 
 		if err := hpTester.Run(); err != nil {
--- a/tests/platform/suite.go
+++ b/tests/platform/suite.go
@@ -502,7 +502,12 @@ func testSysctlConfigs(check *checksdb.Check, env *provider.TestEnvironment) {
 			continue
 		}
 
-		mcKernelArgumentsMap := bootparams.GetMcKernelArguments(env, cut.NodeName)
+		mcKernelArgumentsMap, err := bootparams.GetMcKernelArguments(env, cut.NodeName)
+		if err != nil {
+			check.LogError("Could not get MachineConfig kernel arguments for node %q, error: %v", cut.NodeName, err)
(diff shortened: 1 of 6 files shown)


tests/platform/suite.go
```diff
@@ -502,7 +502,12 @@ func testSysctlConfigs(check *checksdb.Check, env *provider.TestEnvironment) {
 			continue
 		}
 
-		mcKernelArgumentsMap := bootparams.GetMcKernelArguments(env, cut.NodeName)
+		mcKernelArgumentsMap, err := bootparams.GetMcKernelArguments(env, cut.NodeName)
+		if err != nil {
+			check.LogError("Could not get MachineConfig kernel arguments for node %q, error: %v", cut.NodeName, err)
```
aabughosh: lets skip here and not fail because if there is no mc (hypershift) we dont want to mark it as err

aabughosh:
look good to me

edcdavid (commented):
Looks good to me, but don't we need also a nil check for the hugepage code using the machineconfig too? https://github.com/redhat-best-practices-for-k8s/certsuite/blob/6d5fb023c3119d7802c9154ec4362c7ff849dd90/tests/platform/hugepages/hugepages.go

tests/platform/suite.go
sebrandon1, line 445:
```suggestion
nonCompliantObjects = append(nonCompliantObjects, testhelper.NewNodeReportObject(nodeName, "Unable to get node hugepages tester", false))
continue
}
```

sebrandon1: Add a `continue` here if we're unable to get that `hpTester`.  NewTester() returns a pointer so we should continue out of the loop.  This was a pre-existing bug.

sebrandon1 (changes requested):
Looks good, I think @edcdavid is right we need to add nil checks in: 
- https://github.com/redhat-best-practices-for-k8s/certsuite/blob/main/tests/platform/hugepages/hugepages.go#L132
- https://github.com/redhat-best-practices-for-k8s/certsuite/blob/main/tests/platform/hugepages/hugepages.go#L214

edcdavid (approved):

sebrandon1 (approved):

PR #3869, opened 2026-08-20
fix: probe unsecured ports from the node-local probe pod

sebrandon1:
## Summary
openssl ran from a random probe pod, not the one on the same node as the workload. Same-node probes worked. Cross-node probes hit the 5s timeout (exit 124) and were treated as unreachable, which counts as a pass. A plaintext app could pass or fail depending on which probe was chosen.

On Kind, UBI9 openssl 3 prints TLS 1.3 as `Cipher is TLS_AES_...` with no colon. We only matched the old `Cipher    :` format, so a real TLS listener looked like plaintext.

- Probe from the workload node (same probe `ss` already uses)
- Retry openssl exit 124 when there is no usable output
- Parse OpenSSL 3 `Cipher is` as a negotiated cipher
- Missing node-local probe, empty PodIP, or empty container list is non-compliant
- Kind `plain-http-spread` / `tls-spread` require both anti-affinity replicas to be classified

Smoke CNF pods still serve plaintext `:8080`, so this test now fails for real. GitHub smoke already expects that. DCI still expects a pass until its allowlist is updated. sample-workload is unchanged.

## Related PRs
| PR | Title | Status |
|----|-------|--------|
| [redhat-best-practices-for-k8s/certsuite#3614](https://github.com/redhat-best-practices-for-k8s/certsuite/pull/3614) | feat: add unsecured container ports TLS detection test | Merged |
| [redhat-best-practices-for-k8s/checks#59](https://github.com/redhat-best-practices-for-k8s/checks/pull/59) | feat: add TLS minimum version and unsecured container ports checks | Merged |
| [redhat-best-practices-for-k8s/certsuite#3554](https://github.com/redhat-best-practices-for-k8s/certsuite/pull/3554) | feat: migrate to checks library as single source of truth | Open |
| [dci-labs/bos2-ci-config#704](https://github.com/dci-labs/bos2-ci-config/pull/704) | feat: add networking-unsecured-container-ports to expected results exclusion | Merged |
| [dci-labs/vcp-config#1631](https://github.com/dci-labs/vcp-config/pull/1631) | feat: add networking-unsecured-container-ports to expected results exclusion | Merged |

## Test Plan
- [x] `go test ./tests/networking/ ./tests/networking/tlsversion/ ./internal/crclient/ -count=1`
- [x] Kind `plain-http-spread`: test fails with at least 2 non-compliant plaintext ports
- [x] Kind `tls-spread`: test passes with at least 2 compliant objects containing "uses TLS"
- [x] GitHub Actions (unit, lint, local smoke, container smoke, QE image)
- [ ] DCI `verify_tests` is expected to fail: sample-workload plaintext `:8080` plus an unrelated `lifecycle-pod-recreation` miss

### DIFF
--- a/.github/actions/test-scenarios/networking/README.md
+++ b/.github/actions/test-scenarios/networking/README.md
@@ -14,6 +14,20 @@ Validates that the `networking-tls-minimum-version` test correctly checks servic
 | TLS 1.2 Allowed (compliant) | `tls12-allowed.yaml` | TLS 1.2 + 1.3, Intermediate ciphers | `passed` |
 | Plain HTTP (compliant) | `plain-http.yaml` | No TLS | `passed` |
 
+### unsecured-container-ports
+
+Validates that `networking-unsecured-container-ports` probes plaintext TCP
+listeners on every node, not only the node of a randomly chosen probe pod.
+Smoke Kind clusters have two workers; required pod anti-affinity spreads two
+replicas across them. The runner also requires at least two non-compliant
+objects so one same-node plaintext hit plus a cross-node timeout cannot pass
+as a generic FAIL.
+
+| Scenario | Manifest | Workload | Expected Test Result |
+|---|---|---|---|
+| Plain HTTP spread across nodes | `plain-http-spread.yaml` | Two replicas, plaintext :8080, hostname anti-affinity | `failed` (min 2 non-compliant objects) |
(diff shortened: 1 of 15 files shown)


ramperher (approved):
LGTM, from DCI side, it's behaving as it should now.

PR #3899, opened 2026-09-01
feat: distinguish probe tool execution errors from test case failures

rocrisp:
## Summary

- Introduces `ProbeToolExecError` in `internal/crclient` to wrap failures that originate from the `certsuite-probe` daemonset (unavailable pod, container not found, context deadline exceeded), making them distinguishable from genuine test case failures in the claim file
- `ExecCommandContainerNSEnter` now returns `ProbeToolExecError` on all probe failure paths instead of generic `fmt.Errorf`
- `netutil.GetSSHDaemonPort` and `GetListeningPorts` propagate the error type via `%w` so callers can inspect it with `errors.As`
- `testNoSSHDaemonsAllowed` and the two networking `GetListeningPorts` call sites call `check.SetResultError` (not `AddNonCompliantObject`) when a probe tool execution error is detected, keeping `Failed` reserved for genuine test case non-compliance
- Fixes broken `errors.As` chain in `GetContainerPidNamespace` (`%v` → `%w`) that would have silently defeated `IsProbeToolExecError` detection in the performance test path
- Adds `TestIsProbeToolExecError_doubleWrapped` to cover the two-level wrapping case

## Test plan

- [ ] `make build` compiles cleanly
- [ ] `UNIT_TEST=true go test ./internal/crclient/... -v` — all 7 tests pass including `TestIsProbeToolExecError_doubleWrapped`
- [ ] `make lint` passes
- [ ] End-to-end: when `certsuite-probe` pods are unavailable at runtime, affected checks record `error` result instead of `failed`

### DIFF
--- a/internal/crclient/crclient.go
+++ b/internal/crclient/crclient.go
@@ -17,6 +17,7 @@
 package crclient
 
 import (
+	"errors"
 	"fmt"
 	"regexp"
 	"strconv"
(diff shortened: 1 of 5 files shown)


rocrisp:
upstream already addressed the issue via b8cdd4a6 and c4875ce0

PR #3901, opened 2026-09-01
fix: affiliated-certification-helm-version false failure in disconnected environments

### LINKED ISSUE #3890: affiliated-certification-helm-version incorrectly requires cert DB access in disconnected environments
## Summary

In disconnected environments where `--offline-db` is not provided, the `affiliated-certification-helm-version` test case fails with:

```
cannot access the certification DB
```

This is a false failure. `testHelmVersion` only queries the Kubernetes API for Tiller pods using label selector `app=helm,name=tiller` — it never reads from the certification database.

## Root Cause

In `tests/certification/suite.go`, `LoadChecks()` registers all four affiliated-certification checks under a single group with a shared `beforeEachFn` that unconditionally calls `certdb.GetValidator`:

```go
var beforeEachFn = func(check *checksdb.Check) error {

    var err error
    validator, err = certdb.GetValidator(env.GetOfflineDBPath())  // called for every check
    if err != nil {
        return fmt.Errorf("cannot access the certification DB, err: %w", err)
    }
    return nil
}
```

Three checks actually need `validator`:
- `affiliated-certification-operator-is-certified`
- `affiliated-certification-helmchart-is-certified`
- `affiliated-certification-container-is-certified-digest`

One check does **not** use `validator` at all:
- `affiliated-certification-helm-version` — only calls `clients.K8sClient.CoreV1().Pods("").List(...)` to detect Tiller pods

This was introduced during the Ginkgo removal refactor (PR #1648) where all four checks were grouped under a shared `beforeEachFn` by mistake.

## Fix

Split the shared initialization:

1. Keep `beforeEachFn` for `env` initialization only — all four checks need the test environment.
2. Move `certdb.GetValidator(...)` into the `WithCheckFn` lambda of each of the three checks that actually need it.

**`tests/certification/suite.go`:**

```go
// beforeEachFn only initializes the test environment
var beforeEachFn = func(check *checksdb.Check) error {
    return nil
}
```

In `LoadChecks()`, the three validator-dependent checks each initialize the validator locally:

```go
checksGroup.Add(checksdb.NewCheck(identifiers.GetTestIDAndLabels(identifiers.TestOperatorIsCertifiedIdentifier)).
    WithSkipCheckFn(skipIfNoOperatorsFn).
    WithCheckFn(func(c *checksdb.Check) error {
        v, err := certdb.GetValidator(env.GetOfflineDBPath())
        if err != nil {
            return fmt.Errorf("cannot access the certification DB, err: %w", err)
        }
        testAllOperatorCertified(c, &env, v)
        return nil
    }))

checksGroup.Add(checksdb.NewCheck(identifiers.GetTestIDAndLabels(identifiers.TestHelmIsCertifiedIdentifier)).
    WithSkipCheckFn(skipIfNoHelmChartReleasesFn).
    WithCheckFn(func(c *checksdb.Check) error {
        v, err := certdb.GetValidator(env.GetOfflineDBPath())
        if err != nil {
            return fmt.Errorf("cannot access the certification DB, err: %w", err)
        }
        testHelmCertified(c, &env, v)
        return nil
    }))

checksGroup.Add(checksdb.NewCheck(identifiers.GetTestIDAndLabels(identifiers.TestContainerIsCertifiedDigestIdentifier)).
    WithSkipCheckFn(testhelper.GetNoContainersUnderTestSkipFn(&env)).
    WithCheckFn(func(c *checksdb.Check) error {
        v, err := certdb.GetValidator(env.GetOfflineDBPath())
        if err != nil {
            return fmt.Errorf("cannot access the certification DB, err: %w", err)
        }
        testContainerCertificationStatusByDigest(c, &env, v)
        return nil
    }))
```

The `helm-version` check registration is unchanged — no `certdb.GetValidator` call needed.

The package-level `validator` variable can be removed since it is no longer shared state.

**Files changed:** `tests/certification/suite.go` only.

rocrisp:
## Summary

- Fixes #3890: `affiliated-certification-helm-version` was failing with `cannot access the certification DB` in disconnected environments where `--offline-db` is not provided
- Root cause: the shared `beforeEachFn` unconditionally called `certdb.GetValidator` for all four `affiliated-certification-*` checks, including `helm-version` which never uses the cert DB — it only queries the Kubernetes API for Tiller pods
- Fix: strip `certdb.GetValidator` out of `beforeEachFn` (now only initializes the test environment) and move it into the `WithCheckFn` lambda of the three checks that actually need it (`operator-is-certified`, `helmchart-is-certified`, `container-is-certified-digest`)
- Remove the package-level `validator` variable — no longer needed as shared state
- Add `TestHelmVersion_NoTiller` and `TestHelmVersion_TillerPresent` to verify `testHelmVersion` runs correctly against the k8s API without any cert DB access

## Test plan

- [x] `make build` passes
- [x] `UNIT_TEST=true go test ./tests/certification/... -v` — all 9 tests pass including the two new `TestHelmVersion_*` tests
- [x] `golangci-lint run ./tests/certification/...` — 0 issues
- [ ] In a disconnected environment without `--offline-db`, `affiliated-certification-helm-version` no longer fails with `cannot access the certification DB`


sebrandon1:
I'm fixing that shfmt issue in https://github.com/redhat-best-practices-for-k8s/certsuite/pull/3902

sebrandon1 (approved):

greyerof (approved):

PR #3904, opened 2026-09-01
feat: mark check result as Error when probe pod exec fails

### LINKED ISSUE #3900: feat: mark check result as Error (not Failed) when probe pod exec fails
## Background

PR #3870 improved probe pod exec failure handling by dumping live probe status and adding descriptive failure reasons to the claim. However, the check result is still recorded as **Failed** via `AddNonCompliantObject`, making it indistinguishable from a genuine CNF non-compliance in the claim file.

## Problem

When `certsuite-probe` is unavailable (OOMKilled, CrashLoopBackOff, container not ready), affected checks such as \`access-control-ssh-daemons\` and \`networking-unsecured-container-ports\` record:

- Result: **Failed**
- Reason: "Probe pod exec failed; not a CNF finding: ..."

The descriptive reason is helpful for humans reading logs, but claim file consumers (automated tools, dashboards, partner portals) cannot programmatically distinguish a probe outage from a real test failure — both show up as \`Failed\`.

## Expected Behavior

When a probe exec failure is detected via \`crclient.IsProbeExecFailure(err)\`, the check should call \`check.SetResultError()\` instead of \`result.AddNonCompliantObject()\`. This keeps **Failed** reserved for genuine CNF non-compliance and surfaces probe outages as **Error** in the claim file.

## Affected Call Sites

| File | Function | Current behavior |
|------|----------|-----------------|
| \`tests/accesscontrol/suite.go\` | \`testNoSSHDaemonsAllowed\` | \`AddNonCompliantObject\` with probe message |
| \`tests/networking/suite.go\` | unsecured container ports check | \`AddNonCompliantObject\` with probe message |

## Implementation

Replace the probe-outage branches in the affected call sites:

```go
// before
result.AddNonCompliantObject(testhelper.NewPodReportObject(..., sshDaemonCheckFailureReason(err), false))

// after
if crclient.IsProbeExecFailure(err) {
    check.SetResultError(fmt.Sprintf("probe exec failure for pod %s/%s: %v", put.Namespace, put.Name, err))
    return
}
result.AddNonCompliantObject(testhelper.NewPodReportObject(..., "Failed to get the ssh port for pod", false))
```

## Related

- Closes #3888 (partially addressed by #3870; this completes the result-status distinction)
- #3870 added the probe status dump and descriptive reasons

rocrisp:
## Summary

- When `crclient.IsProbeExecFailure` detects a probe pod outage (OOMKilled, CrashLoopBackOff, container not ready), call `check.SetResultError()` instead of `result.AddNonCompliantObject()` so the claim file records **Error** rather than **Failed**
- Keeps **Failed** reserved for genuine CNF non-compliance; probe infrastructure failures now surface as **Error** for automated consumers (dashboards, partner portals)
- `check.SetResult()` already guards against overwriting an Error result, so calling `SetResultError` inside `ForEachParallel` callbacks is safe

**Affected call sites:**
- `testNoSSHDaemonsAllowed` in `tests/accesscontrol/suite.go` — two error paths (`GetSSHDaemonPort`, `GetListeningPorts`)
- `checkPodPortTLS` in `tests/networking/suite.go` — one error path (`getListeningPorts`)

**Cleanup:** removes dead helper functions `checkFailureReason`, `sshDaemonCheckFailureReason`, `listeningPortsCheckFailureReason`, and the `sshDaemonProbeExecMsg` constant (no remaining callers after the fix).

Closes #3900

## Test plan

- [x] `UNIT_TEST=true go test ./tests/accesscontrol/... ./tests/networking/...` passes
- [x] `make build` compiles cleanly
- [x] `golangci-lint run` introduces no new issues (pre-existing `goconst` in `tlsversion.go` is unrelated)
- [ ] In a cluster with a crashed probe pod, the affected checks now show `error` result in the claim file instead of `failed`


tests/accesscontrol/suite.go
```diff
@@ -1029,7 +1010,11 @@ func testNoSSHDaemonsAllowed(check *checksdb.Check, env *provider.TestEnvironmen
 		port, err := netutil.GetSSHDaemonPort(cut)
 		if err != nil {
 			check.LogError("Could not get ssh daemon port on %q, err: %v", cut, err)
-			result.AddNonCompliantObject(testhelper.NewPodReportObject(put.Namespace, put.Name, sshDaemonCheckFailureReason(err), false))
+			if crclient.IsProbeExecFailure(err) {
+				check.SetResultError(fmt.Sprintf("probe exec failure for pod %s/%s: %v", put.Namespace, put.Name, err))
+				return
+			}
```
sebrandon1, line 1014: This is the right call for the ssh-daemons path.

Nit: the non-probe `AddNonCompliantObject` below dropped the actual `err`. #3870 had `Failed to get the ssh port for pod: %v` — worth putting that back so a real nsenter/ss failure is still diagnosable in the claim.

tests/accesscontrol/suite.go
```diff
@@ -983,27 +983,8 @@ func testNamespaceResourceQuota(check *checksdb.Check, env *provider.TestEnviron
...
-	return checkFailureReason(err, sshDaemonProbeExecMsg, "Failed to get the ssh port for pod")
-}
-
-func listeningPortsCheckFailureReason(err error) string {
-	return checkFailureReason(err, sshDaemonProbeExecMsg, "Failed to get the listening ports for pod")
-}
-
 func probeExecFailureReason(err error) string {
 	if crclient.IsProbeExecFailure(err) {
 		return fmt.Sprintf("Probe pod exec failed; not a CNF finding: %v", err)
```
sebrandon1: `probeExecFailureReason` is now only used by `testOneProcessPerContainer` (GetPidFromContainer / getNbOfProcessesInPidNamespace), which still does `AddNonCompliantObject` on probe exec failures.

If ssh-daemons is Error, that check should be too — same `IsProbeExecFailure` path from #3870. Otherwise a probe OOM still looks like a CNF finding there.

tests/networking/suite.go
```diff
@@ -330,6 +330,10 @@ func checkPodPortTLS(check *checksdb.Check, put *provider.Pod, ch clientsholder.
 	listeningPorts, err := getListeningPorts(put.Containers[0])
 	if err != nil {
 		check.LogError("Failed to get pod %q listening ports, err: %v", put, err)
+		if crclient.IsProbeExecFailure(err) {
+			check.SetResultError(fmt.Sprintf("probe exec failure for pod %s/%s: %v", put.Namespace, put.Name, err))
```
sebrandon1, line 338: Looks correct for unsecured-ports.

The other `GetListeningPorts` call sites in this file (`testUndeclaredContainerPortsUsage` and `testReservedPortsUsageParallel`) still `AddNonCompliantObject` on probe exec failures. Please convert those too, or track them in a follow-up — a dead probe still records Failed on undeclared/reserved ports.

tests/accesscontrol/suite.go
```diff
@@ -1050,7 +1035,11 @@ func testNoSSHDaemonsAllowed(check *checksdb.Check, env *provider.TestEnvironmen
 		listeningPorts, err := netutil.GetListeningPorts(cut)
 		if err != nil {
 			check.LogError("Failed to get the listening ports for Pod %q, err: %v", put, err)
-			result.AddNonCompliantObject(testhelper.NewPodReportObject(put.Namespace, put.Name, listeningPortsCheckFailureReason(err), false))
+			if crclient.IsProbeExecFailure(err) {
+				check.SetResultError(fmt.Sprintf("probe exec failure for pod %s/%s: %v", put.Namespace, put.Name, err))
+				return
+			}
```
sebrandon1, line 1039: Same pattern as the ssh-port branch — good. Same nit: include `: %v` / the error on the non-probe `AddNonCompliantObject` message.

`GetListeningPorts` failures in `testUndeclaredContainerPortsUsage` and `testReservedPortsUsageParallel` (this file) still go straight to Failed. Those should get the same `IsProbeExecFailure` → `SetResultError` treatment, or a follow-up issue.

sebrandon1 (commented):
This is the right follow-up to #3870 — we kept Failed there on purpose so the claim text was clearer; Error is the missing distinction. The two ssh/unsecured-ports branches look correct, and the existing SetResult guard will not overwrite Error.

Please also convert `testOneProcessPerContainer` and the other `GetListeningPorts` error paths (`testUndeclaredContainerPortsUsage`, `testReservedPortsUsageParallel`), or file a follow-up. Otherwise a probe OOM still shows as Failed on those checks.

JUnit (`Errors` is hardcoded to 0 and there is no `<error>` element), the results table (`getResultsSummary` only counts passed/failed/skipped), `failedCtr`/the exit warning, and `certsuite claim show failures` still ignore `error`. Claim JSON will distinguish it; the rest will not.

tests/accesscontrol/suite.go
sebrandon1: Correction: those two `GetListeningPorts` sites are in `tests/networking/suite.go`, not this file. The comment on the unsecured-ports branch covers them.

pkg/claimhelper/claimhelper.go
```diff
@@ -248,6 +264,15 @@ func populateXMLFromClaim(c claim.Claim, startTime, endTime time.Time) TestSuite
 			testCase.Failure = nil
 		}
 
+		// Populate the error message if the test case errored (probe/infrastructure failure)
+		if testCase.Status == TestStateError {
+			testCase.Error = &ErrorMessage{}
+			testCase.Error.Text = c.Results[testID].CheckDetails
+			testCase.Error.Type = "probe-exec-failure"
```
greyerof, line 271: Hardcoding here the error type to "probe-exec-failure" doesn't look like a good idea to me. [Panics (and timeouts, if I'm not wrong)](https://github.com/redhat-best-practices-for-k8s/certsuite/blob/cfd5fcf5b6650f4cdba53e0193b112732e60e34a/pkg/checksdb/checksgroup.go#L110) could also flag a test as errored. What about extending the Check struct with errorType and errorReason so we can use them here? It might require some refactors over there, though...

rocrisp: Good catch.
Opened #3907 to track this refactor separately.

greyerof (approved):

tests/accesscontrol/suite.go
rocrisp: @sebrandon1 These changes have been adopted. Please take a look when you get a chance.
Thank you.

